PR #6, opened 2025-11-13
Add manual XML import for apple health

czajkub:
This allows the developer (or user) to:

- use the /.../import/apple/xml endpoint to create a presigned url to upload an xml file to S3
- trigger a celery task to import records and workouts into the database

How it works:
- First: create an S3 bucket and SQS queue that receives events whenever an xml file is uploaded to the bucket
- Fill environmental variables in config/.env 
- Receive presigned url from the endpoint and upload your xml file (https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html)
- A celery task is queued automatically - it will poll SQS every 20 seconds (less than url expiry) and import data into database when it receives an event

### DIFF
--- a/backend/app/integrations/celery/tasks/poll_sqs_task.py
+++ b/backend/app/integrations/celery/tasks/poll_sqs_task.py
@@ -62,7 +62,7 @@ def poll_sqs_messages():
                         object_key = record["s3"]["object"]["key"]
 
                         # Enqueue Celery task
-                        task = process_uploaded_file.delay(bucket_name, object_key)
+                        process_uploaded_file.delay(bucket_name, object_key)
                         processed_count += 1
 
                 # Delete message from queue after processing
--- a/backend/app/integrations/celery/tasks/process_upload_task.py
+++ b/backend/app/integrations/celery/tasks/process_upload_task.py
@@ -21,50 +21,39 @@ def process_uploaded_file(bucket_name: str, object_key: str):
         bucket_name: S3 bucket name
         object_key: S3 object key (path)
     """
-    db = SessionLocal()
 
-    temp_xml_file = None
+    with SessionLocal() as db:
 
-    try:
-        temp_dir = tempfile.gettempdir()
-        temp_xml_file = os.path.join(temp_dir, f"temp_import_{object_key.split('/')[-1]}")
-        
-        user_id = object_key.split('/')[-3]
-
-        s3_client.download_file(bucket_name, object_key, temp_xml_file)
+        temp_xml_file = None
 
         try:
-            _import_xml_data(db, temp_xml_file, user_id)
-            db.commit()
-        except Exception as e:
-            db.rollback()
-            raise e
+            temp_dir = tempfile.gettempdir()
+            temp_xml_file = os.path.join(temp_dir, f"temp_import_{object_key.split('/')[-1]}")
+            
+            user_id = object_key.split('/')[-3]
+
+            s3_client.download_file(bucket_name, object_key, temp_xml_file)
+
+            try:
+                _import_xml_data(db, temp_xml_file, user_id)
+                db.commit()
+            except Exception as e:
+                db.rollback()
+                raise e
+            
+            result = {
+                "bucket": bucket_name,
+                "input_key": object_key,
+                "user_id": user_id,
+                "status": "success",
+                "message": "Import completed successfully",
+            }
+
+            return result
+
         finally:
-            db.close()
-
-        result = {
-            "bucket": bucket_name,
-            "input_key": object_key,
-            "user_id": user_id,
-            "status": "success",
-            "message": "Import completed successfully",
-        }
-
-        return result
-
-    except Exception as e:
-        result = {
-            "bucket": bucket_name,
-            "input_key": object_key,
-            "user_id": user_id,
-            "status": "failed",
-            "error": str(e),
-        }
-        return result
-
-    finally:
-        if temp_xml_file and os.path.exists(temp_xml_file):
-            os.remove(temp_xml_file)
+            if temp_xml_file and os.path.exists(temp_xml_file):
+                os.remove(temp_xml_file)
 
 
 def _import_xml_data(db: Session, xml_path: str, user_id: str) -> None:


backend/app/api/routes/v1/import_data.py
```diff
@@ -50,3 +52,16 @@ async def import_data_healthion(
     """Import health data from file upload or JSON."""
     content_str, content_type = content[0], content[1]
     return await hk_import_service.import_data_from_request(db, content_str, content_type, user_id)
+
+
+@router.post("/users/{user_id}/import/apple/xml", response_model=PresignedURLResponse)
```
KaliszS, line 56: user_id is not user anywhere in the function body

app/services/apple/apple_xml/aws_service.py
```diff
@@ -0,0 +1,23 @@
...
+AWS_BUCKET_NAME = settings.aws_bucket_name
+AWS_REGION = settings.aws_region
+DEFAULT_EXPIRATION = 300  # 5 minutes
+
+try:
+    s3_client = boto3.client(
+        "s3",
+        region_name=AWS_REGION,
+        aws_access_key_id=os.getenv("AWS_ACCESS_KEY_ID"),
+        aws_secret_access_key=os.getenv("AWS_SECRET_ACCESS_KEY"),
```
KaliszS: Why not to take those values from settings?

app/services/apple/apple_xml/aws_service.py
```diff
@@ -0,0 +1,23 @@
+import os
+from pathlib import Path
```
KaliszS: Unused imports

backend/app/services/apple/apple_xml/aws_service.py
```diff
@@ -0,0 +1,23 @@
...
+from dotenv import load_dotenv
+
+from app.config import settings
+
+
+AWS_BUCKET_NAME = settings.aws_bucket_name
+AWS_REGION = settings.aws_region
+DEFAULT_EXPIRATION = 300  # 5 minutes
+
+try:
```
KaliszS, line 10: Maybe there is some settings to set up retries? If yes, we should add this to the app config and use it here.

app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
+from datetime import datetime
+from logging import Logger, getLogger
+
+from fastapi import HTTPException
+from botocore.exceptions import ClientError
+from datetime import datetime, UTC
```
KaliszS: duplicated import

backend/app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
...
+            error_code = e.response["Error"]["Code"]
+            if error_code == "404":
+                raise HTTPException(status_code=500, detail="S3 bucket not found")
+            elif error_code == "403":
+                raise HTTPException(status_code=500, detail="Access denied to S3 bucket")
+            else:
+                raise HTTPException(status_code=500, detail=f"S3 bucket error: {error_code}")
+
+    def create_presigned_url(self, request: PresignedURLRequest) -> str:
+        self.validate_bucket_exists()
```
KaliszS, line 45: Are we going to switch buckets during run-time? If no, I would suggest to run this validator once (for example using fastapi startup event: https://fastapi.tiangolo.com/advanced/events/#startup-event), because we generate here additional query to s3 with every request.

backend/app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
...
+
+class ImportService:
+    def __init__(self, log: Logger, **kwargs):
+        self.log = log
+
+    def generate_file_key(self, user_id: str, filename: Optional[str] = None) -> str:
+        timestamp = datetime.now(UTC)
+
+        if filename:
+            clean_filename = "".join(c for c in filename if c.isalnum() or c in ".-_")
```
KaliszS, line 19: Maybe we should handle situation where clean_filename will be empty?

app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
...
+            s3_client.head_bucket(Bucket=AWS_BUCKET_NAME)
+            return True
+        except ClientError as e:
+            error_code = e.response["Error"]["Code"]
+            if error_code == "404":
+                raise HTTPException(status_code=500, detail="S3 bucket not found")
+            elif error_code == "403":
+                raise HTTPException(status_code=500, detail="Access denied to S3 bucket")
+            else:
+                raise HTTPException(status_code=500, detail=f"S3 bucket error: {error_code}")
```
KaliszS: I suggest to use status codes from fastapi for better IDE support.

app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
...
+
+    def create_presigned_url(self, request: PresignedURLRequest) -> str:
```
KaliszS: PresignedURLResponse is the correct return type

backend/app/services/apple/apple_xml/presigned_url_service.py
```diff
@@ -0,0 +1,81 @@
...
+from datetime import datetime, UTC
+from typing import Optional
+
+from app.services.apple.apple_xml.aws_service import s3_client, AWS_BUCKET_NAME
+from app.schemas.apple.apple_xml.aws import PresignedURLRequest, PresignedURLResponse
+
+
+class ImportService:
+    def __init__(self, log: Logger, **kwargs):
+        self.log = log
```
KaliszS, line 13: We don't use logs, maybe it would be cool to include logging file key names? Of course with debug level.

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+                        continue
+                    statistic = stat.attrib.copy()
+                    statistic["workout_id"] = workout_create.id
+                    statistics_create = self.update_record("stat", statistic, user_id)
+                    if statistics_create:
+                        statistics.append(statistics_create)
+            elem.clear()
+
+        # yield remaining records and workout pairs
+        yield records, zip(workouts, statistics)
```
KaliszS: Are you sure it will work? have you testes that solution? Because you zip here list and list of lists.

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+        records: list[HKRecordCreate] = []
+        workouts: list[HKWorkoutCreate] = []
+        statistics: list[list[HKWorkoutStatisticCreate]] = []
+
+        for event, elem in ET.iterparse(self.xml_path, events=("start",)):
+            if elem.tag == "Record" and event == "start":
+                if len(records) >= self.chunk_size:
+                    yield records, zip(workouts, statistics)
+                    records = []
+                    workouts = []
```
KaliszS, line 144: Shouldn't we also clear statistics? And not only here

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+        Create schema objects from XML document data.
+
+        Args:
+            kind: Type of record to create ('record', 'workout', or 'stat')
+            document: Dictionary of attributes from XML element
+            user_id: User ID to associate with the records
+        """
+        for field in self.DATE_FIELDS:
+            if field in document:
+                document[field] = datetime.strptime(document[field], "%Y-%m-%d %H:%M:%S %z")
```
KaliszS: We should handle error if dat format will be unparsable.

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+        "startDate",
+        "endDate",
+        "sum",
+        "average",
+        "maximum",
+        "minimum",
+        "unit",
+    )
+
+    def update_record(
```
KaliszS: Maybe it would be more readable to have 3 sub-methods called here? `_create_record`, `_create_workout`, `_create_statistics`

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+from uuid import uuid4
+from xml.etree import ElementTree as ET
+
+from app.schemas import HKWorkoutStatisticCreate, HKWorkoutCreate, HKRecordCreate
+
+
+class XMLService:
+    def __init__(self, path: Path):
+        self.xml_path: Path = path
+        self.chunk_size: int = 50_000
```
KaliszS: It should come from our app config.

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+                            value=document[field],
+                            unit=document["unit"],
+                        )
+                    )
+
+        return statistics if statistics else None
+
+    def parse_xml(
+        self, user_id: str = None
+    ) -> Generator[tuple[list[HKWorkoutCreate], list[list[HKWorkoutStatisticCreate]]], Any, None]:
```
KaliszS: I think second type should be `None` instead of `Any`

app/schemas/apple/apple_xml/aws.py
```diff
@@ -0,0 +1,53 @@
+from enum import Enum
+from typing import Optional
```
KaliszS: We are using python above 3.10, so we should use `type | None` instead of `Optional[type]`

app/schemas/apple/apple_xml/aws.py
```diff
@@ -0,0 +1,53 @@
...
+
+from pydantic import BaseModel, Field
+
+
+DEFAULT_EXPIRATION = 300  # 5 minutes
```
KaliszS: It should be included into config, not as a global const.

app/schemas/apple/apple_xml/aws.py
```diff
@@ -0,0 +1,53 @@
...
+        default=DEFAULT_EXPIRATION,
+        ge=60,
+        le=3600,
+        description="URL expiration time in seconds (1 min - 1 hour)",
+    )
+    max_file_size: Optional[int] = Field(
+        default=50 * 1024 * 1024,  # 50MB
+        ge=1024,  # 1KB minimum
+        le=500 * 1024 * 1024,  # 500MB maximum
+        description="Maximum file size in bytes",
```
KaliszS: Both fields have default values different from None, so they shouldn't be marked as optional.

KaliszS: We also shouldn't use numbers like that. They should be moved into separate consts or to config settings.

app/schemas/apple/apple_xml/aws.py
```diff
@@ -0,0 +1,53 @@
...
+
+class FileType(str, Enum):
+    JSON = "application/json"
+    XML = "application/xml"
+    CSV = "text/csv"
+    TEXT = "text/plain"
+
+
+class PresignedURLRequest(BaseModel):
+    user_id: str = Field(..., min_length=1, max_length=100, description="Unique user identifier")
```
KaliszS: Now I see where user_id usage went from endpoint router ;) I believe we should do it as in other endpoints, so just take user_id from url and pass it to the function separately.

app/services/apple/apple_xml/xml_service.py
```diff
@@ -0,0 +1,164 @@
...
+                # append workout statistics to workout
+                for stat in elem:
+                    if stat.tag != "WorkoutStatistics":
+                        continue
+                    statistic = stat.attrib.copy()
+                    statistic["workout_id"] = workout_create.id
+                    statistics_create = self.update_record("stat", statistic, user_id)
+                    if statistics_create:
+                        statistics.append(statistics_create)
+            elem.clear()
```
KaliszS: I don't think that's enough to clear it from memory, because references are still saved in the parent. Probably we should also follow parent tag and clear it here too. But it won't work in `start` event (there's `end` event type for that, in the `start` phase children are not rendered yet, because it means only open tag in XML).

app/services/apple/apple_xml/aws_service.py
czajkub: it was an omission, now it comes from settings

app/services/apple/apple_xml/aws_service.py
czajkub: removed

backend/app/services/apple/apple_xml/aws_service.py
czajkub: removed it completely, it only tries once

app/services/apple/apple_xml/presigned_url_service.py
czajkub: removed

backend/app/services/apple/apple_xml/presigned_url_service.py
czajkub: I image there might be some weird situation where a dev might want to use more than one bucket for data or it's switched sometime after startup, I would personally leave it this way unless you believe it should be done with startup events

backend/app/services/apple/apple_xml/presigned_url_service.py
czajkub: Yup, reverting to timestamp if its empty

app/services/apple/apple_xml/presigned_url_service.py
czajkub: done

app/services/apple/apple_xml/presigned_url_service.py
czajkub: yup

backend/app/services/apple/apple_xml/presigned_url_service.py
czajkub: Done

app/services/apple/apple_xml/xml_service.py
czajkub: It didn't, now workouts and statistics aren't zipped, just a tuple of 3 elements is yielded - i realised after parsing stats don't need to be associated with workouts in any way since the workout id is already passed to statistic beforehand

czajkub: I mean, it did, then I changed something, then it didn't and now it works the way i described

backend/app/services/apple/apple_xml/xml_service.py
czajkub: yes, forgot that statistics werent a part of workouts list

app/services/apple/apple_xml/xml_service.py
czajkub: definitely makes more sense, split it up as well as parsing date fields

app/services/apple/apple_xml/xml_service.py
czajkub: right

app/schemas/apple/apple_xml/aws.py
czajkub: see 1 lower

app/schemas/apple/apple_xml/aws.py
czajkub: I put it into settings as a tuple like this:
```
# min, default, max
presigned_url_expiration_seconds: tuple[int, int, int] = (60, 300, 3600)  # 1 min, 5 min, 1 hour
presigned_url_max_filesize: tuple[int, int, int] = (1024, 50 * 1024 * 1024, 1024 * 1024 * 1024)  # 1KB, 50MB, 1GB
```
do you think this is a good idea, or should it just be split into 3 separate settings?
in a .env file, it would be like this:
`PRESIGNED_URL_EXPIRATION_SECONDS=60,300,3600`, but i didn't put those 2 in the .env.example, since for max filesize you would need to precalculate it and put a weird magic number for 1gb for example

app/schemas/apple/apple_xml/aws.py
czajkub: yes, now it's removed from request and put like in other endpoints

app/services/apple/apple_xml/xml_service.py
czajkub: Changed it to end event, it should also clear child tags as well
```
clear()
Resets an element. This function removes all subelements, clears all attributes, and sets the text and tail attributes to None.
```
the clearing is done after the record / workout tag separately and not at the end, but it shouldn't be an issue since there are only record and workout tags in apple's xml. The other way (just clearing it at the end of the iteration) messed up workout statistics, so i did it this way

app/schemas/apple/apple_xml/aws.py
KaliszS: Why do we even need so many predefined values for expiration? If all of them are requred, then maybe it would be more readable to leave them inside the code after all, not in config.

backend/app/integrations/celery/tasks/poll_sqs_task.py
```diff
@@ -0,0 +1,113 @@
...
+
+    num_polls = expiration_seconds // 20
+    print(f"[poll_sqs_task] Starting with {num_polls} polls (every 20 seconds)")
+
+    for i in range(num_polls):
+        print(f"[poll_sqs_task] Poll {i + 1}/{num_polls}")
+        poll_sqs_messages()
+
+        if i < num_polls - 1:
+            time.sleep(20)
```
KaliszS: why so? I think celery have some kind of cron schedules, when you can set schedule for every 20 seconds. But you would have to investigate it, I'm not sure about that. AI should help, but anyway I suggest to learn about it by yourself, because it may use some very old versions.

backend/app/integrations/celery/tasks/poll_sqs_task.py
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Poll SQS messages for file uploads at regular intervals.
+    Polls every 20 seconds for expiration_seconds // 20 iterations.
+
+    Args:
+        expiration_seconds: Total time to poll for (in seconds)
+    """
+
+    num_polls = expiration_seconds // 20
+    print(f"[poll_sqs_task] Starting with {num_polls} polls (every 20 seconds)")
```
KaliszS: I believe it would be better to have logs instead of prints here. And I would also reduce number of those printed messages anyway.

backend/app/integrations/celery/tasks/poll_sqs_task.py
```diff
@@ -0,0 +1,113 @@
...
+                        print(f"[poll_sqs_messages] Queued task {task.id}")
+                        processed_count += 1
+
+                # Delete message from queue after processing
+                sqs.delete_message(QueueUrl=QUEUE_URL, ReceiptHandle=receipt_handle)
+
+            except Exception as e:
+                print(f"[poll_sqs_messages] Error processing message {message_id}: {str(e)}")
+                failed_count += 1
+                continue
```
KaliszS, line 74: Is there any retry logic if we have failed counts?

backend/app/integrations/celery/tasks/poll_sqs_task.py
```diff
@@ -0,0 +1,113 @@
...
+
+                # Process S3 records
+                for record in message_body["Records"]:
+                    if record.get("eventSource") == "aws:s3":
+                        bucket_name = record["s3"]["bucket"]["name"]
+                        object_key = record["s3"]["object"]["key"]
+                        print(f"[poll_sqs_messages] Queuing file upload task: s3://{bucket_name}/{object_key}")
+
+                        # Enqueue Celery task
+                        task = process_uploaded_file.delay(bucket_name, object_key)
```
KaliszS: Maybe we can pack them into batches before enqueueing?

backend/app/integrations/celery/tasks/process_upload_task.py
```diff
@@ -0,0 +1,87 @@
...
+        s3_client.download_file(bucket_name, object_key, temp_xml_file)
+
+        try:
+            _import_xml_data(db, temp_xml_file, user_id)
+            db.commit()
+        except Exception as e:
+            db.rollback()
+            raise e
+        finally:
+            db.close()
```
KaliszS: Dangerous thing here - database is closed in finally block, so if it goes to except namespace, database won't be closed and we're going to have memory leak.

But anyway, I would just use `with` statement (context manager) instead of try except finaly. It will close database automatically. I think you can just use here AI to rewrite this,, it should handle that kind of tasks.

app/schemas/apple/apple_xml/aws.py
czajkub: It's for min/max validation in the api request and a default value for the field, I'll put it in the code then

backend/app/integrations/celery/tasks/poll_sqs_task.py
czajkub: I did it that way so that the database can sync when the user uploads the data, and the check (polling messages from the queue) happens only when the user creates the pre-signed url, since we don't know how long will the file upload take, and using webhooks instead means adding more aws services like sns and adding more complexity (which is the opposite of what we're aiming for)

backend/app/integrations/celery/tasks/poll_sqs_task.py
czajkub: If a message is failed, it usually means that something is not set up right (like a notification comes from a different source than an s3 upload), so instead of retrying, an error message is returned and the message is skipped

backend/app/integrations/celery/tasks/poll_sqs_task.py
czajkub: I don't know if it's a good idea, since these XML files can be very large (the limit is 500mb), so I think that queueing a lot of these tasks into one could pose some memory issues

Or do you mean batching the tasks themselves and sending a few of them at once? I don't think the amount of data will be large enough to feel a difference, since uploading these XML's shouldn't be done so frequently (since it can take a long time because of the filesize)

backend/app/integrations/celery/tasks/poll_sqs_task.py
czajkub: yep, changed to logger

backend/app/integrations/celery/tasks/process_upload_task.py
czajkub: using context manager now

czajkub: also changed the way the tasks are scheduled - it doesnt block threads now

PR #7, opened 2025-11-17
Move backend to separate directory

bartmichalak:
Since we're heading to monorepo, we need to separate backend to the dedicated directory.


bartmichalak:
Linter issues will be resolved via separate PR

Issue #16, opened 2025-11-19
Unable to provide Redis credentials (Redis Authentication Error in Railway Deployment)


bartmichalak:
## Problem
OAuth endpoints return `500 Internal Server Error` on Railway due to Redis authentication failure.

## Error
```
redis.exceptions.AuthenticationError: Authentication required.
```

**Affected endpoint:** `GET /api/v1/oauth/{provider}/authorize`

## Root Cause
Railway managed Redis requires authentication (`redis://default:password@host:port/db`), but current code doesn't properly parse and use the password from `REDIS_URL`.

## Environment
- **Works:** ✅ Local development (Docker Redis without password)
- **Fails:** ❌ Railway deployment (managed Redis with password)

## Solution
Update Redis client initialization to properly handle Railway's connection string format:

```python
# Current (broken):
redis_client = Redis(
    host=settings.redis_host,  # localhost
    password=None  # No password parsed
)

# Fixed:
parsed = urlparse(settings.redis_url)
redis_client = Redis(
    host=parsed.hostname,
    port=parsed.port or 6379,
    password=parsed.password,  # Extract from URL
    username=parsed.username,  # 'default' for Redis 6.0+
    db=int(parsed.path.lstrip('/')) if parsed.path else 0
)
```

## Required Changes
1. Update `app/core/config.py` to prioritize `REDIS_URL`/`REDIS_PUBLIC_URL`
2. Parse connection string properly in Redis client initialization
3. Support both local (no password) and Railway (with password) environments

## Railway Variables
Railway provides:
- `REDIS_URL` / `REDIS_PUBLIC_URL` - Full connection string with credentials
- `REDISHOST`, `REDISPORT`, `REDISPASSWORD`, `REDISUSER` - Individual parameters (fallback)

## Related
- Railway Redis uses `default` username (Redis 6.0+ ACL)
- Requires `redis-py >= 4.2.0` for full ACL support

PR #17, opened 2025-11-19
fix: Allow to provide credentials for Redis connection

bartmichalak:
- Added support for REDIS_URL to retrieve Redis connection parameters.
- Introduced `get_redis_connection_params` method to parse Redis URL or fallback to individual settings.
- Updated OAuthService to utilize the new Redis connection parameters.

Resolves https://github.com/the-momentum/open-wearables/issues/16

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -139,21 +139,6 @@ class Settings(BaseSettings):
     def auth0_issuer_url(self) -> str:
         return f"https://{self.auth0_domain}/"
 
-    def get_redis_connection_params(self) -> dict[str, Any]:
-        """
-        Get Redis connection parameters from individual settings.
-        
-        Returns:
-            dict with keys: host, port, db, password, username
-        """
-        return {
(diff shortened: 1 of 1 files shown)


backend/app/config.py
```diff
@@ -61,9 +62,14 @@ class Settings(BaseSettings):
     token_lifetime: int = 3600
 
     # REDIS SETTINGS
+    # Priority: REDIS_URL > individual settings
+    redis_url: str | None = None  # Connection string: redis://username:password@host:port/db
```
KaliszS: Wouldn't be better to use `@property`, like with db_uri and built it from separate settings below?

bartmichalak: Requested changes implemented. Additionally I used `from_url` method to initialise the connection with Redis. 

I think you could incorporate those changes in this PR: https://github.com/the-momentum/open-wearables/pull/45/files

and then this PR might be closed.

PR #44, opened 2025-11-24
[old] Update models to unified workout model

czajkub:
Update models to new unified model, clean up unused code and adjust current apple import to new models

### DIFF
--- a/backend/app/api/routes/v1/garmin_webhooks.py
+++ b/backend/app/api/routes/v1/garmin_webhooks.py
@@ -6,6 +6,7 @@ from typing import Annotated
 from fastapi import APIRouter, Header, HTTPException, Request
 
 from app.database import DbSession
+from app.services import garmin_import_service
 
 router = APIRouter()
 logger = getLogger(__name__)
@@ -158,6 +159,14 @@ async def garmin_ping_notification(
                     logger.error(f"Error processing activity notification: {str(e)}")
                     results["errors"].append(str(e))
 
+
+            # temporary visual change
+            try:
+                garmin_import_service.load_data(db, payload['activities'], internal_user_id)
+            except Exception as e:
+                logger.error(f"Error loading data from Garmin: {str(e)}")
+                results["errors"].append(f"Error loading data from Garmin: {str(e)}")   
+
         # Process other summary types (activityDetails, dailies, etc.)
         for summary_type in ["activityDetails", "dailies", "epochs", "sleeps"]:
             if summary_type in payload:
@@ -258,11 +267,19 @@ async def garmin_push_notification(
                         }
                     )
                     results["processed"] += 1
+                    
 
                 except Exception as e:
                     logger.error(f"Error processing activity notification: {str(e)}")
                     results["errors"].append(str(e))
 
+            # no internal user id, just temporary visual change
+            try:
+                garmin_import_service.load_data(db, payload['activities'], internal_user_id)
+            except Exception as e:
+                logger.error(f"Error loading data from Garmin: {str(e)}")
+                results["errors"].append(f"Error loading data from Garmin: {str(e)}")
+            
         return results
 
     except Exception as e:
--- a/backend/app/api/routes/v1/vendor_workouts.py
+++ b/backend/app/api/routes/v1/vendor_workouts.py
@@ -114,6 +114,7 @@ async def get_user_workouts(
             filter_by_modification_time=filter_by_modification_time,
         )
         
+        # temporary, for testing before services refactor
         suunto_import_service.load_data(db, raw_workouts, user_id)
         return raw_workouts
 
--- a/backend/app/services/__init__.py
+++ b/backend/app/services/__init__.py
@@ -5,6 +5,7 @@ from .apple.healthkit.import_service import import_service as hk_import_service
 from .apple.workout_service import workout_service
 from .apple.workout_statistic_service import workout_statistic_service
 from .suunto_import_service import import_service as suunto_import_service
+from .garmin_import_service import import_service as garmin_import_service
 from .developer_service import (
     DeveloperDep,
     current_active_user_optional,
@@ -27,6 +28,7 @@ __all__ = [
     "ae_import_service",
     "hk_import_service",
     "suunto_import_service",
+    "garmin_import_service",
     
     "workout_service",
     "workout_statistic_service",


backend/app/models/workout.py
```diff
@@ -12,16 +12,15 @@
...
     str_10,
     str_100,
-    str_50,
 )
 
 
 class Workout(BaseDbModel):
     id: Mapped[PrimaryKey[UUID]]
-    provider_id: Mapped[UUID | None] = None
+    provider_id: Mapped[str_100 | None] = None
```
FBegiello, line 20: If `provider ` is meant to point to device this is fine, but if `provider` means source platform (i.e. AppleHealth, Suunto) this should be an enum

backend/app/models/workout.py
```diff
@@ -12,16 +12,15 @@
...
     user_id: Mapped[FKUser]
 
-    type: Mapped[str_50 | None] = None
+    type: Mapped[str_100 | None] = None
```
FBegiello, line 23: Same here - for raw data this would be fine as we do not need to account for each detail in names between providers, but aiming for unified health record we should have an enum of supported workouts/workout metrics to properly handle them and assign preprocessing pipeline per type - with a predefined list this is easy to manage

backend/app/models/workout.py
```diff
@@ -12,16 +12,15 @@
...
     duration: Mapped[numeric_10_2]
     durationUnit: Mapped[str_10]
```
FBegiello, line 25: I'd be against defining duration and duration unit separately - rather, select a base unit (i.e. seconds) and have `duration_seconds` as a field - easier to operate on for analytical modules as you have a guaranteed unified base and dont risk mismanaged units.

Even for longer workout sessions (like few hours) this should be manageble as we at most hit 5 digits but gain a more robust structure

backend/app/models/workout_statistic.py
```diff
@@ -1,25 +1,32 @@
...
     user_id: Mapped[FKUser]
-    workout_id: Mapped[FKWorkout]
+    workout_id: Mapped[FKWorkout | None] = None
 
-    type: Mapped[str_50]
-    value: Mapped[numeric_10_2]
+    type: Mapped[str_100]
+    sourceName: Mapped[str_100]
+    startDate: Mapped[datetime_tz]
+    endDate: Mapped[datetime_tz]
```
FBegiello, line 26: I'd suggest `start_timestamp` - better naming convention and more to the point

backend/app/models/workout_statistic.py
```diff
@@ -1,25 +1,32 @@
...
-    id: Mapped[PrimaryKey[int]]
+    id: Mapped[PrimaryKey[UUID]]
     user_id: Mapped[FKUser]
-    workout_id: Mapped[FKWorkout]
+    workout_id: Mapped[FKWorkout | None] = None
 
-    type: Mapped[str_50]
-    value: Mapped[numeric_10_2]
+    type: Mapped[str_100]
+    sourceName: Mapped[str_100]
```
FBegiello, line 24: We have possible semantic conflict between here and `workout` table - if we assume that the statistics can come from different devices (one device covers more/less than others) then the general workout is already a unified structure and does not have a direct source. If the workout is registered only from one device, and workout statistics are inherited only from what was avaliable then this is redundant

I'd advise assuming the former - workout is an intro structure working as unified record, and we pick and hoose statistics from different devices. This is in line with the unified health record vision

backend/app/models/workout_statistic.py
```diff
@@ -1,25 +1,32 @@
...
+    startDate: Mapped[datetime_tz]
+    endDate: Mapped[datetime_tz]
+    min: Mapped[numeric_10_3 | None] = None
+    max: Mapped[numeric_10_3 | None] = None
+    avg: Mapped[numeric_10_3 | None] = None
```
FBegiello, line 29: We can add standard deviation and variance as base statistical measures as well - both should be usefull in most cases and easy to calculate, especially for closed period of time with clearly defined time points

backend/app/repositories/apple/auto_export/workout_repository.py
```diff
@@ -77,20 +78,20 @@ def get_workout_summary(self, db_session: DbSession, workout_id: UUID) -> dict:
         # Get heart rate summary
         hr_stats = (
             db_session.query(
-                func.avg(HeartRateData.avg).label("avg_hr"),
-                func.max(HeartRateData.max).label("max_hr"),
-                func.min(HeartRateData.min).label("min_hr"),
+                func.avg(WorkoutStatistic.avg).label("avg_hr"),
+                func.max(WorkoutStatistic.max).label("max_hr"),
+                func.min(WorkoutStatistic.min).label("min_hr"),
             )
```
FBegiello, line 84: This is fine for a base approach but we should be carefull with raw data handling like that - TODO note is to leave room for proper preprocessing (i.e. data cleaning, outlier detection/removal)

backend/app/repositories/apple/auto_export/workout_repository.py
```diff
@@ -77,20 +78,20 @@ def get_workout_summary(self, db_session: DbSession, workout_id: UUID) -> dict:
         # Get heart rate summary
         hr_stats = (
             db_session.query(
-                func.avg(HeartRateData.avg).label("avg_hr"),
-                func.max(HeartRateData.max).label("max_hr"),
-                func.min(HeartRateData.min).label("min_hr"),
+                func.avg(WorkoutStatistic.avg).label("avg_hr"),
```
FBegiello, line 81: For heart rate be especially carefull with averages - if we have any erroneous datapoints, this can get messy quickly. Assuming we can receive null datapoints as 0 value (possible with some approaches to sampling/recording) then those values would dramatically lower the general average, and we can safely assume that the user's heart rate was not 0 as she/he is probably still alive

backend/app/repositories/apple/healthkit/workout_repository.py
```diff
@@ -93,20 +95,20 @@ def get_workout_summary(self, db_session: DbSession, workout_id: UUID) -> dict:
...
         hr_stats = (
             db_session.query(
-                func.avg(HeartRateData.avg).label("avg_hr"),
-                func.max(HeartRateData.max).label("max_hr"),
-                func.min(HeartRateData.min).label("min_hr"),
+                func.avg(WorkoutStatistic.avg).label("avg_hr"),
+                func.max(WorkoutStatistic.max).label("max_hr"),
+                func.min(WorkoutStatistic.min).label("min_hr"),
             )
-            .filter(HeartRateData.workout_id == workout_id)
```
FBegiello, line 100: See comments for same case below

backend/app/schemas/suunto/workout_import.py
```diff
@@ -0,0 +1,40 @@
+from pydantic import BaseModel
+
+    
+class HeartRateJSON(BaseModel):
+    workoutMaxHR: int   
+    workoutAvgHR: int
+    userMaxHR: int
+    avg: int
+    hrmax: int
+    max: int
```
FBegiello, line 10: Can be extended as per model data with variance and standard deviation

backend/app/schemas/suunto/workout_import.py
```diff
@@ -0,0 +1,40 @@
+from pydantic import BaseModel
+
+    
+class HeartRateJSON(BaseModel):
+    workoutMaxHR: int   
+    workoutAvgHR: int
+    userMaxHR: int
+    avg: int
```
FBegiello, line 8: Average (as well as mentioned variance and stdv) should be float to avoid forcibly rounding values

backend/app/schemas/suunto/workout_import.py
```diff
@@ -0,0 +1,40 @@
...
+    
+class WorkoutJSON(BaseModel):
+    workoutId: int
+    # unix timestamp (ms)
+    startTime: int
+    stopTime: int
+    # seconds
+    totalTime: int
+    
+    totalDistance: int
```
FBegiello, line 29: I'd specify unit - if meters then `totalDistanceMeters` if kilometers, then change type to float and change name accordingly

backend/app/schemas/suunto/workout_import.py
```diff
@@ -0,0 +1,40 @@
...
+    stepCount: int
+    energyConsumption: int
```
FBegiello, line 31: This should be float as well to avoid rounding up

backend/app/schemas/suunto/workout_import.py
```diff
@@ -0,0 +1,40 @@
...
+    
+    hrdata: HeartRateJSON
+    
+    gear: DeviceJSON | None = None
```
FBegiello, line 35: See mentions of source and unified model in db models - as a goal we are aiming for a unified health record, so some workout values might come from different devices

backend/app/schemas/common.py
```diff
@@ -1,46 +1,39 @@
...
-    )
-    end_date: str | None = Field(
-        None, description="ISO 8601 format (e.g., '2023-12-31T23:59:59Z')"
-    )
-    limit: int | None = Field(
-        20, ge=1, le=100, description="Number of results to return"
-    )
+
+    start_date: str | None = Field(None, description="ISO 8601 format (e.g., '2023-12-01T00:00:00Z')")
+    end_date: str | None = Field(None, description="ISO 8601 format (e.g., '2023-12-31T23:59:59Z')")
```
FBegiello, line 14: I'd change date to timestamp for consistency and clearer naming convention

backend/app/schemas/workout.py
FBegiello, line 1: See previous comments about data, aditional fields and datatypes

backend/app/schemas/workout_statistics.py
FBegiello, line 1: ^ditto

backend/app/models/workout.py
czajkub: This field is mostly meant for id's for workouts from the provider, so that we don't allow duplicates during imports. The reason it's now a string instead of UUID is because some providers (like Polar) just use integers instead of actual UUID's

czajkub: Maybe we could add another field (enum) just for differentiating providers?

backend/app/models/workout.py
czajkub: That's definitely worth considering, but access to what activity types are supported (from what i've seen) is somewhat limited and would require some digging. We definitely should do a mapping for these types, but it's gonna take some time

backend/app/models/workout.py
czajkub: Good idea, for now the duration unit is always minutes, so we could just drop the field altogether, or do you think it should be seconds?

backend/app/models/workout.py
FBegiello: Adding new field will be fine - I'd suggest to preserve current `provider_id` as `workout_provider_id` for example and introduce enum as `provider_name ` but just adding new enum is fine for now

backend/app/models/workout.py
FBegiello: We can mark that as TODO for later if this info is not easily accessible

backend/app/models/workout_statistic.py
czajkub: Should we call it `timestamp` when it's a datetime object?

backend/app/models/workout_statistic.py
czajkub: Yes, the statistics are bound to one workout, so sourceName is redundant, however `workout` type and `workout_statistic`, since workout type is something like running and workout_statistic type heart_rate or step_count

backend/app/models/workout.py
FBegiello: I'd drop the field entirely and rename the `duration` to clearly state the unit - would be most convenient to work with for external developers. 
I'm also a bit sceptical as to recording in minutes - with such approach, we have little wiggle room for recording shorter durations or lengths with fractions of a minute. With seconds you can easily use fractions even for milliseconds (can be overkill for workouts, but might be useful - always better to preserve more and do not use it than less and lose information)

backend/app/models/workout_statistic.py
czajkub: If we want to add those fields, we would need to add it into `Workout` - `WorkoutStatistic` only contains one record. The issue with that is that most providers (like Suunto, Garmin, Polar) provide only a single measurement for a workout - usually one maximum and one average for heart rate for example, so we can't calculate these metrics (we can, but with only one record it provides no valuable information). It would be useful if a provider (for example apple does that) gives multiple measurements for a single workout

czajkub: These statistics would be useful if, for example, we would import standalone heart rate measurements over some time period (or any other metric), but with workouts (so far we only have workouts) most providers don't include those measurements

backend/app/repositories/apple/auto_export/workout_repository.py
czajkub: Yes, that is definitely an important thing to consider

backend/app/schemas/suunto/workout_import.py
czajkub: This schema is only for importing data from a JSON returned by providers, we could include those fields for responses and/or include them in workouts (saved to our db)

backend/app/schemas/suunto/workout_import.py
czajkub: information about the unit is included in the `WorkoutStatistic` model, but a thing to consider is normalizing this data - changing miles to km etc.

backend/app/models/workout_statistic.py
FBegiello: We can go for `start_datetime` as well - the main point is limiting it to `startDate` when we measure events in minutes to hours might be a bit inconsistent.

backend/app/models/workout_statistic.py
FBegiello: Then I'd stay with `type` for the `workout `table as this is in line with what you're writing, but change `type` in `workout_statistic` to `data_type` or something like that  - when working down the line with complex analytics will be easier to handle and avoid confusion in my opinion

Issue #15, opened 2025-11-19
UserCreate schema allows client-controlled id and created_at


bartmichalak:
## Problem
The `POST /api/v1/users` endpoint accepts `id` and `created_at` in the request body, allowing clients to override server-generated values. This violates REST API best practices and creates security risks.

**Current schema:**
```json
{
  "UserCreate": {
    "properties": {
      "id": {"type": "string", "format": "uuid"},
      "created_at": {"type": "string", "format": "date-time"}
    }
  }
}
```

## Issues

**1. Data Integrity**
- Client can set `created_at` to any date (past or future)
- Breaks audit trails and analytics

**2. Violates REST Principles**
- Server should always control resource identifiers and timestamps
- Inconsistent with industry standards 

## Recommended Fix

**Remove these fields from UserCreate:**
```json
{
  "UserCreate": {
    "properties": {},
    "type": "object"
  }
}
```

**Request:**
```json
POST /api/v1/users
{}
```

**Response:**
```json
{
  "id": "server-generated-uuid",
  "created_at": "2025-11-19T12:00:00Z"
}
```

## Optional Enhancement
If client needs to store user information or mapping to their system, add meaningful fields:
```json
{
  "UserCreate": {
    "properties": {
      "first_name": {"type": "string"},
      "last_name": {"type": "string"},
      "email": {"type": "string", "format": "email"},
      "client_user_id": {"type": "string"},
      "metadata": {"type": "object"}
    }
  }
}
```

All fields optional, allowing developers to store:
- **first_name/last_name**: User's name for display purposes
- **email**: Contact information (not used for authentication)
- **client_user_id**: Developer's internal user ID for mapping
- **metadata**: Custom key-value pairs for additional data

This should be fixed before MVP/public release as it's a fundamental API design issue.

PR #46, opened 2025-11-25
Remove Client-Controlled Fields from User Creation & Add User Profile Fields

bartmichalak:
## Summary

This PR removes client-controlled fields (`id` and `created_at`) from user creation and adds optional user profile fields. The service layer now enforces server-side generation of identifiers and timestamps, improving security and data integrity.

## High-Level Architectural Changes

**Security hardening**: The `UserCreate` schema no longer accepts `id` or `created_at`, preventing clients from controlling these fields. `UserService` overrides the base `create()` method to generate these server-side, ensuring consistent, secure user creation.

**Data model extension**: The User model now includes optional profile fields (`first_name`, `last_name`, `email`) and a required `client_user_id` for external system integration. This supports richer user data while maintaining backward compatibility through nullable fields.

**Schema alignment**: All user schemas (`UserRead`, `UserCreate`, `UserUpdate`) are updated to reflect the new fields, ensuring API consistency across read, create, and update operations.

## Why

Allowing clients to control `id` and `created_at` creates security risks: ID collisions, timestamp manipulation, and potential enumeration attacks. Server-side generation ensures data integrity and prevents these vulnerabilities.

The addition of profile fields enables better user experience and integration capabilities. The `client_user_id` field allows mapping users to external systems without exposing internal UUIDs, while optional profile fields support progressive data collection without breaking existing functionality.


bartmichalak:
Resolves https://github.com/the-momentum/open-wearables/issues/15

backend/app/models/user.py
```diff
@@ -12,6 +13,12 @@ class User(BaseDbModel):
     id: Mapped[PrimaryKey[UUID]]
     created_at: Mapped[datetime_tz]
 
+    # Optional user information fields
+    first_name: Mapped[Optional[str_100]] = None
```
KaliszS: Optional ... = None is redundant
It should have been: `first_name: Mapped[str_100 | None]`. The same applies for the rest of the new fields

backend/app/services/user_service.py
```diff
@@ -15,5 +18,20 @@ def __init__(self, log: Logger, **kwargs):
...
         )
 
+    def create(self, db_session: DbSession, creator: UserCreate) -> User:
+        """Create a user with server-generated id and created_at."""
+        creation_data = creator.model_dump()
+        # Server generates id and created_at
+        creation_data["id"] = uuid4()
+        creation_data["created_at"] = datetime.now(timezone.utc)
+        
+        creation = self.crud.model(**creation_data)
```
KaliszS: It's a BIG NO :P W don't operate on database in services, we have repositories for that. Please, don't mix those layers.

backend/app/schemas/user.py
```diff
@@ -1,18 +1,27 @@
...
+    first_name: str | None = None
+    last_name: str | None = None
+    email: EmailStr | None = None
+    client_user_id: str
 
 
 class UserCreate(BaseModel):
-    id: UUID = Field(default_factory=uuid4)
-    created_at: datetime = Field(default_factory=lambda: datetime.now(timezone.utc))
+    first_name: str | None = Field(None, max_length=100)
```
KaliszS, line 19: The better solution is to have two schemas - one for creating fields lke  username etc and second one for DB bridge with autogenerated id and created_at. We should use both.

I can handle it later (in seprate PR maybe)

PR #48, opened 2025-11-25
Fix Celery task registration for dummy_task

bartmichalak:
## Summary
Fixes Celery task registration so dummy_task is discoverable and registered. Resolves Sentry issue `OPEN-WEARABLES-BACKEND-1` where Celery workers received unregistered task errors.

## Changes
- Fixed autodiscovery path: changed from task path `app.integrations.celery.tasks.dummy_task` to package path `app.integrations.celery.tasks` in `autodiscover_tasks()`
- Added explicit task import: imported dummy_task in core.py to ensure registration at module load
- Set explicit task name: added name="app.integrations.celery.tasks.dummy_task" to the @shared_task decorator to match the beat schedule configuration

## Why
`autodiscover_tasks()` was given a task path instead of a package path, so tasks weren't discovered. The task name wasn't explicitly set, causing a mismatch with the beat schedule. These changes ensure the task is registered with the correct name and discoverable by Celery workers.

## Impact
- Eliminates unregistered task errors in Celery workers
- Ensures scheduled tasks execute correctly
- Improves reliability of the Celery task system

### DIFF
--- a/backend/app/integrations/celery/core.py
+++ b/backend/app/integrations/celery/core.py
@@ -4,6 +4,9 @@ from celery.schedules import crontab
 
 from app.config import settings
 
+# Import tasks to ensure they're registered
+from app.integrations.celery.tasks import dummy_task  # noqa: F401
+
 
 def create_celery() -> Celery:
     celery_app: Celery = current_celery_app  # type: ignore[assignment]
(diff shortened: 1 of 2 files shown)


czajkub:
The branch with xml import of apple health data (#6) uses celery tasks and the paths are configured correctly there, so the issue should be fixed with merging the pull request.

backend/app/integrations/celery/core.py
```diff
@@ -4,6 +4,9 @@
 
 from app.config import settings
 
+# Import tasks to ensure they're registered
+from app.integrations.celery.tasks import dummy_task  # noqa: F401
```
KaliszS, line 8: Have you tested if that's required? Because registration is being done using direct path to the tasks (and the path is just a string in our code).

PR #55, opened 2025-11-26
fix: Backend auth errors and frontend API integration

farce1:
## Summary

Fixes several backend errors encountered during auth testing and updates frontend to properly integrate with the API.

### Backend Fixes

- **Timestamp defaults**: Added `datetime_tz_now` and `datetime_tz_updated` types with `insert_default` to auto-set `created_at`/`updated_at` fields, fixing `NotNullViolation` errors on Developer model
- **Exception handler**: Fixed `handle_exception()` call using keyword arg instead of positional arg (`err_msg=` → positional)
- **Validation error handling**: Fixed `KeyError: 'ctx'` by using `.get()` for optional fields in `RequestValidationError` handler

### Frontend Fixes

- Added `postForm()` method to API client for OAuth2 form data
- Updated login to send `username`/`password` as form data (required by fastapi-users)
- Added `/api` prefix to all API endpoints
- Added registration, forgot-password, and reset-password routes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,8 +14,8 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
-lib64/
+/lib/
+/lib64/
 parts/
 sdist/
 var/
--- a/frontend/src/lib/validation/auth.schemas.ts
+++ b/frontend/src/lib/validation/auth.schemas.ts
@@ -1,0 +1,47 @@
+import { z } from 'zod';
+
+export const emailSchema = z
+  .string()
+  .min(1, 'Email is required')
+  .email('Please enter a valid email address');
+
+export const passwordSchema = z
+  .string()
+  .min(8, 'Password must be at least 8 characters')
+  .refine(
+    (val) => {
+      const hasUpperAndLower = /[A-Z]/.test(val) && /[a-z]/.test(val);
+      const hasNumber = /[0-9]/.test(val);
+      return hasUpperAndLower || hasNumber;
+    },
+    { message: 'Password must contain mixed case or a number' }
+  );
+
+export const registerSchema = z
+  .object({
+    email: emailSchema,
+    password: passwordSchema,
+    confirmPassword: z.string().min(1, 'Please confirm your password'),
+  })
+  .refine((data) => data.password === data.confirmPassword, {
+    message: 'Passwords do not match',
+    path: ['confirmPassword'],
+  });
+
+export const forgotPasswordSchema = z.object({
+  email: emailSchema,
+});
+
+export const resetPasswordSchema = z
+  .object({
+    password: passwordSchema,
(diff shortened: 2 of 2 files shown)


backend/app/mappings.py
```diff
@@ -24,6 +24,8 @@
 
 # Custom types
 datetime_tz = Annotated[datetime, mapped_column(DateTime(timezone=True))]
+datetime_tz_now = Annotated[datetime, mapped_column(DateTime(timezone=True), insert_default=lambda: datetime.now(timezone.utc))]
```
KaliszS: We definitelly don't want to do that on the models side, that's something for pydantic schemas. it's because python code operations are still cheaper than database operations. Of course it won't be significat difference, but we want to keep consistency anyway.

Also changing models without regenerating migrations didn't change anything at database anyway.

backend/app/models/developer.py
```diff
@@ -2,11 +2,11 @@
...
-from app.mappings import datetime_tz
+from app.mappings import datetime_tz_now, datetime_tz_updated
 
 
 class Developer(BaseDbModel, SQLAlchemyBaseUserTableUUID):
     """Admin of the portal model"""
 
-    created_at: Mapped[datetime_tz]
-    updated_at: Mapped[datetime_tz]
+    created_at: Mapped[datetime_tz_now]
```
KaliszS: As I said in previous comment - we don't want to implement this that way.

frontend/src/lib/api/config.ts
```diff
@@ -6,31 +6,33 @@ export const API_CONFIG = {
...
-  heartRate: '/v1/heart-rate',
-  sleep: '/v1/sleep',
-  workouts: '/v1/workouts',
+  heartRate: '/api/v1/heart-rate',
+  sleep: '/api/v1/sleep',
+  workouts: '/api/v1/workouts',
 
-  apiKeys: '/v1/api-keys',
-  apiKeyDetail: (id: string) => `/v1/api-keys/${id}`,
+  apiKeys: '/api/v1/api-keys',
```
KaliszS: api_keys use /developer endpoint now: https://github.com/the-momentum/open-wearables/blob/main/backend/app/api/routes/v1/__init__.py

KaliszS (changes requested):

backend/app/mappings.py
```diff
@@ -1,10 +1,10 @@
-from datetime import datetime
+from datetime import datetime, timezone
```
KaliszS: we still have leftovers in imports :D

KaliszS (approved):

backend/app/mappings.py
farce1: true, sorry about that. after linter fix on backend it wont happen

Issue #54, opened 2025-11-25
Fix linter issues


bartmichalak:
Ruff linter is reporting 394 code quality violations across the codebase, with 283 automatically fixable. 

>Found 394 errors.
[*] 283 fixable with the `--fix` option (17 hidden fixes can be enabled with the `--unsafe-fixes` option).

Let's tackle them and not merge code with linter issues anymore :)

PR #57, opened 2025-11-26
Fix most linter issues

czajkub:
This PR fixes most of the linter issues, leaving behind:

return types / type annotations in utils/ (e.g. mappings_meta.py)
garmin services / webhooks
other minor, not easily fixable warnings
fixes https://github.com/the-momentum/open-wearables/issues/54

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -64,6 +64,8 @@ class Settings(BaseSettings):
     redis_host: str = "localhost"
     redis_port: int = 6379
     redis_db: int = 0
+    redis_password: SecretStr | None = None
+    redis_username: str | None = None  # Redis 6.0+ ACL
 
     # SUUNTO OAUTH SETTINGS
     suunto_client_id: str | None = None


backend/app/api/routes/v1/api_keys.py
```diff
@@ -1,16 +1,18 @@
...
+from app.services import DeveloperDep, api_key_service
 
 router = APIRouter()
 
 
 @router.get("/api-keys")
 async def list_api_keys(db: DbSession, _developer: DeveloperDep) -> list[ApiKeyRead]:
     """List all API keys."""
-    return api_key_service.list_api_keys(db)
+    return [ApiKeyRead.model_validate(key) for key in api_key_service.list_api_keys(db)]
```
czajkub, line 15: methods like `list_api_keys` return an object of class `ApiKey`, meanwhile the type annotation is `ApiKeyRead`, so we validate the model from the return of the called method

czajkub: same thing happens with `User`

backend/app/schemas/api_key.py
```diff
@@ -1,10 +1,12 @@
 from datetime import datetime, timezone
 from uuid import UUID
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class ApiKeyRead(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
```
czajkub, line 8: for model validation mentioned before

backend/app/schemas/apple/auto_export/json_schemas.py
```diff
@@ -1,3 +1,5 @@
+# ruff: noqa: N815
```
czajkub, line 1: N815 is mixed case variable names instead of snake case (models use mixed case)

backend/app/schemas/__init__.py
czajkub, line 1: most of these imports are gonna be gone with the unified model anyways

backend/app/config.py
```diff
@@ -131,19 +128,6 @@ def db_uri(self) -> str:
             f"@{self.db_host}:{self.db_port}/{self.db_name}"
         )
 
-    @property
-    def redis_url(self) -> str:
```
KaliszS, line 135: why it got removed?

backend/app/services/oauth_service.py
```diff
@@ -32,9 +32,12 @@ class OAuthService:
     def __init__(self, log: Logger):
         self.logger = log
         self.repository = UserConnectionRepository()
-        
-        # Create Redis client from URL
-        self.redis_client = redis.from_url(settings.redis_url, decode_responses=True)
+        self.redis_client = redis.Redis(
```
KaliszS: Ok, now I see, but still - why not to just hold leave it inside settings, the same way we store db_uri as proeprty? of course old form had some legacy options we don't need to maintina, but that simple Redis() class copuld be imported entirely from settings. What do you think?

backend/app/config.py
KaliszS: Check the next comment :)

.pre-commit-config.yaml
```diff
@@ -1,24 +1,33 @@
...
       # Run the formatter
       - id: ruff-format
+        files: ^backend/.*\.py$
 
   - repo: local
     hooks:
       - id: ty
         name: ty check
-        entry: .venv/bin/ty check .
+        entry: sh -c 'cd backend && .venv/bin/ty check .'
```
czajkub, line 19: its done this way, because for some reason just running `backend/.venv/bin/ty` includes ignored files, probably because while it's in `backend/` it doesn't recognise the `backend/*` regex?

czajkub: also this way, it can be run both from the root directory and from `backend/`

KaliszS (approved):

Issue #53, opened 2025-11-25
Create default developer account with admin credentials


bartmichalak:
Create a seeded developer account with credentials `admin@admin.com` / `secret` that is automatically available after running docker compose up to allow developers to immediately access the portal without reaching registration endpoint.

PR #73, opened 2025-12-01
Seed admin account. Use docker compose exec to run db operations in Makefile

bartmichalak:
**Summary:**
- Seed admin account 
- Move database commands from local uv to docker

https://github.com/user-attachments/assets/db9a8b16-c120-48f7-a6c7-7d44fdbb52d8

Resolves https://github.com/the-momentum/open-wearables/issues/53

### DIFF
--- a/backend/scripts/init/seed_admin.py
+++ b/backend/scripts/init/seed_admin.py
@@ -11,8 +11,7 @@ ADMIN_PASSWORD = "secret123"
 
 def seed_admin() -> None:
     """Create default admin developer if it doesn't exist."""
-    db = SessionLocal()
-    try:
+    with SessionLocal() as db:
         existing = developer_service.crud.get_all(
             db,
             filters={"email": ADMIN_EMAIL},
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,15 +28,17 @@ test:	## Run the tests.
 	cd backend && uv run pytest -v --cov=app
 
 migrate:  ## Apply all migrations
-	$(ALEMBIC_CMD) upgrade head
+	$(DOCKER_EXEC) $(ALEMBIC_CMD) upgrade head
+
+init:  ## Seed sample data
+	$(DOCKER_EXEC) uv run python scripts/init/seed_admin.py
```
bartmichalak, line 34: The way I see it, as we add more seed types, we will need more commands, such as:
$(DOCKER_EXEC) uv run python scripts/init/seed_wokrouts.py

As an alternative, we could keep everything in one file, but I think separating it would be more readable

backend/scripts/init/seed_admin.py
```diff
@@ -0,0 +1,37 @@
...
+            print(f"Admin developer {ADMIN_EMAIL} already exists, skipping.")
+            return
+
+        developer_service.register(db, DeveloperCreate(email=ADMIN_EMAIL, password=ADMIN_PASSWORD))
+        print(f"✓ Created default admin developer: {ADMIN_EMAIL} / {ADMIN_PASSWORD}")
+    except Exception as exc:
+        db.rollback()
+        raise exc
+    finally:
+        db.close()
```
KaliszS: Maybe we can switch to with context manager? It would be more comfortable to use without explicit rollback or close.

bartmichalak: Good call, changed that!

PR #78, opened 2025-12-01
Add activity data seeding script and update Makefile and dependencies

### LINKED ISSUE #77: Create sample activity data
In similar way as in #53.

bartmichalak:
Summary:
- Creates sample users and workouts that belongs to them. 
- Using Faker to set some attributes 

image

Resolves #77

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -24,7 +24,6 @@ dependencies = [
     "requests>=2.32.5",
     "bcrypt>=5.0.0",
     "isodate>=0.7.2",
-    "faker>=33.1.0",
 ]
 
 [dependency-groups]
@@ -37,6 +36,7 @@ dev = [
     "pytest>=8.4.1",
     "pytest-asyncio>=1.0.0",
     "pytest-cov>=6.2.1",
+    "faker>=33.1.0",
 ]
 
 [tool.ruff]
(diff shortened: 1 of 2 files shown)


backend/pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "requests>=2.32.5",
     "bcrypt>=5.0.0",
     "isodate>=0.7.2",
+    "faker>=33.1.0",
```
KaliszS: I'm just not sure if that would belong to main dependencies group. What do you think?

backend/scripts/init/seed_activity_data.py
```diff
@@ -0,0 +1,112 @@
...
+            db.commit()
+            print(f"  ✓ Completed 100 workouts for user {user_num}\n")
+
+        print(f"✓ Successfully created {users_created} users with {workouts_created} total workouts")
+    except Exception as exc:
+        db.rollback()
+        print(f"✗ Error occurred: {exc}")
+        raise exc
+    finally:
+        db.close()
```
KaliszS: I would also use with context manager here.

backend/pyproject.toml
bartmichalak: Good call, I've moved this library to `dev` group and added `uv sync --group dev` to `init` command in `Makefile`. WDYT?

Issue #79, opened 2025-12-02
Add user/{id}/connections endpoint


bartmichalak:
Create backend endpoint that will return all user's connections.

PR #80, opened 2025-12-02
Add /users/{id}/connections endpoint

bartmichalak:
Resolves https://github.com/the-momentum/open-wearables/issues/79

### DIFF
--- a/backend/app/models/user_connection.py
+++ b/backend/app/models/user_connection.py
@@ -1,6 +1,6 @@
 from uuid import UUID
 
-from sqlalchemy import Enum, Index, UniqueConstraint
+from sqlalchemy import Index, UniqueConstraint
 from sqlalchemy.orm import Mapped, mapped_column
 
 from app.database import BaseDbModel
@@ -36,7 +36,7 @@ class UserConnection(BaseDbModel):
     scope: Mapped[str | None]
 
     # Metadata
-    status: Mapped[ConnectionStatus] = mapped_column(Enum(ConnectionStatus, name="connection_status"))
+    status: Mapped[ConnectionStatus]
     last_synced_at: Mapped[datetime_tz | None]
     created_at: Mapped[datetime_tz]
     updated_at: Mapped[datetime_tz]
--- a/backend/app/database.py
+++ b/backend/app/database.py
@@ -20,6 +20,7 @@ from sqlalchemy.orm import (
 
 from app.config import settings
 from app.mappings import email, str_10, str_50, str_64, str_100, str_255
+from app.schemas.oauth import ConnectionStatus
 from app.utils.mappings_meta import AutoRelMeta
 
 engine = create_engine(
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -16,6 +16,9 @@ run:	## Runs the envionment in detached mode
 up:	## Runs the non-detached environment
 	$(DOCKER_COMMAND) up --force-recreate
 
+watch:	## Runs the environment with hot-reload
+	$(DOCKER_COMMAND) --watch
```
KaliszS, line 20: It's just `watch`, not `--watch`.

backend/app/api/routes/v1/connections.py
```diff
@@ -0,0 +1,22 @@
...
+
+from fastapi import APIRouter
+
+from app.database import DbSession
+from app.repositories import UserConnectionRepository
+from app.schemas import UserConnectionRead
+from app.services import ApiKeyDep
+
+router = APIRouter()
+connection_repo = UserConnectionRepository()
```
KaliszS, line 11: I think it's better (and consistent with architecture) to implement corresponding method in user connection service (if you will give Cursor an example of user_service.py and ask to create service base on user connection repository with implementing one custom method - get by user id - it should handle it perfectly)

KaliszS: The reason to split it is - in services we can later on implement more business logic, like validation and error handling.

backend/app/api/routes/v1/connections.py
```diff
@@ -0,0 +1,22 @@
...
+
+@router.get("/users/{user_id}/connections", response_model=list[UserConnectionRead])
+async def get_connections_endpoint(
+    user_id: str,
+    db: DbSession,
+    _api_key: ApiKeyDep,
+):
+    """Get all connections for a user."""
+    connections = connection_repo.get_by_user_id(db, UUID(user_id))
+    return [UserConnectionRead.model_validate(conn) for conn in connections]
```
KaliszS, line 22: It should be tested first, but I believe `model_response`would be enough here, so we don't need to use `model_validate`. It would be just `return connections`

backend/app/models/user_connection.py
```diff
@@ -35,7 +36,7 @@ class UserConnection(BaseDbModel):
     scope: Mapped[str | None]
 
     # Metadata
-    status: Mapped[str_64]  # 'active', 'revoked', 'expired'
+    status: Mapped[ConnectionStatus] = mapped_column(Enum(ConnectionStatus, name="connection_status"))
```
KaliszS: - It require to generate new migration.
- `status: Mapped[ConnectionStatus]` should be enough (if migration won't be correctly generated, just let me know, I will handle this on database.py side)

Makefile
bartmichalak: Are you sure? 
`watch` doesn't work in my local env, also Docker documents this as `docker compose up --watch`:
https://docs.docker.com/compose/how-tos/file-watch/

KaliszS: Yeah Im using docker compose watch all the time, but I will read more about that flag option later nad update if that will be needed.

backend/app/api/routes/v1/connections.py
KaliszS: to do later

PR #82, opened 2025-12-02
Saving workout data from providers to database

czajkub:
A new endpoint
 `/api/v1/vendors/{provider}/users/{user_id}/sync` is created, allowing for manually synchronising data from a provider. It's adapted from the import_service added for each provider from previous PR, also abstract class for base workout service is updated for providers.

### DIFF
--- a/backend/app/services/providers/polar/workouts.py
+++ b/backend/app/services/providers/polar/workouts.py
@@ -49,6 +49,20 @@ class PolarWorkouts(BaseWorkoutsTemplate):
         route = kwargs.get("route", False)
         return self.get_exercise_detail(db, user_id, workout_id, samples, zones, route)
 
+    def _extract_dates_with_offset(
+        self,
+        start_time: str,
+        start_time_utc_offset: int,
+        duration: str,
+    ) -> tuple[datetime, datetime]:
+        """Extract start and end dates from timestamps with UTC offset."""
+        start_date = isodate.parse_datetime(start_time)
+        offset = timedelta(minutes=start_time_utc_offset)
+        start_date = start_date + offset
+        duration_td = isodate.parse_duration(duration)
+        end_date = start_date + duration_td
+        return start_date, end_date
+
     def _normalize_workout(
         self,
         raw_workout: PolarExerciseJSON,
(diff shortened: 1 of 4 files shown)


backend/app/services/providers/polar/workouts.py
```diff
@@ -41,23 +49,118 @@ def get_workout_detail_from_api(self, db: DbSession, user_id: UUID, workout_id:
...
+        raw_workout: PolarExerciseJSON,
+        user_id: UUID,
+    ) -> WorkoutCreate:
         """Normalize Polar exercise to WorkoutCreate."""
-        # Placeholder mapping
-        duration_str = raw_workout.get("duration", "PT0S").replace("PT", "").replace("S", "")
-        duration_val = Decimal(duration_str) if duration_str else Decimal(0)
+        workout_id = uuid4()
+
+        start_date: datetime = isodate.parse_datetime(raw_workout.start_time)
```
czajkub: maybe extract the date parsing to another method? since its used in both workout and statistic normalizing

KaliszS: Sounds like a good idea. We can add this to base_workouts template (acepting input_time paraemeter, so we could call it with raw_workout.start_time in provider implementation).

backend/app/api/routes/v1/sync_data.py
```diff
@@ -0,0 +1,80 @@
...
+    strategy = factory.get_provider(provider.value)
+
+    if not strategy.workouts:
+        raise HTTPException(
+            status_code=status.HTTP_501_NOT_IMPLEMENTED,
+            detail=f"Provider '{provider.value}' does not support workouts",
+        )
+
+    # Collect all parameters
+    params = {
```
bartmichalak, line 67: I'm wondering if this endpoint should even accept such parameters? Shouldn't its operational logic be more like it fetches all data since the last sync (well, the logic can be a bit more sophisticated, we can also check data from before the last sync, in case someone added an entry backdated to the past)

IMO consumer of this endpoint shouldn't really have control over this, the logic should be on the backend side (the consumer only cares that a given user has the latest data state)

backend/app/api/routes/v1/sync_data.py
```diff
@@ -0,0 +1,80 @@
...
+        int,
+        Query(description="Maximum number of workouts (Suunto: max 100)", le=100),
+    ] = 50,
+    offset: Annotated[int, Query(description="Offset for pagination (Suunto only)")] = 0,
+    filter_by_modification_time: Annotated[
+        bool,
+        Query(description="Filter by modification time instead of creation time (Suunto only)"),
+    ] = True,
+    # Polar-specific parameters
+    samples: Annotated[bool, Query(description="Synchronize sample data (Polar only)")] = False,
```
bartmichalak, line 36: are we doing anything with this data currently (writing to the database)?

Issue #83, opened 2025-12-02
Add endpoint with available providers


bartmichalak:
Backend should expose an endpoint that returns the available providers to handle this FE view:

<img width="847" height="787" alt="Image" src="https://github.com/user-attachments/assets/494c6420-422a-469a-a004-143bc6911f75" />

introduced in this PR
https://github.com/the-momentum/open-wearables/pull/72

PR #84, opened 2025-12-02
Added providers settings

### LINKED ISSUE #86: Provider icons should be provided by the backend API
Provider icons (introduced [here](https://github.com/the-momentum/open-wearables/pull/72/files)) should be returned by the backend instead of being handled on the frontend (as the list of available providers is handled by the backend). 
This would centralise icon management and ensure consistency. 

It requires both frontend and backend changes.

KaliszS:
Resolves #86 
Resolves https://github.com/the-momentum/open-wearables/issues/83


backend/app/api/routes/v1/oauth.py
```diff
@@ -78,3 +88,61 @@ async def oauth_success(
         "user_id": user_id,
         "provider": provider,
     }
+
+
+@router.get("/available_providers")
```
bartmichalak: I'm wondering if the response of these two endpoints 
- available_providers 
- providers 

shouldn't be the same (or maybe it should even be a single endpoint with the ability to filter by is_enabled)

Because right now, when we want to display a list of all providers on the frontend and mark those that are enabled, the frontend will have to hit two endpoints

backend/app/api/routes/v1/oauth.py
```diff
@@ -78,3 +88,61 @@ async def oauth_success(
...
+async def get_available_providers(
```
bartmichalak: Also, maybe enabled_providers? We even have an is_enabled column, so this name seems to fit better

PR #85, opened 2025-12-02
Add sync tasks in celery

czajkub:
Create two tasks for synchronizing vendor data in celery:
- sync_vendor_data - takes user_id, start and end date - synchronizes for all active providers for given user
- sync_all_users - runs on an 1-hour crontab, calls the sync_vendor_data task for all users with active connections within that hour

### DIFF
--- a/backend/app/schemas/__init__.py
+++ b/backend/app/schemas/__init__.py
@@ -64,6 +64,11 @@ from .suunto.workout_import import (
 from .suunto.workout_import import (
     WorkoutJSON as SuuntoWorkoutJSON,
 )
+from .sync import (
+    ProviderSyncResult,
+    SyncAllUsersResult,
+    SyncVendorDataResult,
+)
 from .user import (
     UserCreate,
     UserCreateInternal,
@@ -81,11 +86,6 @@ from .workout_statistics import (
     WorkoutStatisticCreate,
     WorkoutStatisticResponse,
     WorkoutStatisticUpdate,
-)
-from .sync import (
-    ProviderSyncResult,
-    SyncAllUsersResult,
-    SyncVendorDataResult,
 )
 
 __all__ = [


backend/app/integrations/celery/tasks/sync_vendor_data_task.py
```diff
@@ -0,0 +1,177 @@
...
+    factory = ProviderFactory()
+    user_connection_repo = UserConnectionRepository()
+
+    results: dict[str, Any] = {
+        "user_id": user_id,
+        "start_date": start_date,
+        "end_date": end_date,
+        "providers_synced": {},
+        "errors": {},
+    }
```
FBegiello: This could be a pydantic object for full type checking and consistency - can be addressed later during a refactor

backend/app/integrations/celery/tasks/periodic_sync_task.py
```diff
@@ -0,0 +1,29 @@
...
+
+    with SessionLocal() as db:
+        user_ids = user_connection_repo.get_all_active_users(db)
+
+        logger.info(f"[sync_all_users] Found {len(user_ids)} users with active connections")
+
+        for user_id in user_ids:
+            sync_vendor_data.delay(str(user_id), start_date, end_date)
+
+        return {"users_synced": len(user_ids)}
```
FBegiello: Minor thing but would be clearer as ` "users_scheduled_for_syncinc"` or something to that effect - we just registered tasks in celery

backend/app/api/routes/v1/oauth.py
```diff
@@ -63,6 +64,9 @@ async def oauth_callback(
     # Redirect to success page or developer's redirect_uri
     redirect_url = oauth_state.redirect_uri or f"/oauth/success?provider={provider.value}&user_id={oauth_state.user_id}"
 
+    # schedule sync task
+    sync_vendor_data.delay(str(oauth_state.user_id), start_date=None, end_date=None)
```
bartmichalak: I think we should run a task dedicated for specific provider here, but for now it's nothing critical, it can be adressed in a separate PR (please create issue for that)

czajkub: Yes, thats a good idea, probably should add a list of providers to the task so we can only sync one provider instead of every one for the user on new connections

backend/app/integrations/celery/tasks/sync_vendor_data_task.py
czajkub: sure, ill add an issue for that

backend/app/integrations/celery/tasks/periodic_sync_task.py
czajkub: yes, definitely makes more sense

Issue #24, opened 2025-11-19
Duplicate detection strategy


bartmichalak:
## Problem Statement

When integrating health data from multiple wearable providers, duplicate data entries are inevitable and problematic. Users often:

- Wear multiple devices simultaneously (e.g., Apple Watch + Oura Ring)
- Have overlapping data from different sources (e.g., steps from phone + watch)
- Experience sync delays causing the same measurement to appear multiple times
- Switch between devices throughout the day

Without proper duplicate detection, this leads to:
- Inaccurate aggregations - Double counting steps, calories, heart rate measurements
- Misleading analytics - Sleep data appearing twice, skewing averages
- Poor developer experience - Developers must implement their own deduplication logic
- Loss of trust - End-users see inflated metrics and question data accuracy

This issue addresses the strategy for detecting and handling duplicates across all providers. Implementation will likely require provider-specific sub-tasks.

## Proposed Solution

Design and document a multi-layered duplicate detection strategy:

### Layer 1: Exact Duplicate Detection (Mandatory)

Primary key: (user_id, provider, data_type, timestamp, value)
Detect identical measurements from the same provider
Most common case: API retry, webhook replay

### Layer 2: Cross-Provider Fuzzy Matching (High Priority)

Time window matching (e.g., ±5 minutes for heart rate, ±30 minutes for sleep)
Value similarity threshold (e.g., heart rate within ±3 bpm)
Provider priority ranking (which source is more reliable?)
Different strategies per data type (steps vs. sleep vs. heart rate)

### Layer 3: Same-Provider Temporal Deduplication (Medium Priority)

Handle cases where provider sends corrected/updated data
Keep latest value, mark old as superseded
Maintain audit trail
Key 

## Design Decisions Needed:

**When to deduplicate?**
At ingestion time (real-time)
At query time (on-demand)
Background job (periodic cleanup)
**Recommendation:** Layer 1 at ingestion, Layer 2 at query time for MVP

**How to handle conflicts?**
Provider priority (e.g., Apple Health > Fitbit for steps)
Data quality score (based on device type, sampling rate)
User preference (let user choose preferred source)
Keep all + flag duplicates (transparency)

**Storage strategy:**
- Mark duplicates with is_duplicate: true + duplicate_of: <id>
- Separate table for duplicate relationships
- Materialized view for deduplicated data
- **Recommendation:** Flagging approach for MVP (easier to debug)

PR #88, opened 2025-12-02
Data model refactor - Draft for activity and time series storing

FBegiello:
A general refactor of data models related to the core raw data - designed for generic storage of events (workouts, sleep events etc.) in a unified format, with raw time series data like HR or steps stored in separate tables, ready to be processed and treated.

The rework follows the architecture outlined below
Data model

`HealthRecord ` is meant as a base table, holding only general references (like start/end times of the event) and aggregates (like min/max/avg). Supporting raw time series tables like `HeartRateSample` or `StepsSample` are meant to store all the recorded datapoints, even those not directly related to an event. Event-related data, like specific datapoints in a workout, are to be accessed via queries litered by timestamps from those time series tables.

Changelog:
- Replace the workout/workout_statistic schema with a new health_record table built around explicit min/max/avg metrics for heart rate and steps. Introduced new HeartRateSample/StepSample tables without FK ties to users or records, updated all models, repositories, schemas, and services accordingly, and wrote an Alembic migration to create/drop the new structures and migrate legacy data.
- Reworked every ingestion path (HealthKit XML/auto-export, Garmin, Polar, Suunto, seeding, etc.) so they populate the new metric fields and emit standalone time‑series samples keyed only by device/timestamp; simplified the TimeSeries API and service layer to filter by device + time window rather than record/user IDs.
- Updated backend routes and the React client: /users/{id}/workouts now returns the new metric fields, /users/{id}/heart-rate streams raw samples scoped by device, and the frontend service/hooks/types were aligned with the new payloads. 

NOTE - Some changes has been made to the FE modules, those might require additional look

Resolves https://github.com/the-momentum/open-wearables/issues/96


backend/app/schemas/health_record.py
```diff
@@ -0,0 +1,100 @@
...
+from decimal import Decimal
+from typing import Literal
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+from app.schemas.common import BaseQueryParams
+
+
+class HealthRecordCreate(BaseModel):
```
bartmichalak: I'm not sure if a single model for those aggregates:
- sleep
- workouts
- daily activity (simply a summary of the day)

is definitely a good idea. Training will have different data than sleep (and this is currently missing from this model), for example:
- max_speed
- max_watts
- moving time
- total_elevation_gain
- average_speed
- average_watts
- elev_high
- elev_low
- others

However, separating these aggregates seems like a better idea to me

backend/app/api/routes/v1/heart_rate.py
```diff
@@ -1,17 +1,22 @@
...
+@router.get("/users/{user_id}/heart-rate", response_model=list[HeartRateSampleResponse])
 async def get_heart_rate_endpoint(
     user_id: str,
     db: DbSession,
     _api_key: ApiKeyDep,
+    query_params: Annotated[TimeSeriesQueryParams, Depends()],
 ):
     """Get heart rate data with filtering, sorting, and pagination."""
-    return await workout_statistic_service.get_user_heart_rate_statistics(db, user_id)
+    if not query_params.device_id:
```
czajkub: Why do we only want to retrieve heart rate data from one device? Wouldn't this become an issue when a user switches devices / has multiple devices connected? I assume this serves the purpose of not duplicating data, but shouldnt this logic be moved to services?

backend/app/api/routes/v1/heart_rate.py
```diff
@@ -1,17 +1,22 @@
...
+        raise HTTPException(status_code=400, detail="device_id query parameter is required")
+    return await workout_statistic_service.get_user_heart_rate_series(db, user_id, query_params)
```
czajkub: shouldn't we rename this to time_series_service or similar?

backend/app/models/health_record.py
```diff
@@ -0,0 +1,40 @@
...
+
+    duration_seconds: Mapped[numeric_10_3 | None] = None
+
+    start_datetime: Mapped[datetime_tz]
+    end_datetime: Mapped[datetime_tz]
+
+    heart_rate_min: Mapped[numeric_10_3 | None] = None
+    heart_rate_max: Mapped[numeric_10_3 | None] = None
+    heart_rate_avg: Mapped[numeric_10_3 | None] = None
+    steps_min: Mapped[numeric_10_3 | None] = None
```
czajkub: Why do we have min, max, avg for steps? Shouldn't this be a single value of the summed up number of steps?

backend/app/models/heart_rate_sample.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class HeartRateSample(BaseDbModel):
+    __tablename__ = "heart_rate_sample"
+    __table_args__ = (
+        Index("idx_hr_sample_device_time", "device_id", "recorded_at"),
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    device_id: Mapped[str_100 | None] = None
```
czajkub: If device_id here is nullable, wouldn't heart rate data from devices without a provided id be non-retrievable from the `/heart-rate` endpoint, since it requires a device id?

czajkub: Also, is there a reason why we are adding the `source_name` field to `HealthRecord` but not here?

backend/app/repositories/data_point_series_repository.py
czajkub, line 1: Will we be adding an endpoint for steps in the future, similarly to the `heart-rate` endpoint?

backend/app/schemas/time_series.py
```diff
@@ -0,0 +1,41 @@
+from __future__ import annotations
+
+from datetime import datetime
+from decimal import Decimal
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+
+class TimeSeriesQueryParams(BaseModel):
```
czajkub: Can't we add a `value` parameter here, or something similar? As it stands now, if we hypothetically wanted to filter heart rate data below 180bpm we have no way of doing that via the `/heart-rate` endpoint

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -68,125 +57,169 @@ def _parse_date_fields(self, document: dict[str, Any]) -> dict[str, Any]:
...
-            max=document["value"],
-            avg=document["value"],
-            unit=document["unit"],
-        )
-
-    def _create_workout(self, document: dict[str, Any], user_id: str) -> WorkoutCreate:
+        metric_type = document.get("type", "")
+        value = Decimal(str(document["value"]))
+
+        if "HeartRate" in metric_type:
```
czajkub: These mappings are incorrect, the correct one for heart rate is `HKQuantityTypeIdentifierHeartRate` and for steps `HKQuantityTypeIdentifierStepCount`

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -68,125 +57,169 @@ def _parse_date_fields(self, document: dict[str, Any]) -> dict[str, Any]:
...
+        document: dict[str, Any],
+        user_id: str,
+        metrics: HealthRecordMetrics | None = None,
+    ) -> HealthRecordCreate:
         document = self._parse_date_fields(document)
 
         document["type"] = document.pop("workoutActivityType")
 
-        duration_seconds = (document["endDate"] - document["startDate"]).total_seconds()
+        duration_seconds = Decimal(str((document["endDate"] - document["startDate"]).total_seconds()))
```
czajkub: Do we need to cast it to `str` first? Wouldn't a conversion from float to Decimal work just fine?

backend/app/services/workout_service.py
czajkub, line 1: We should either rename this to `workout_service` or `health_record_service` (imo it should be the latter), since having two names of services is really confusing

backend/app/services/workout_statistic_service.py
czajkub, line 1: same case as above

backend/app/models/event_record_detail.py
```diff
@@ -0,0 +1,29 @@
...
+
+from typing import TYPE_CHECKING
+from uuid import UUID
+
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from app.database import BaseDbModel
+from app.mappings import FKEventRecord, ManyToOne, PrimaryKey, str_64
+
+if TYPE_CHECKING:  # pragma: no cover
```
czajkub: Is this for linter issues with the annotations on line 23? This shouldn't be necessary, whole `models/` directory is excluded in the pre-commit config

backend/app/models/event_record_detail.py
```diff
@@ -0,0 +1,29 @@
...
+    from .event_record import EventRecord
+
+
+class EventRecordDetail(BaseDbModel):
+    """Base polymorphic detail model used by specific aggregates (workout, sleep, etc.)."""
+
+    __tablename__ = "event_record_detail"
+
+    record_id: Mapped[PrimaryKey[UUID]] = mapped_column(FKEventRecord)
```
czajkub: Instead of doing it this way, I'd suggest either changing the `FKEventRecord` in `mappings.py` to also be a primary key, or create a new mapping that does that instead of changing the old one

backend/app/models/event_record_detail.py
```diff
@@ -0,0 +1,29 @@
...
+    detail_type: Mapped[str_64]
+
+    record: Mapped[ManyToOne["EventRecord"]] = relationship(back_populates="detail")
```
czajkub: `relationship(back_populates="detail")`
Is this part necessary?

backend/app/models/personal_record.py
```diff
@@ -0,0 +1,35 @@
...
+from typing import TYPE_CHECKING
+from uuid import UUID
+
+from sqlalchemy import Integer, UniqueConstraint
+from sqlalchemy.orm import Mapped, mapped_column, relationship
+
+from app.database import BaseDbModel
+from app.mappings import FKUser, PrimaryKey, numeric_10_3, str_64
+
+if TYPE_CHECKING:  # pragma: no cover
```
czajkub: Same as before, models are excluded from pre-commit

backend/app/models/personal_record.py
```diff
@@ -0,0 +1,35 @@
...
+    from .user import User
+
+
+class PersonalRecord(BaseDbModel):
+    """Slow-changing physical attributes linked to a user."""
+
+    __tablename__ = "personal_record"
+    __table_args__ = (
+        UniqueConstraint("user_id", name="uq_personal_record_user_id"),
```
czajkub: There exists a `Unique` mapping in `mappings.py` - instead of adding `__table_args__`, you could do:
`user_id: Mapped[Unique[FKUser]]`

backend/app/models/personal_record.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    user_id: Mapped[FKUser]
+
+    age_years: Mapped[int | None] = mapped_column(Integer)
```
czajkub: we could add another mapping for this case, something like
`integer = Annotated[int, mapped_column(Integer)]` (probably should be some other name than integer)

backend/app/models/sleep_details.py
```diff
@@ -0,0 +1,29 @@
...
+from .event_record_detail import EventRecordDetail
+
+
+class SleepDetails(EventRecordDetail):
+    """Per-sleep aggregates and metrics."""
+
+    __tablename__ = "sleep_details"
+    __mapper_args__ = {"polymorphic_identity": "sleep"}
+
+    record_id: Mapped[PrimaryKey[UUID]] = mapped_column(
```
czajkub: same case as in `event_record_detail` model

backend/app/schemas/event_record.py
```diff
@@ -10,19 +10,34 @@
...
-    """Schema for creating a health record entry."""
+    max_speed: Decimal | None
+    max_watts: Decimal | None
+    moving_time_seconds: Decimal | None
+    total_elevation_gain: Decimal | None
+    average_speed: Decimal | None
+    average_watts: Decimal | None
+    elev_high: Decimal | None
+    elev_low: Decimal | None
+    sleep_total_duration_minutes: Decimal | None
```
czajkub: Why are we keeping sleep details in the event_record schemas? Shouldn't we create separate schemas for that?

backend/migrations/versions/a1db66fde2ce_health_record_details.py
czajkub, line 1: Maybe we could squash the three migrations into one? since there already are a lot and it's becoming hard to navigate through them

backend/app/schemas/event_record.py
czajkub: Also i suggest moving the rest of the details into a separate place

backend/app/models/event_record.py
czajkub, line 1: This entire logic should be moved from models. The only thing that should stay here is the names of the fields and relationship with details. What I suggest doing is instead of putting everything into the `EventRecordCreate`, to create another schema like `EventDetailCreate` that takes `**metrics` (or just assign what should be there, everything there should be nullable anyways), and call `workout_service.create()` on the `EventRecordCreate`s and add a `WorkoutDetailService` (similar to the old WorkoutStatisticService), since we probably will want to extract only some metrics instead of the whole record in the future.

czajkub: I'd suggest looking at [current Suunto service](https://github.com/the-momentum/open-wearables/blob/76054bc3a160bd5d38c48c628bbd636953c32496/backend/app/services/providers/suunto/workouts.py) for reference

backend/app/models/event_record_detail.py
czajkub: This definitely needs to be changed, without this the migrations dont work since it's trying to map an already mapped column.

czajkub: Temporarily it could be changed to something like 
```
record_id: Mapped[UUID] = mapped_column(
        ForeignKey("event_record.id", ondelete="CASCADE"), primary_key=True
    )
```
but we probably should make a separate mapping for this

backend/app/models/personal_record.py
czajkub: Little side-note - if you do that, when there is a record that violates this constraint, it will throw an exception instead of failing silently, so it could result in some data not being imported
#24

backend/app/models/personal_record.py
bartmichalak: Why do we want to keep age as an integer instead of having dob as a datetime? Calculating a user's age based on the integer might be painful

backend/app/models/personal_record.py
```diff
@@ -0,0 +1,35 @@
...
+    height_cm: Mapped[numeric_10_3 | None] = None
+    weight_kg: Mapped[numeric_10_3 | None] = None
```
bartmichalak: I'd be skeptical about whether `weight`, `body_fat_percentage`, or `rhr` can be treated as Slow-changing physical attributes linked to a user.

I'd be more into having these attributes as `Body metrics` (tbd how and where to store them), and in the Personal Record model (which could also just be called Profile) we'd store only `gender` and `dob`.

WDYT?

backend/app/schemas/health_record.py
FBegiello: Fair point - this was addressed in the changes below, as the data there was initial draft. Extended the HealthRecord to represent events as EventRecord with wariable part EventRecordDetails, which is meant to represent multiple detailed metrics.

backend/app/models/heart_rate_sample.py
FBegiello: device_id issue is a fair pont - we need to go through this, as for future data, some points might be coming from interpolation or other augmentation methods so it will have no device id per se (maybe a device id to dummy device named "synthetic_data" or something like that)

As to `source_name` - the time series tables need to be as minimal as possible - if we have id, we can get name from the device entity. We need to hold as little as possible per one data point record

backend/app/repositories/data_point_series_repository.py
FBegiello: Yes, this will need to be implemented for each of the supported types - currently I think we will pivot to base time series table with additional `type` field, so you can retriev data using one endpoint with `type` as parameter (`steps `for steps, `hear_rate `for HR and so on)

backend/app/schemas/time_series.py
FBegiello: Fair point - this will need to be added onto the draft here

backend/app/models/personal_record.py
FBegiello: True, this will need to be calculated from birthdate - will need to be changed from the draft version before merge

backend/app/models/personal_record.py
FBegiello: I'm open to having that as `body_metrics `or `body_state` recorded as slow changing time series - we can move those to a dedicated table

backend/app/schemas/time_series.py
```diff
@@ -0,0 +1,65 @@
...
+
+from datetime import datetime
+from decimal import Decimal
+from enum import Enum
+from uuid import UUID
+
+from pydantic import BaseModel, Field
+
+
+class SeriesType(str, Enum):
```
bartmichalak, line 10: Let's expand this list a little bit:

**Body Measurements**
- bodyMassIndex
- bodyFatPercentage
- height
- bodyMass
- leanBodyMass
- waistCircumference

**Activity**
- stepCount
- distanceWalkingRunning
- distanceCycling
- distanceWheelchair
- basalEnergyBurned
- activeEnergyBurned
- flightsClimbed
- pushCount
- distanceSwimming
- swimmingStrokeCount

**Vital Signs**
- heartRate
- bodyTemperature
- basalBodyTemperature
- bloodPressureSystolic
- bloodPressureDiastolic
- respiratoryRate
- restingHeartRate
- walkingHeartRateAverage
- heartRateVariabilitySDNN
- oxygenSaturation

**Cardio Fitness**
- vo2Max

might be a good start, WDYT?

bartmichalak: I think Blood Pressure is an interesting and tricky case - on the one hand, it's a single measurement, on the other hand, it has two separate values, which for us, will be easier to treat that way (as two separate records) ...but then we'll have a problem with retrieving the complete measurement from the database, because they won't have any relation to each other.

Any thoughts here? 🤔

backend/app/services/apple/apple_xml/xml_service.py
czajkub: fixed

backend/app/services/apple/apple_xml/xml_service.py
czajkub: fixed

backend/app/services/workout_service.py
czajkub: fixed

backend/app/models/event_record_detail.py
czajkub: yea

backend/app/models/event_record_detail.py
czajkub: created new mapping

backend/app/schemas/event_record.py
czajkub: moved

backend/app/models/event_record.py
czajkub: created another schema with details for scalability and readability

backend/app/models/data_point_series.py
```diff
@@ -0,0 +1,24 @@
...
+
+    __tablename__ = "data_point_series"
+    __table_args__ = (
+        Index("idx_data_point_series_device_type_time", "device_id", "series_type", "recorded_at"),
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    device_id: Mapped[str_100 | None] = None
+    recorded_at: Mapped[datetime_tz]
+    value: Mapped[numeric_10_3]
```
bartmichalak, line 21: I think we definitely need a unit here.

Even if we assume that each type will have a single, standard unit selected (we still have to decide how we approach this), I would still keep this information in the database.

backend/app/api/routes/v1/heart_rate.py
bartmichalak: I agree, we definitely shouldn't require `device_id` (it may be optional, but how exactly is the client supposed to know the value of this parameter?)

bartmichalak: I also have an additional question - what exactly is `device_id`? Is it a parameter we can control (e.g., a unique device_id assigned by the mobile SDKs we'll be releasing soon, or is it a parameter generated by specific providers, independent of us?)

If it's a parameter defined by external providers, which of them support it?

I'm just wondering if it wouldn't be worth separating it into a separate model - then the device would also have a name, model, etc.

backend/app/models/body_state.py
```diff
@@ -0,0 +1,21 @@
...
+
+class BodyState(BaseDbModel):
+    """Slow-changing physical body measurements captured as separate observations."""
+
+    __tablename__ = "body_state"
+
+    id: Mapped[PrimaryKey[UUID]]
+    user_id: Mapped[FKUser]
+
+    height_cm: Mapped[numeric_10_3 | None] = None
```
KaliszS: if we have `Mapped[data | None]` then `= None` is redundant.

backend/app/models/body_state.py
```diff
@@ -0,0 +1,21 @@
...
+from sqlalchemy.orm import Mapped
+
+from app.database import BaseDbModel
+from app.mappings import FKUser, PrimaryKey, numeric_10_3
+
+
+class BodyState(BaseDbModel):
+    """Slow-changing physical body measurements captured as separate observations."""
+
+    __tablename__ = "body_state"
```
KaliszS: it is better to rewrite function in database.py probably (using regex maybe)

backend/app/models/event_record.py
```diff
@@ -0,0 +1,36 @@
...
+    __table_args__ = (
+        Index("idx_event_record_user_category", "user_id", "category"),
+        Index("idx_event_record_time", "user_id", "start_datetime", "end_datetime"),
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    provider_id: Mapped[str_100 | None] = None
+    user_id: Mapped[FKUser]
+
+    category: Mapped[str_64] = mapped_column(default="workout")
```
KaliszS: I would put defaults on pydantc schemas side tbh, database usage is heavier.

backend/app/models/data_point_series.py
```diff
@@ -0,0 +1,24 @@
...
+    series_type: Mapped[SeriesType]
```
KaliszS: I think that enum should be a backend only thing. In database we should store integers. We can map them to our enum here or even create sepaate table with integers as PrimaryKeys/ids and series types names as value column.

backend/app/models/event_record.py
KaliszS: It should also be just a string mapped here to the value.

backend/app/models/event_record.py
```diff
@@ -0,0 +1,36 @@
...
+class EventRecord(BaseDbModel):
+    __tablename__ = "event_record"
+    __table_args__ = (
+        Index("idx_event_record_user_category", "user_id", "category"),
+        Index("idx_event_record_time", "user_id", "start_datetime", "end_datetime"),
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    provider_id: Mapped[str_100 | None] = None
+    user_id: Mapped[FKUser]
```
KaliszS: provider_id, user_id, source_name and device_id mght be highly repetitive, I would also consider to store that in separate table and leave her just a reference column.

backend/app/models/event_record_detail.py
```diff
@@ -0,0 +1,21 @@
...
+from app.mappings import FKEventRecord, ManyToOne, str_64
+
+
+class EventRecordDetail(BaseDbModel):
+    """Base polymorphic detail model used by specific aggregates (workout, sleep, etc.)."""
+
+    __tablename__ = "event_record_detail"
+
+    record_id: Mapped[FKEventRecord]
+    detail_type: Mapped[str_64]
```
KaliszS: again - int would be better imo

backend/app/models/personal_record.py
```diff
@@ -0,0 +1,21 @@
...
+class PersonalRecord(BaseDbModel):
+    """Slow-changing physical attributes linked to a user."""
+
+    __tablename__ = "personal_record"
+
+    id: Mapped[PrimaryKey[UUID]]
+    user_id: Mapped[Unique[FKUser]]
+
+    birth_date: Mapped[date_col | None] = None
+    gender: Mapped[str_64 | None] = None
```
KaliszS: do we need str_64 just to store two values? Maybe boolean would be better?

KaliszS (changes requested):

backend/app/models/body_state.py
```diff
@@ -0,0 +1,21 @@
+from uuid import UUID
+
+from sqlalchemy.orm import Mapped
+
+from app.database import BaseDbModel
+from app.mappings import FKUser, PrimaryKey, numeric_10_3
+
+
+class BodyState(BaseDbModel):
```
bartmichalak: I think all of this data should be stored in the `data_point_series` table - height is the least frequently changing attribute among these values, yet it can still change, especially in younger individuals.

Weight is tracked and logged daily by many users through their wearables, the same applies to body fat.

backend/app/models/data_point_series.py
```diff
@@ -0,0 +1,24 @@
...
+
+from sqlalchemy import Index
+from sqlalchemy.orm import Mapped
+
+from app.database import BaseDbModel
+from app.mappings import PrimaryKey, datetime_tz, numeric_10_3, str_100
+from app.schemas.time_series import SeriesType
+
+
+class DataPointSeries(BaseDbModel):
```
bartmichalak, line 10: We lost `user` relation here

FBegiello: This was intentional - we already have device id which is connected to a user anyway - I think we can skip holding direct user relation here, especially as one user can have a few devices

backend/app/models/data_point_series.py
FBegiello: This is up to debate - I'd be wary of storing units in the data points themselves, that's additional fields which will be identicall for all datapoints in one type. 
I'd be more in favour of having a simple mapping table with series_type -> unit which can be looked up if need be.
My idea is for the time series data record to be as minimal as possible (size concerns)

backend/app/models/health_record.py
FBegiello: Summed up steps is one thing but avg and min max could still be usefull for monitoring of consistency

backend/app/models/body_state.py
FBegiello: Moving those provisionary to time series table then - this will need to be addressed in future migrations/releaases

backend/app/models/personal_record.py
FBegiello: Addresed this as sex/gender separation in new model

backend/app/models/heart_rate_sample.py
FBegiello: As per other comments, solving most of the keys discrepancies by moving all the keys to a general mapping betwen user-device-provider

backend/app/schemas/time_series.py
FBegiello: The blood presure issue and other complex metrics are an intresting case - I'd wager we could resolve all this by moving eventually to more warehousing inspired data model strictly for time series, with separate tables for each type of series. While it is more implementation heavy and results for some cases in structurally identical tables but with different names, it gives us lots of wiggle room to have minimal required implementation for each type of data series (i.e. double record values for series like blood pressure without redundancy on simples series)

FBegiello: As per other values listed - I'd start with a base version of main types of series we support for the first release and eventually add more as we expand the structrure - resolving this for now as I'm adding some of those types

backend/app/models/body_state.py
FBegiello: Noted and solved

backend/app/models/event_record.py
FBegiello: This is a neat solution - I'm moving the general id mess towards that. Implementing a general mapping table, with id's shared and only reference as proposed

KaliszS (approved):

Issue #89, opened 2025-12-02
Handle duplicate email registration


bartmichalak:
Registration endpoint returns 500 error when attempting to register with an existing email. Should return 409 Conflict with a clear error message instead of letting the database constraint violation bubble up.

PR #90, opened 2025-12-02
Implement email uniqueness check in developer registration

bartmichalak:
Closes https://github.com/the-momentum/open-wearables/issues/89

### DIFF
--- a/backend/app/services/developer_service.py
+++ b/backend/app/services/developer_service.py
@@ -1,6 +1,8 @@
 from logging import Logger, getLogger
 from uuid import UUID
 
+from fastapi import HTTPException, status
+
 from app.database import DbSession
 from app.models import Developer
 from app.repositories.developer_repository import DeveloperRepository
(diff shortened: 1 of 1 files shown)


backend/app/services/developer_service.py
```diff
@@ -20,6 +22,20 @@ def __init__(self, log: Logger, **kwargs):
 
     def register(self, db_session: DbSession, creator: DeveloperCreate) -> Developer:
         """Create a developer with hashed password and server-generated fields."""
+        # Check if developer with this email already exists
+        existing_developers = self.crud.get_all(
```
KaliszS, line 26: I think that entire PR is redundant here, because we already have Unique constraint on developer email in databse, so the only thing we have to do is implement catching the right error. Using crud repository here will add unnecesary overload.

KaliszS (changes requested):

backend/app/services/developer_service.py
bartmichalak: Sure, we can close it for now and address this kind of issues globally

PR #101, opened 2025-12-03
Created settings page, with credentials and providers tabs

### LINKED ISSUE #91: Implement OAuth Provider Management in Settings
mplement settings interface allowing developers to configure which cloud OAuth providers are available to their end users. The backend endpoint /api/v1/oauth/providers is serves provider data with SVG icons. Frontend needs to display provider checkboxes in settings, save developer preferences via PUT request, and show only enabled providers during authorization flows.

Gasiek:
Resolves #91

### DIFF
--- a/frontend/src/lib/api/types.ts
+++ b/frontend/src/lib/api/types.ts
@@ -102,16 +102,16 @@ export type WearableProvider =
   | 'withings';
 
 export interface UserConnection {
+  user_id: string;
+  provider: string;
+  provider_user_id?: string;
+  provider_username?: string;
+  scope?: string;
   id: string;
-  userId: string;
-  providerId: string;
-  providerName: string;
-  status: 'active' | 'error' | 'pending' | 'disconnected';
-  connectedAt: string;
-  lastSyncAt: string | null;
-  syncStatus: 'success' | 'failed' | 'pending' | 'syncing';
-  syncError?: string;
-  dataPoints: number;
+  status: 'active' | 'revoked' | 'expired';
+  last_synced_at?: string;
+  created_at: string;
+  updated_at: string;
 }
 
 export interface HeartRateData {
--- a/backend/app/api/routes/v1/__init__.py
+++ b/backend/app/api/routes/v1/__init__.py
@@ -20,8 +20,8 @@ v1_router.include_router(oauth_router, prefix="/oauth", tags=["providers oauth"]
 # Garmin webhooks for push/ping notifications
 v1_router.include_router(garmin_webhooks_router, prefix="/garmin/webhooks", tags=["garmin webhooks"])
 # New unified vendor workouts endpoint
-v1_router.include_router(vendor_workouts_router, prefix="/vendors", tags=["providers workouts"])
-v1_router.include_router(sync_data_router, prefix="/vendors", tags=["sync data"])
+v1_router.include_router(vendor_workouts_router, prefix="/providers", tags=["providers workouts"])
+v1_router.include_router(sync_data_router, prefix="/providers", tags=["sync data"])
 v1_router.include_router(users_router, tags=["users"])
 v1_router.include_router(connections_router, tags=["data"])
 v1_router.include_router(heart_rate_router, tags=["data"])
--- a/backend/app/api/routes/v1/oauth.py
+++ b/backend/app/api/routes/v1/oauth.py
@@ -5,6 +5,7 @@ from fastapi import APIRouter, HTTPException, Query, status
 from fastapi.responses import RedirectResponse
 
 from app.database import DbSession
+from app.integrations.celery.tasks import sync_vendor_data
 from app.schemas import (
     AuthorizationURLResponse,
     BulkProviderSettingsUpdate,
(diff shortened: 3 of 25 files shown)


frontend/src/routes/_authenticated/settings/providers-tab.tsx
farce1, line 1: **Empty catch block** - This silently swallows errors. While the mutation's `onError` handles the toast, the empty catch block is still problematic. Consider removing the try/catch entirely since mutateAsync handles errors via onError:
```typescript
const handleSave = async () => {
  if (!providers) return;
  await updateMutation.mutateAsync({ providers: localToggleStates });
};
```

frontend/src/components/settings/providers/provider-item.tsx
farce1, line 1: **Image error handler doesn't show fallback** - When the image fails to load, it's hidden but the fallback letter is not displayed. The user sees an empty white box.

Current behavior:
- Image loads → shows image ✓
- Image fails → empty white box ✗
- No icon_url → shows fallback letter ✓

**Suggested fix:**
```typescript
const [imageError, setImageError] = useState(false);

// In the render:
{iconUrl && !imageError ? (
  <img
    src={iconUrl}
    alt={provider.name}
    className="h-12 w-12 object-contain"
    onError={() => setImageError(true)}
  />
) : (
  <div className="h-12 w-12 bg-white text-zinc-500 font-medium rounded-lg flex items-center justify-center">
    {provider.name.charAt(0).toUpperCase()}
  </div>
)}
```

farce1, line 1: **Unused data attribute** - `data-has-icon` is set but never used in CSS or JavaScript.

Either remove it or add corresponding styling:
```css
[data-has-icon="false"] { /* styles for no icon */ }
```

If it's not needed, remove to keep the code clean:
```diff
- data-has-icon={!!provider.icon_url}
```

frontend/src/components/ui/tabs.tsx
farce1, line 1: **Missing accessibility attributes** - The custom Tabs implementation is missing important ARIA attributes for screen readers:

1. **TabsTrigger** needs `aria-controls` pointing to its panel:
```typescript
aria-controls={`panel-${value}`}
```

2. **TabsContent** needs an `id` that matches:
```typescript
id={`panel-${value}`}
aria-labelledby={`tab-${value}`}
```

3. **TabsTrigger** needs an `id`:
```typescript
id={`tab-${value}`}
```

**Alternative:** Consider using `@radix-ui/react-tabs` which handles all accessibility automatically. The project already uses other Radix primitives.

frontend/src/components/settings/providers/provider-item.tsx
farce1, line 1: **Missing accessibility label on Switch** - The Switch toggle lacks an accessible label. Screen readers won't know what this switch controls.

**Suggested fix:** Add aria-label or associate with a label element:
```tsx
<Switch 
  checked={localToggleState} 
  onCheckedChange={onToggle}
  aria-label={`Toggle ${provider.name} provider`}
/>
```

Or wrap with a label:
```tsx
<label className="flex items-center gap-2">
  <span className="sr-only">Enable {provider.name}</span>
  <Switch checked={localToggleState} onCheckedChange={onToggle} />
</label>
```

frontend/src/routes/_authenticated/settings/providers-tab.tsx
farce1, line 1: **Potential state sync issue with useEffect** - This useEffect resets local state whenever `providers` changes. If the user has unsaved changes and the query refetches (e.g., window focus, interval), their changes will be lost without warning.

**Consider:**
1. Only sync on initial load:
```typescript
const [hasInitialized, setHasInitialized] = useState(false);

useEffect(() => {
  if (providers && providers.length > 0 && !hasInitialized) {
    // ... set initial state
    setHasInitialized(true);
  }
}, [providers, hasInitialized]);
```

2. Or show a warning when there are unsaved changes and new data arrives

3. Or disable refetchOnWindowFocus for this query if the current behavior is intentional

frontend/src/components/settings/providers/provider-item.tsx
farce1, line 1: **URL construction may fail silently** - If `VITE_API_URL` is undefined, the URL constructor might produce unexpected results.

**Suggested fix:** Add a fallback or validation:
```typescript
const apiUrl = import.meta.env.VITE_API_URL;
const iconUrl = provider.icon_url && apiUrl
  ? new URL(provider.icon_url, apiUrl).toString()
  : null;
```

Or if the icon_url can be an absolute URL:
```typescript
const iconUrl = provider.icon_url
  ? provider.icon_url.startsWith('http')
    ? provider.icon_url
    : new URL(provider.icon_url, import.meta.env.VITE_API_URL).toString()
  : null;
```

frontend/src/routes/__root.tsx
farce1, line 1: **Good: Exported queryClient for consistency** - Moving the QueryClient to a shared location and adding ReactQueryDevtools is a good practice. This ensures the same client instance is used throughout the app.

Minor note: Make sure `/lib/query/client.ts` exists and properly exports the queryClient. I don't see it in this PR's changes - was it added in a previous PR or is it missing?

frontend/src/lib/api/types.ts
farce1, line 1: **Type change may be a breaking change** - The `Provider` interface was significantly changed:

Old:
```typescript
id, name, description, logoUrl, isAvailable, features, authType
```

New:
```typescript
provider, name, has_cloud_api, is_enabled, icon_url
```

Please verify no other parts of the codebase depend on the old interface fields. Run a search for usages:
```bash
grep -r 'Provider' --include='*.ts' --include='*.tsx' | grep -v node_modules
```

frontend/src/lib/api/services/oauth.service.ts
farce1, line 1: **Good: Clean service layer structure** - The oauth.service.ts follows good patterns:
- Proper use of URLSearchParams for query building  
- Clean async/await syntax
- Type-safe return values

Minor suggestion: Consider adding JSDoc comments for better IDE support and documentation.

frontend/src/routes/_authenticated/settings/credentials-tab.tsx
farce1, line 1: **Good: Proper error handling in copy operation** - The async/await with try/catch for clipboard operations is the correct approach. The Clipboard API can fail (e.g., when permissions are denied or in non-secure contexts).

This was a good improvement from the previous version!

frontend/src/lib/query/keys.ts
farce1, line 1: **Good: Well-structured query keys** - The hierarchical query key structure follows best practices:
- `all` for invalidating everything
- `list` with parameters for specific queries

This makes cache invalidation predictable and efficient.

frontend/src/routes/_authenticated/settings.tsx
farce1, line 1: **Consider: Tab config outside component** - The `tabs` array is static and doesn't depend on props or state. Moving it outside the component prevents recreation on each render.

```typescript
// Move outside component
const tabs: TabConfig[] = [
  { id: 'credentials', label: 'Credentials', component: CredentialsTab },
  { id: 'providers', label: 'Providers', component: ProvidersTab },
];

function SettingsPage() {
  const [activeTab, setActiveTab] = useState(tabs[0].id);
  // ...
}
```

This is a minor optimization - React handles this well in practice.

frontend/src/components/ui/tabs.tsx
Gasiek: Good catch. I will add aria attributes. I will keep `tabs.tsx` since it's from Shadcn.

frontend/src/lib/api/types.ts
Gasiek: Yes, the type changed significantly. Everything works.

frontend/src/routes/__root.tsx
Gasiek: The query client was already created, but now that I look at this change, it hasn't been used indeed. It definitely requires configuration check.

frontend/src/routes/_authenticated/settings.tsx
Gasiek: Isn't it already outside?

frontend/src/routes/_authenticated/settings/providers-tab.tsx
Gasiek: Good catch. I will change that.

farce1 (approved):

Issue #107, opened 2025-12-04
Dashboard endpoint in API


KaliszS:
We need to create endoint `GET /api/v1/system-info` with tag `dashboard` (name & tag can be changed).

It should have returned dict:
```
{
    total_users: {count: int, weekly_growth: float),
    active_conn: {count: int, weekly_growth: float),
    data_points: {weekly_histogram: list[int], count: int, weekly_growth: float},
}
```

`data_points['weekly_histogram`] should contain 7 integers (number of data points in the last week, per day counting).
Weekly growth is a % of growth.

Todo in later tasks:
1. update data points read (#108)
2. add API calls count (#109)

bartmichalak:
Let's mock `data_points` for now, because it can be implemented only once https://github.com/the-momentum/open-wearables/issues/96 is completed.

KaliszS:
@Gasiek skip it for now, it's not planed for 0.1 and we can't count it right now.

PR #111, opened 2025-12-04
Created `GET /api/v1/system-info/dashboard`

Gasiek:
Resolves: #107

### DIFF
--- a/backend/app/schemas/__init__.py
+++ b/backend/app/schemas/__init__.py
@@ -64,14 +64,15 @@ from .suunto.workout_import import (
 from .suunto.workout_import import (
     WorkoutJSON as SuuntoWorkoutJSON,
 )
-from .system_info import (
-    CountWithGrowth,
-    DataPointsInfo,
-    SystemInfoResponse,
 from .sync import (
     ProviderSyncResult,
     SyncAllUsersResult,
     SyncVendorDataResult,
+)
+from .system_info import (
+    CountWithGrowth,
+    DataPointsInfo,
+    SystemInfoResponse,
 )
 from .user import (
     UserCreate,


backend/app/api/routes/v1/dashboard.py
```diff
@@ -0,0 +1,13 @@
...
+
+from app.database import DbSession
+from app.schemas.system_info import SystemInfoResponse
+from app.services import DeveloperDep, system_info_service
+
+router = APIRouter()
+
+
+@router.get("/system-info", response_model=SystemInfoResponse, tags=["dashboard"])
+async def get_system_info(db: DbSession, _developer: DeveloperDep):
```
KaliszS, line 11: @bartmichalak who should have access to that endpoint, just developer? No access from API key, right?

backend/app/services/system_info_service.py
```diff
@@ -0,0 +1,52 @@
...
+    """Service for system dashboard information."""
+
+    def __init__(self, log: Logger):
+        self.logger = log
+
+    def _calculate_weekly_growth(self, current: int, previous: int) -> float:
+        """Calculate weekly growth percentage."""
+        if previous == 0:
+            return 0.0 if current == 0 else 100.0
+        return ((current - previous) / previous) * 100.0
```
KaliszS, line 28: if I'm correct that will count growth % comapring to all historical data, so we will have lesser value each week. We probably need another query in repositories to get data between 7 and 14 days ago (maybe it's enough to have one function which accepts days period, like 7-14 or now-7).

But again, @bartmichalak can confirm.

KaliszS (changes requested):

backend/app/api/routes/v1/dashboard.py
bartmichalak: Right, just a developer

backend/app/services/system_info_service.py
bartmichalak: You're right @KaliszS , it should be a week-to-week comparison

backend/app/services/system_info_service.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_system_info(self, db_session: DbSession) -> SystemInfoResponse:
+        """Get system dashboard information."""
+        # Get total users data
+        total_users_current = user_service.crud.get_total_count(db_session)
+        total_users_week_ago = user_service.crud.get_total_count_week_ago(db_session)
+        total_users_growth = self._calculate_weekly_growth(total_users_current, total_users_week_ago)
+
+        # Get active connections data
+        connection_repo = UserConnectionRepository()
```
KaliszS: it should also use service, not repository

PR #138, opened 2025-12-08
Unified workout type mappings

### LINKED ISSUE #97: Standardize workout types
Currently the type field in /api/v1/users/{user_id}/workouts endpoint (and accordingly, in the database) is a plain string that stores the raw value from each provider. The problem is that every provider (e.g. Suunto, Polar, Apple Health) has its own taxonomy for workout types:

- **Apple HealthKit**: HKWorkoutActivityType.running (int: 37), HKWorkoutActivityType.cycling (int: 13), HKWorkoutActivityType.swimming (int: 46)
- **Suunto**: Running, Trail running, Cycling, Mountain biking, Pool swimming
- **Polar**: RUNNING, CYCLING, SWIMMING, STRENGTH_TRAINING

Without standarization, developers have to handle these differences themselves, which defeats the purpose of Open Wearables as a data unification layer.

We need to:
- [ ] define our own types 
- [ ] add provider mappings
- [ ] create dedicated docs page (Mintlify) with workout types

### LINKED ISSUE #132: Fix providers sync to newest data model
Right now we are getting some errors because of null constraints.

czajkub:
- add exhaustive workout type mappings for Suunto, Polar, Garmin and Healthkit
- add unique constraint to `EventRecord` model so no duplicates are allowed in database
- fix exception handling during creation of events

Resolves #132
Resolves #97

### DIFF
--- a/backend/app/services/apple/apple_xml/xml_service.py
+++ b/backend/app/services/apple/apple_xml/xml_service.py
@@ -15,7 +15,6 @@ from app.schemas import (
     HeartRateSampleCreate,
     StepSampleCreate,
 )
-from app.constants.workout_types.apple import get_unified_workout_type
 
 
 class XMLService:


backend/app/repositories/event_record_repository.py
```diff
@@ -29,7 +31,7 @@ def create(self, db_session: DbSession, creator: EventRecordCreate) -> EventReco
 
         creation_data = creator.model_dump()
         creation_data["external_mapping_id"] = mapping.id
-        for redundant_key in ("user_id", "provider_id", "device_id", "external_mapping_id"):
```
czajkub: we were creating the external mapping id, and then dropping it immediately afterwards, this was the issue in #132

backend/app/repositories/event_record_detail_repository.py
```diff
@@ -15,6 +16,7 @@ class EventRecordDetailRepository(
     def __init__(self, model: type[EventRecordDetail]):
         super().__init__(model)
 
+    @handle_exceptions
```
czajkub, line 19: so that no 500 internal server errrors are thrown

backend/app/models/event_record.py
```diff
@@ -16,6 +16,12 @@ class EventRecord(BaseDbModel):
     __table_args__ = (
         Index("idx_event_record_mapping_category", "external_mapping_id", "category"),
         Index("idx_event_record_mapping_time", "external_mapping_id", "start_datetime", "end_datetime"),
+        UniqueConstraint(
```
czajkub, line 19: external mapping id not enough because of the possibility of missing provider id

backend/app/repositories/event_record_repository.py
czajkub: though it looks like it was fixed with the recent push to main ;)

Issue #151, opened 2025-12-09
Enhance seed activity data script to include all available time series sample types


bartmichalak:
Enhance the `generate_time_series_samples` function in the seed activity data script to include all 14 available series types from `series_types.py`, with realistic sampling frequencies that reflect how often each metric is typically measured during workouts.

This will allows us:
- More comprehensive test data covering all (currently) available health metrics
- Realistic data distribution that reflects real-world usage patterns
- Better testing coverage for time series endpoints and data processing

Issue #150, opened 2025-12-09
Optimize init script: add flags to control generated data volume


bartmichalak:
The current init script creates 10 users and ~50k time series samples. This takes 1-2 minutes to complete, which is painful for local development when you just need a few records to test against.

We should add CLI flags to customize the generated dataset
`make init --users=2 --time-series-sample=50` 

Lower defaults to something reasonable.

PR #152, opened 2025-12-09
Enhance time series sample generation in seed_activity_data.py

bartmichalak:
Apart from seeding more time series data types, it limits the number of created users to 2 (it will be configurable when https://github.com/the-momentum/open-wearables/issues/150 is resolved)

Resolves: https://github.com/the-momentum/open-wearables/issues/151


bartmichalak:
Tested:

image

PR #168, opened 2025-12-10
Update Apple Health integration

### LINKED ISSUE #24: Duplicate detection strategy

### LINKED ISSUE #164: Handle exercise data from Apple Health in sync flow
We already have an endpoint for exercise sync ([reference](https://docs.openwearables.io/api-reference/import-data/import-data-healthion)), but after introducing the new data model we need to verify the entire flow works correctly with Apple Health.

We also need to:
- change the endpoint path, maybe `/api/v1/sdk/users/{user_id}/apple-health/exercises` 
- update (or actually provide) the documentation

### LINKED ISSUE #165: Handle health metrics data from Apple Health in sync flow
We already have an endpoint for health metrics sync ([reference](https://docs.openwearables.io/api-reference/import-data/import-data-healthion)), but after introducing the new data model we need to verify the entire flow works correctly with Apple Health.

We also need to:
- change the endpoint path, maybe /api/v1/sdk/users/{user_id}/apple-health/exercises
- update (or actually provide) the documentation

czajkub:
- update Healthion sdk import
- update xml import
- hotfixes to exception handling durning creation
- rename endpoints for import:
  - `/users/{user_id}/import/apple/` to `/sdk/users/{user_id}/sync/apple/` for healthion and auto-health-export
  - `xml` router stays the same, but now under `xml-import` tag

Resolves #164 
Resolves #165 
Resolves #24

### DIFF
--- a/backend/app/repositories/event_record_repository.py
+++ b/backend/app/repositories/event_record_repository.py
@@ -38,7 +38,7 @@ class EventRecordRepository(
             creation_data.pop(redundant_key, None)
 
         creation = self.model(**creation_data)
-        
+
         try:
             db_session.add(creation)
             db_session.commit()
@@ -59,7 +59,6 @@ class EventRecordRepository(
             if existing:
                 return existing
             raise
-
 
     def get_records_with_filters(
         self,
--- a/backend/app/constants/workout_types/__init__.py
+++ b/backend/app/constants/workout_types/__init__.py
@@ -1,4 +1,5 @@
-from .apple import get_healthkit_activity_name, get_unified_workout_type as get_unified_apple_workout_type
+from .apple import get_healthkit_activity_name
+from .apple import get_unified_workout_type as get_unified_apple_workout_type
 from .garmin import get_unified_workout_type as get_unified_garmin_workout_type
 from .polar import get_unified_workout_type as get_unified_polar_workout_type
 from .suunto import get_unified_workout_type as get_unified_suunto_workout_type
--- a/backend/app/utils/duplicates.py
+++ b/backend/app/utils/duplicates.py
@@ -29,14 +29,13 @@ def handle_duplicates(func: Callable[..., T]) -> Callable[..., T | None]:
 
                 query = db_session.query(model)
                 creator_data = creator.model_dump()
-                
+
                 mapper = inspect(model)
                 table = mapper.local_table
-                
+
                 unique_constraint = next(
-                    (constraint for constraint in table.constraints 
-                     if isinstance(constraint, UniqueConstraint)),
-                    None
+                    (constraint for constraint in table.constraints if isinstance(constraint, UniqueConstraint)),
+                    None,
                 )
 
                 if unique_constraint:
@@ -44,17 +43,16 @@ def handle_duplicates(func: Callable[..., T]) -> Callable[..., T | None]:
                         column_name = column.name
                         if column_name in creator_data:
                             query = query.filter(getattr(model, column_name) == creator_data[column_name])
-                            
+
                     existing = query.one_or_none()
                     if existing:
                         return existing
                     raise
-                
-                
+
                 for key in mapper.primary_key:
                     if key.name in creator_data:
                         query = query.filter(getattr(model, key.name) == creator_data[key.name])
-                        
+
                 existing = query.one_or_none()
                 if existing:
                     return existing


backend/app/repositories/event_record_detail_repository.py
```diff
@@ -33,10 +35,23 @@ def create(
         else:
             raise ValueError(f"Unknown detail type: {detail_type}")
 
-        db_session.add(detail)
-        db_session.commit()
-        db_session.refresh(detail)
-        return detail
+        try:
```
czajkub: hotfix

bartmichalak:
@czajkub  since we don't have a test suite yet and this functionality has no tests, we can't see what the payload should look like - could you share it here in the comments so I can test it internally using Postman?

backend/app/repositories/event_record_repository.py
```diff
@@ -35,10 +38,27 @@ def create(self, db_session: DbSession, creator: EventRecordCreate) -> EventReco
             creation_data.pop(redundant_key, None)
 
         creation = self.model(**creation_data)
-        db_session.add(creation)
-        db_session.commit()
-        db_session.refresh(creation)
-        return creation
+
```
czajkub, line 41: The `@handle_duplicates` decorator won't help in this case, since we are creating a new mapping from the id fields in the creator data. The decorator only receives the creator and cannot query based on that mapping, since it doesnt exist in the original data, so we need to handle this edge case manually

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -106,15 +110,15 @@ def _create_workout(
...
+        workout_type = get_unified_apple_workout_type(raw_type)
 
         duration_seconds = int((document["endDate"] - document["startDate"]).total_seconds())
 
         record = EventRecordCreate(
             category="workout",
             type=workout_type.value,
             source_name=document["sourceName"],
-            device_id=document.get("device"),
+            device_id=document.get("device", "")[:100],
```
czajkub, line 121: edge case handling - normally device id shouldn't go past 100 characters but I had singular cases where it exceeded that limit

KaliszS (approved):
Approved, since its merged into #175

PR #172, opened 2025-12-10
Implemented pagination for users on frontend and backend

### LINKED ISSUE #47: Add pagination to `/users` endpoint
## Problem
The `GET /users` endpoint currently returns all users without pagination support, which can cause performance issues as the user base grows.

## Current Implementation
- Endpoint uses `db.query(user_service.crud.model).all()` directly
- `CrudRepository.get_all()` method already supports pagination with `offset`, `limit`, `filters`, and `sort_by` parameters
- Other endpoints (e.g., `/workouts`, `/records`) already implement pagination using query param schemas

## Solution
Update the `/users` endpoint to:
1. Accept pagination query parameters (offset, limit, filters, sort_by)
2. Use `user_service.crud.get_all()` instead of direct query
3. Follow the same pattern as other paginated endpoints (e.g., use `FilterParams` or similar query schema)

## Reference
- Repository pagination support: `backend/app/repositories/repositories.py` (lines 30-46)
- Example implementations: `backend/app/api/routes/v1/workouts.py`, `backend/app/api/routes/v1/records.py`
- Filter params schema: `backend/app/schemas/filter_params.py`

Gasiek:
Resolves #47

### DIFF
--- a/frontend/src/lib/api/services/users.service.ts
+++ b/frontend/src/lib/api/services/users.service.ts
@@ -1,5 +1,6 @@
 import { apiClient } from '../client';
 import { API_ENDPOINTS } from '../config';
+import { appendSearchParams } from '@/lib/utils/url';
 import type {
   UserRead,
   UserCreate,
@@ -12,15 +13,17 @@ export const usersService = {
   async getAll(params?: UserQueryParams): Promise<PaginatedUsersResponse> {
     const searchParams = new URLSearchParams();
 
-    if (params?.page != null) searchParams.set('page', params.page.toString());
-    if (params?.limit != null)
-      searchParams.set('limit', params.limit.toString());
-    if (params?.sort_by) searchParams.set('sort_by', params.sort_by);
-    if (params?.sort_order) searchParams.set('sort_order', params.sort_order);
-    if (params?.search) searchParams.set('search', params.search);
-    if (params?.email) searchParams.set('email', params.email);
-    if (params?.external_user_id)
-      searchParams.set('external_user_id', params.external_user_id);
+    if (params) {
+      appendSearchParams(searchParams, {
+        page: params.page,
+        limit: params.limit,
+        sort_by: params.sort_by,
+        sort_order: params.sort_order,
+        search: params.search,
+        email: params.email,
+        external_user_id: params.external_user_id,
+      });
+    }
 
     const queryString = searchParams.toString();
     const url = queryString
--- a/frontend/src/lib/utils/index.ts
+++ b/frontend/src/lib/utils/index.ts
@@ -1,2 +1,3 @@
 export * from './format';
 export * from './health';
+export * from './url';
--- a/frontend/src/lib/utils/url.ts
+++ b/frontend/src/lib/utils/url.ts
@@ -1,0 +1,13 @@
+/**
+ * Appends non-null, non-empty values to URLSearchParams
+ */
+export function appendSearchParams(
+  searchParams: URLSearchParams,
+  params: Record<string, string | number | undefined | null>
+): void {
+  for (const [key, value] of Object.entries(params)) {
+    if (value != null && value !== '') {
+      searchParams.set(key, String(value));
+    }
+  }
+}


backend/app/services/user_service.py
```diff
@@ -22,6 +25,34 @@ def get_count_in_range(self, db_session: DbSession, start_date: datetime, end_da
         """Get count of users created within a date range."""
         return self.crud.get_count_in_range(db_session, start_date, end_date)
 
+    @handle_exceptions
+    def get_users_paginated(
```
czajkub, line 51: May be a bit too nitpicky, but I'd put the `get_users_paginated` method below the `create` and `update` overrides, just a suggestion though

farce1:
#### 1. SQL Injection Risk via `sort_by` - Strengthen Validation
**File:** `backend/app/repositories/user_repository.py:63-65`

```python
sort_by_column = query_params.sort_by or "created_at"
if not hasattr(self.model, sort_by_column):
    raise ValueError(f"Invalid sort_by column: {sort_by_column}")
```

**Problem:** While there's a `hasattr` check, the `ValueError` exposes internal details and isn't caught properly at the API layer. If the Pydantic `Literal` type ever gets out of sync with the model, this could be exploited.

**Recommendation:** Use an explicit allowlist constant that's shared between the schema and repository:

```python
# In a constants file or at module level
ALLOWED_SORT_COLUMNS = frozenset({"created_at", "email", "first_name", "last_name"})

# In repository
if sort_by_column not in ALLOWED_SORT_COLUMNS:
    raise ValueError("Invalid sort column")
```

---

#### 2. Race Condition in Frontend State Sync
**File:** `frontend/src/components/users/users-table.tsx:78-91`

```tsx
useEffect(() => {
  const sortColumn = sorting[0];
  // ... builds params
  onQueryChangeRef.current({ ... });
}, [pagination, sorting, debouncedSearch]);

useEffect(() => {
  setPagination((prev) => ({ ...prev, pageIndex: 0 }));
}, [debouncedSearch]);
```

**Problem:** The order of these effects isn't guaranteed by React. When `debouncedSearch` changes, both effects fire but the pagination reset may happen *after* the query fires with the stale page number, causing an incorrect initial fetch.

**Recommendation:** Combine the logic or use a reducer pattern:

```tsx
useEffect(() => {
  const sortColumn = sorting[0];
  const sortBy = sortColumn ? columnToSortBy[sortColumn.id] : 'created_at';
  const sortOrder = sortColumn?.desc ? 'desc' : 'asc';

  onQueryChangeRef.current({
    page: 1, // Always reset to page 1 when search changes
    limit: pagination.pageSize,
    sort_by: sortBy,
    sort_order: sortOrder,
    search: debouncedSearch || undefined,
  });
  
  setPagination((prev) => ({ ...prev, pageIndex: 0 }));
}, [debouncedSearch]); // Separate effect for search

useEffect(() => {
  // Only for pagination/sorting changes (not search)
  // ...
}, [pagination.pageIndex, pagination.pageSize, sorting]);
```

---

#### 3. N+1 Query Risk
**File:** `backend/app/services/user_service.py:49-51`

```python
return PaginatedResponse[UserRead](
    items=[UserRead.model_validate(user) for user in users],
    ...
)
```

**Problem:** If `UserRead` ever needs lazy-loaded relationships, this will cause N+1 queries since each validation could trigger additional DB calls.

**Recommendation:** Add eager loading in the repository query:

```python
from sqlalchemy.orm import selectinload

query = db_session.query(self.model).options(
    selectinload(self.model.some_relationship)  # if needed in future
)
```

#### 4. Division by Zero - Belt & Suspenders
**File:** `backend/app/schemas/common.py:21-24`

The `Field(gt=0)` validator protects against this in API requests, but `PaginatedResponse` can be instantiated directly in code. Consider adding a Pydantic `field_validator` as a safety net.

---

#### 5. Frontend Type Inconsistency
**File:** `frontend/src/lib/api/types.ts`

Two pagination types exist that could diverge:
- `PaginatedResponse<T>` (generic, lines 8-14) - uses `hasMore`
- `PaginatedUsersResponse` (lines 47-55) - uses `has_next`/`has_prev`

**Recommendation:** Remove or update the generic `PaginatedResponse<T>` to match backend naming, or consolidate into a single type.

---

#### 6. Missing Clipboard API Guard
**File:** `frontend/src/components/users/users-table.tsx:97-106`

```tsx
const handleCopyId = async (id: string) => {
  try {
    await navigator.clipboard.writeText(id);
    // ...
  } catch (error) {
    toast.error('Failed to copy user ID to clipboard');
  }
};
```

**Problem:** In non-HTTPS contexts or older browsers, `navigator.clipboard` may be `undefined`, causing the error to be less helpful.

**Recommendation:**
```tsx
const handleCopyId = async (id: string) => {
  if (!navigator.clipboard) {
    toast.error('Clipboard API not available');
    return;
  }
  // ... rest of code
};
```

PR #175, opened 2025-12-11
Model changes

czajkub:
- Fix external mapping - uses provider name now instead of external id
- external id moved to models from the mappings
- renamed workout_details, sleep_details to workout_detail, sleep_detail
- provider id renamed to external_id
- series_type renamed to series_type_definition_id


backend/app/integrations/celery/tasks/process_upload_task.py
```diff
@@ -29,12 +30,19 @@ def process_uploaded_file(bucket_name: str, object_key: str) -> dict[str, str]:
...
+
+            # Validate that the user exists before processing
+            _ = user_service.get(db, user_id, raise_404=True)
 
             s3_client.download_file(bucket_name, object_key, temp_xml_file)
 
             try:
-                _import_xml_data(db, temp_xml_file, user_id)
+                _import_xml_data(db, temp_xml_file, user_id_str)
                 db.commit()
```
KaliszS: btw, what are we commiting here?

backend/app/constants/series_types/definitions.py
```diff
@@ -0,0 +1,32 @@
+from app.schemas.time_series import SeriesType
+
+# Stable integer identifiers for each series type. These IDs are persisted in the database,
+# so update with caution.
+SERIES_TYPE_DEFINITIONS: list[tuple[int, SeriesType, str]] = [
```
KaliszS: Its design is going to change in #191, take it into consideration ;)

backend/app/api/routes/v1/garmin_webhooks.py
```diff
@@ -84,6 +84,8 @@ async def garmin_ping_notification(
                     logger.info(f"Mapped Garmin user {garmin_user_id} to internal user {internal_user_id}")
 
                     # Extract parameters from callback URL (including pull token)
+                    from urllib.parse import parse_qs, urlparse
```
KaliszS: Imports shuldn't be moved outside module header.

backend/app/api/routes/v1/garmin_webhooks.py
```diff
@@ -84,6 +84,8 @@ async def garmin_ping_notification(
                     logger.info(f"Mapped Garmin user {garmin_user_id} to internal user {internal_user_id}")
```
KaliszS, line 84: Do you think this is notification we will be getting often enough to change this into debug? The same question goes to the previous logger info.

backend/app/models/data_point_series.py
```diff
@@ -4,19 +4,27 @@
...
     __tablename__ = "data_point_series"
     __table_args__ = (
-        Index("idx_data_point_series_mapping_type_time", "external_mapping_id", "series_type_id", "recorded_at"),
+        Index("idx_data_point_series_mapping_type_time", "external_device_mapping_id", "series_type_definition_id", "recorded_at"),
     )
 
     id: Mapped[PrimaryKey[UUID]]
-    external_mapping_id: Mapped[FKExternalMapping]
+    external_id: Mapped[str_100 | None]
+    external_device_mapping_id: Mapped[FKExternalMapping]
```
KaliszS, line 27: So for last clarification - what's exactly external_id again?

backend/app/models/external_device_mapping.py
```diff
@@ -24,7 +24,7 @@ class ExternalDeviceMapping(BaseDbModel):
 
     id: Mapped[PrimaryKey[UUID]]
     user_id: Mapped[FKUser]
-    provider_id: Mapped[str_100 | None]
+    provider_name: Mapped[str_10]
     device_id: Mapped[str_100 | None]
```
KaliszS, line 28: we can also add a reference id for provider's gear software version model

backend/app/models/sleep_detail.py
```diff
@@ -5,10 +5,10 @@
 from .event_record_detail import EventRecordDetail
 
 
-class SleepDetails(EventRecordDetail):
+class SleepDetail(EventRecordDetail):
     """Per-sleep aggregates and metrics."""
 
-    __tablename__ = "sleep_details"
+    __tablename__ = "sleep_detail"
```
KaliszS: I'm not sure if this is correct. I mean it contains multiple details, right? Same concern with workout_detail.

backend/app/models/version.py
```diff
@@ -0,0 +1,14 @@
+from uuid import UUID
+
+from sqlalchemy.orm import Mapped
+
+from app.database import BaseDbModel
+from app.mappings import FKDevice, PrimaryKey, str_10, str_100
+
+
+class Version(BaseDbModel):
```
KaliszS: I was thinking about the best name. Maybe `DeviceSoftware`?
@bartmichalak @FBegiello

docs/docs.json
```diff
@@ -17,8 +17,7 @@
             "group": "Getting started",
             "pages": [
               "index",
-              "quickstart",
-              "roadmap"
```
KaliszS, line 21: Why roadmap got removed?

backend/app/repositories/repositories.py
```diff
@@ -18,6 +19,7 @@ def __init__(self, model: type[ModelType]):
         self.model = model
 
     @handle_exceptions
```
KaliszS, line 21: Oh, I'm not sure how it got there, but handle_exceptions was destined for services. Not sure if it even works here. Probably your handle_duplicates decorator makes it redundant.

backend/app/utils/duplicates.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs) -> T | None:
+        try:
+            return func(*args, **kwargs)
+        except SQLAIntegrityError as e:
+            if isinstance(e.orig, UniqueViolation):
+                model = args[0].model
+                db_session = args[1]
+                creator = args[2]
```
KaliszS, line 29: Are we 100% sure that args always have at least 3 elements? Otherwise we will get here another exception.

backend/app/utils/duplicates.py
```diff
@@ -0,0 +1,62 @@
...
+                if unique_constraint:
+                    for column in unique_constraint.columns:
+                        column_name = column.name
+                        if column_name in creator_data:
+                            query = query.filter(getattr(model, column_name) == creator_data[column_name])
+
+                    existing = query.one_or_none()
+                    if existing:
+                        return existing
+                    raise
```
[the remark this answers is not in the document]
KaliszS: yeah, it's true I think

backend/app/utils/duplicates.py
```diff
@@ -0,0 +1,62 @@
...
+
+from psycopg.errors import UniqueViolation
+from sqlalchemy import UniqueConstraint
+from sqlalchemy.exc import IntegrityError as SQLAIntegrityError
+from sqlalchemy.inspection import inspect
+
+T = TypeVar("T")
+
+
+def handle_duplicates(func: Callable[..., T]) -> Callable[..., T | None]:
```
KaliszS, line 12: The question is - what is expected behaviour? Will it break code execution flow? Because it probaly shouldn't. Maybe it should just have debug logs. That decorator will be fired during syncs most times. I guess...

backend/app/api/routes/v1/garmin_webhooks.py
czajkub: My bad, still had some uncommited changes

backend/app/integrations/celery/tasks/process_upload_task.py
czajkub: actually nothing, this commit is unnecessary

backend/app/models/data_point_series.py
czajkub: external_id being the id's assigned to each record by vendors, and external mapping contains info about the device from which the data was captured (with information like which provider as well)

backend/app/models/sleep_detail.py
czajkub: Now that you mention it, plural form might be better after all. I initially changed it to singular so that every table name has the same format

backend/app/repositories/repositories.py
czajkub: I think without it there were 500 errors thrown if something went wrong (client-side), but with the decorator the correct http status was returned (4xx). Now it is probably redundant though

backend/app/utils/duplicates.py
czajkub: It won't break flow, actually the opposite, when we try to import a duplicate record into the database, instead of throwing an error, the already existing one is returned instead

backend/app/utils/duplicates.py
czajkub: yes, every `create` method in repositories (and its inheritors as well) take session and creator as arguments, and the first is just the repository with the `model` attribute (e.g. `EventRecord`)

docs/docs.json
czajkub: i think it was some rebase issue

backend/app/models/version.py
bartmichalak: `DeviceSoftware` sounds good 👍

KaliszS (approved):

Issue #167, opened 2025-12-10
Developer (team member) management system


bartmichalak:
Currently, there's no way to add another team member to the system. The Developer model exists and supports basic CRUD operations through the API, but there's no UI for managing team members.

Tasks
- [ ] Admin panel section for team management
  - [ ] List view (first name, last name, email, created_at)
  - [ ] Add new team member (developer)
  - [ ] Remove team member
- [ ] Invitation module – handles the flow for adding new developers
- [ ] Email configuration – we need to decide on a provider. Should be free for low volume, e.g., [Resend](https://resend.com/pricing)

PR #185, opened 2025-12-15
Dev member management system

Gasiek:
Resolves: #167 

image
image
image
image

### DIFF
--- a/backend/app/api/routes/v1/__init__.py
+++ b/backend/app/api/routes/v1/__init__.py
@@ -9,7 +9,6 @@ from .events import router as events_router
 from .garmin_webhooks import router as garmin_webhooks_router
 from .heart_rate import router as heart_rate_router
 from .import_xml import router as import_xml_router
-from .import_data import router as import_data_router
 from .invitations import router as invitations_router
 from .oauth import router as oauth_router
 from .sdk_sync import router as sdk_sync_router


backend/app/mappings.py
```diff
@@ -17,6 +17,7 @@
     mapped_column(primary_key=True, autoincrement=True),
 ]  # use for composite integer primary keys (single PK int will have it auto enabled)
 Unique = Annotated[T, mapped_column(unique=True)]
+UniqueIndex = Annotated[T, mapped_column(unique=True, index=True)]
```
KaliszS: It's not the first time it gets into review ;) But I believe it's redundant. Let me clarify!

Default index structure in postgres (and most SQL DBMS) is B-tree. Unique constrain also using B-tree to make uniqueness check. So if you have `unique=True` it already created B-tree index. Of course, if you would like to use different index (like GIN or GIST), then it would be reasonable to use both unique and index withint `mapped_column`.

backend/app/models/developer.py
```diff
@@ -13,5 +13,7 @@ class Developer(BaseDbModel):
     created_at: Mapped[datetime_tz]
     updated_at: Mapped[datetime_tz]
 
+    first_name: Mapped[str_100 | None]
+    last_name: Mapped[str_100 | None]
```
KaliszS, line 17: What's purpose of those two fields?

backend/app/models/invitation.py
```diff
@@ -0,0 +1,25 @@
...
+    email: Mapped[str_255]
+    token: Mapped[UniqueIndex[str_255]]
+    status: Mapped[str_50]  # pending, sent, failed, accepted, expired, revoked
+    expires_at: Mapped[datetime_tz]
+    created_at: Mapped[datetime_tz]
+
+    invited_by_id: Mapped[FKDeveloper | None]
+    invited_by: Mapped[ManyToOne["Developer"] | None] = relationship(
+        "Developer",
+        foreign_keys="[Invitation.invited_by_id]",
```
KaliszS, line 24: We have many-to-one relationship, but it is stored inside the same model. Relations within SQLalchemy exists only in Python (field invited_by won't be included into database table). So maybe it makes more sense to create bi-directional references?

backend/app/repositories/invitation_repository.py
```diff
@@ -0,0 +1,44 @@
...
+        return db_session.execute(stmt).scalar_one_or_none()
+
+    def get_active_invitations(self, db_session: DbSession) -> list[Invitation]:
+        """Get all active invitations (pending, sent, or failed)."""
+        stmt = (
+            select(self.model)
+            .where(self.model.status.in_(VISIBLE_INVITATION_STATUSES))
+            .order_by(self.model.created_at.desc())
+        )
+        return list(db_session.execute(stmt).scalars().all())
```
KaliszS: Isn't `all()` returning list already?

backend/app/services/invitation_service.py
```diff
@@ -0,0 +1,224 @@
...
+            token=token,
+            status=InvitationStatus.PENDING,
+            expires_at=expires_at,
+            created_at=datetime.now(timezone.utc),
+            invited_by_id=invited_by.id,
+        )
+
+        db_session.add(invitation)
+        db_session.commit()
+        db_session.refresh(invitation)
```
KaliszS: No db operations within services ;) Please, refactor that servcie, a lot of code should probably be moved into repository.

backend/app/utils/email_client.py
```diff
@@ -0,0 +1,105 @@
...
+import re
+from html import escape
+from typing import Any, cast
+
+import resend
+
+from app.config import settings
+
+logger = logging.getLogger(__name__)
+EMAIL_REGEX = re.compile(r"^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$", re.IGNORECASE)
```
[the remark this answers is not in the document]
KaliszS: Won't `EmailStr` from pydantic solve the issue?

backend/app/services/invitation_service.py
```diff
@@ -0,0 +1,224 @@
...
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"Cannot resend invitation with status: {invitation.status}",
+            )
+
+        # Generate new token, extend expiry, and reset status to PENDING
+        invitation.token = self._generate_token()  # type: ignore[assignment]
+        invitation.expires_at = datetime.now(timezone.utc) + timedelta(  # type: ignore[assignment]
+            days=settings.invitation_expire_days,
+        )
+        invitation.status = InvitationStatus.PENDING  # type: ignore[assignment]
```
KaliszS: Of course that should be inside repository, but are those type ignores required? It sounds like issue caused by mypy, while we are using ty. Is that complaining? If so, have you checked how it suggest to fix that?

backend/app/config.py
```diff
@@ -78,6 +78,13 @@ class Settings(BaseSettings):
     polar_redirect_uri: str = "http://localhost:8000/api/v1/oauth/polar/callback"
     polar_default_scope: str = "accesslink.read_all"
 
+    # EMAIL SETTINGS (Resend)
+    resend_api_key: SecretStr | None = None
+    email_from_address: str | None = None
+    email_from_name: str = "Open Wearables"
+    frontend_url: str = "http://localhost:3000"
```
KaliszS, line 85: Are you sure that backend should require url from frontend? I thought it's usually the other way around.

backend/app/config.py
```diff
@@ -78,6 +78,13 @@ class Settings(BaseSettings):
     polar_redirect_uri: str = "http://localhost:8000/api/v1/oauth/polar/callback"
     polar_default_scope: str = "accesslink.read_all"
 
+    # EMAIL SETTINGS (Resend)
+    resend_api_key: SecretStr | None = None
+    email_from_address: str | None = None
+    email_from_name: str = "Open Wearables"
```
KaliszS, line 84: What does those names mean?

backend/app/integrations/celery/tasks/send_email_task.py
```diff
@@ -0,0 +1,130 @@
+from logging import getLogger
+from typing import Any
+
+from billiard.einfo import ExceptionInfo
+
+from app.database import SessionLocal
```
KaliszS, line 8: maybe we should use DbSession?

backend/app/integrations/celery/tasks/send_email_task.py
```diff
@@ -0,0 +1,130 @@
...
+from app.models import Invitation
+from app.schemas.invitation import InvitationStatus
+from app.utils.email_client import send_invitation_email
+from celery import Task, shared_task
+
+logger = getLogger(__name__)
+
+MAX_RETRIES = 5
```
KaliszS: I would add this to application settings.

backend/app/integrations/celery/tasks/send_email_task.py
```diff
@@ -0,0 +1,130 @@
+from logging import getLogger
+from typing import Any
+
+from billiard.einfo import ExceptionInfo
```
KaliszS, line 5: What's that?

KaliszS (changes requested):

backend/app/models/developer.py
bartmichalak: Those attributes are displayed in the list of team members:

image

and it's fine

backend/app/config.py
bartmichalak: The backend is responsible for sending invitation emails and in the content there's a button for accepting the invitation that contains a URL for the frontend - I think it's the easiest and most intuitive way to make it work 🤔

backend/app/config.py
bartmichalak: image

Name of the sender

backend/app/mappings.py
bartmichalak: Since this error keeps getting into review, maybe it's time to add this to the rules?

Would you mind adding this to backend rules in
https://github.com/the-momentum/open-wearables/pull/187?

and maybe separately to cursor rules, since we haven't decided yet which approach we want to take with those rules files.

backend/app/models/invitation.py
Gasiek: Keeping it unidirectional. The relationship is only accessed in one direction (getting inviter's email when resending). There's no current use case for developer.invitations_sent. If needed later, we'd add a repository method rather than a relationship — more explicit and testable.

backend/app/services/invitation_service.py
Gasiek: It was `ty` that complained. But there were no suggestions.
```
error[invalid-assignment]: Object of type `str` is not assignable to attribute `token`...
    |
208 |         invitation.token = self._generate_token()
    |         ^^^^^^^^^^^^^^^^
    |
info: rule `invalid-assignment` is enabled by default
```
Already fixed it by using schmeas

backend/app/config.py
Gasiek: @KaliszS You are right but for API calls. Emails are a special case - they're rendered by the backend but consumed outside the app entirely. So _this is the way_

backend/app/integrations/celery/tasks/send_email_task.py
Gasiek: No - DbSession uses FastAPI's Depends() which only works inside request handlers. Celery tasks run outside FastAPI, so we use SessionLocal() directly.

backend/app/integrations/celery/tasks/send_email_task.py
Gasiek: ExceptionInfo is the correct type for the einfo parameter in Celery's on_failure callback. We import it from billiard (Celery's multiprocessing dependency) to properly type-annotate the method override. Without it, the type checker would complain about an untyped parameter. It provides access to the exception traceback when a task fails after all retries.

backend/app/repositories/invitation_repository.py
Gasiek: This `db_session.execute(stmt).scalars().all()` actually returns a `Sequence[T]` and AFAIK this is the modern SQLAlchemy API, compared to the legacy style `db_session.query(self.model).filter(...).all()` that returns `list[T]`. So maybe a good idea would be switching to the modern one in other places to. At runtime, both return the same thing (a list-like object). The difference is only what the type checker allows you to do with it. And Sequence is actually better because it enforces immutability at the type level.

backend/app/config.py
KaliszS: so maybe that would make more sense to name those `sender_address` and `sender_name`?

bartmichalak: Agreed, @Gasiek we can change those setting names to:
- `email_sender_address` 
- `email_sender_name`

Gasiek: @bartmichalak @KaliszS 
Resend's official API docs: https://resend.com/docs/api-reference/emails
Their API request uses from, not sender:
```
{
  "from": "Acme <onboarding@resend.dev>",
  "to": ["delivered@resend.dev"],
  "subject": "hello world",
  "html": "<p>it works!</p>"
}
```

- RFC 5322 defines From: as the author's mailbox
- Sender: is a separate header used only when someone sends on behalf of another person
So I would keep it unless you really want it your way.

PR #199, opened 2025-12-18
feat: Timeseris endpoint

### LINKED ISSUE #196: Implement timeseries endpoint logic
- Endpoint: `GET /users/{user_id}/timeseries`.
- Serice already exists, needs to be reworked.
- Support filtering by `SeriesType`.
- Return `PaginatedResponse`

KaliszS:
Resolves #196

Example data based on makefile seed:
- start_date: 1761994565
- end_time: 1764586565
(it's entire November)


bartmichalak (changes requested):
1. When I try to use cursor I get 500 error:

```
Invalid cursor format: '2025-06-20T15:27:12.371754 00:00|4c33c01c-69bb-47bd-8e1b-8d452df6c767'. Expected 'timestamp|id'.
```

I think we could go with base-64 encoded cursor. 

2. API should return data in consistent way

```
{
  "value": ...,
  "unit": ...,
  "type": ...,
  "timestamp": ...
}
```


Now, all data types have different field names (like `bpm` for heart rate)

3. Currently, the API supports filtering by only one data type. I definitely see scenarios where a client would want to fetch multiple data types in one request.

PR #253, opened 2026-01-06
  chore(backend): Add default-groups to uv config for dev setup

healthkowshik:
## Description

```bash
# Error the very first time.
docker compose exec app bash -c "cd /root_project && uv run ty check"
error: Failed to spawn: `ty`
  Caused by: No such file or directory (os error 2)

# Works when code-quality is added.
docker compose exec app bash -c "cd /root_project && uv run --group code-quality ty check"
Installed 8 packages in 10ms
All checks passed!

# Works the second time, without code-quality
docker compose exec app bash -c "cd /root_project && uv run ty check"                     
All checks passed!
```

### Here's what's happening:

1. First run fails - `ty` isn't installed because `code-quality` group isn't part of the default sync.
2. With `--group code-quality` - uv installs the group dependencies (including `ty`).
3. Second run works - packages are now cached/installed from the previous run.

## Changes

Configure uv to include `dev` and `code-quality` dependency groups by default when running `uv sync`, removing the need to manually specify `--group` flags for local development. This is a small config change that improves DX by making the default sync behavior match what developers actually need.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -36,7 +36,7 @@
 
 - [ ] `uv run ruff check` passes
 - [ ] `uv run ruff format --check` passes
-- [ ] `uv run ty check` passes
+- [ ] `uv run --group code-quality ty check` passes
 
 ### Frontend Changes
 
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -31,7 +31,7 @@ dependencies = [
 code-quality = [
     "pre-commit>=4.3.0",
     "ruff>=0.14.3",
-    "ty>=0.0.10",
+    "ty>=0.0.1a25",
 ]
 dev = [
     "pytest>=8.4.1",


KaliszS (changes requested):
This is not what we really want. We don't want to install dev ddpendencies on production build and that's how it would have ended.

healthkowshik:
Do you recommend we change the Pull Request template then?

Ref: https://github.com/the-momentum/open-wearables/blob/main/.github/PULL_REQUEST_TEMPLATE.md

```bash
# The following line exist at the moment.
uv run ty check

# Instead, we change it to the line below?
uv run --group code-quality ty check
```

KaliszS:
> Do you recommend we change the Pull Request template then?
> 
> Ref: https://github.com/the-momentum/open-wearables/blob/main/.github/PULL_REQUEST_TEMPLATE.md
> 
> ```shell
> # The following line exist at the moment.
> uv run ty check
> 
> # Instead, we change it to the line below?
> uv run --group code-quality ty check
> ```

Yeah I think so.  Also I would use just pre-commit instead of ty and ruff separately. Someone generated this with AI, but it's redundant.

PR #273, opened 2026-01-09
feat(api): implement daily activity summary with metrics calculation

### LINKED ISSUE #267: Data summaries endpoint: Activity/workout summary
**Issue**
Current summary is a draft approach with limited data access and calculation:

- Currently, there is no dedicated method in the `SummariesService` to serve any activity summaries
- Data served needs to be calculated from the raw fields in the database in some cases (see especially `floors_climbed` (summary) vs `total_elevation_gain` (data model) as well as composits such as `active_calories_kcal` and `total_calories_kcal` vs only `energy_burned` in the data model)
- Intensity needs to be established based on raw data - currently we store only aggregates, with time points wich needs to be fetched between the activity/workout start/end
-  Some aggregates (especially max and averages) might be usefull in the summary - stuff like aggregate change over time or even aggregate for the period of summary

**Recommended solution**

- [ ] Extend the `SummariesService` to cover the activities feature, relying on the workout data. This assumes an analogous construction and fetching, but adapted for the type of data present in the workout table
- [ ] Add conversion from `total_elevation_gain` in the data model to  `floors_climbed` in summary, with additional raw field  `total_elevation_gain` - while floors might be more intuitive for the user, there will be use cases requiring detailed elevation
- [ ] Either add calculation for active and total calories based on the raw `energy_burned`  (more complicated, might be a long-term goal) or preserve only `total_calories_kcal`
- [ ] Add intensity calculation based on the granular time series data available in the time series for heart rate - ideally, this will be user-defined interval. This is a complex issue as you cannot simply use HR or steps, as those might mean different things for different baselines - I'd go for % of average HR or a metric like HROS (heart rate over steps - HR/(steps+1)) to establish what intensity should be assigned
- [ ] Add `active` and `sedentary` period calculation - this is a simpler aggregate that the intensity, as here we can assume an arbitrary level of steps for example
- [ ] Add aggregates in the summary with max/avr HR and steps as those might be important for the activity summary

Gasiek:
## Description

Implements full calculation and aggregation for the daily activity summary endpoint (`/users/{user_id}/summaries/activity`). Previously, this endpoint returned mostly null values. Now it aggregates time-series data and workout details to provide comprehensive daily activity metrics.

**Metrics implemented:**
- **Steps & Distance**: Aggregated from time-series data
- **Calories**: Active calories + basal calories from time-series (workout energy not double-counted)
- **Elevation**: Raw `elevation_meters` from workouts + calculated `floors_climbed` (prefers `flights_climbed` time-series, falls back to elevation/3m)
- **Heart Rate Stats**: Daily avg/max/min BPM from heart rate time-series
- **Active/Sedentary Minutes**: Based on step threshold (30 steps/min)
- **Intensity Minutes**: Light/moderate/vigorous zones based on age-derived max HR (220 - age)

**Technical improvements:**
- Added TypedDict return types for repository aggregate methods
- Extracted activity constants to module level for configurability
- Fixed zero-as-falsy bugs (0 now correctly returned instead of null)
- Updated frontend TypeScript types to match new schema

### Related Issue

Closes #267 

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

**Breaking changes:**
- `ActivitySummary` schema fields renamed/restructured (see frontend type updates)
- Fields now use `_minutes` suffix instead of `_seconds`

### General

### Backend Changes

- [x] `uv run ty check` passes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Start the application with `docker compose up -d` and run `make init` to seed data
2. Get a user ID and API key from the database
3. Call the activity summary endpoint with curl
4. Run backend tests: `cd backend && uv run pytest tests/api/v1/test_summaries.py -v`

**Expected behavior:**
- Returns daily activity summaries with all implemented metrics
- Metrics are null only when no data exists (0 returned when measured as zero)
- Intensity minutes calculated based on user's age (or default 190 max HR if no birth date)

## Screenshots

N/A - API endpoint changes

## Additional Notes

- **Double-counting prevention**: Workout energy and distance are NOT added to daily totals as they're typically already included in time-series data from providers
- **HR Zone thresholds**: Light (50-63%), Moderate (64-76%), Vigorous (77-93%) of max HR
- **Configurable constants**: All thresholds defined at module level in `summaries_service.py`
- **Frontend types updated**: `ActivitySummary` interface in `types.ts` now matches backend schema

### DIFF
--- a/backend/app/repositories/data_point_series_repository.py
+++ b/backend/app/repositories/data_point_series_repository.py
@@ -1,7 +1,8 @@
-from datetime import datetime
+from datetime import date, datetime
+from typing import TypedDict
 from uuid import UUID
 
-from sqlalchemy import Date, asc, cast, func, tuple_
+from sqlalchemy import Date, asc, case, cast, func, literal_column, tuple_
 
 from app.database import DbSession
 from app.models import DataPointSeries, ExternalDeviceMapping
@@ -14,6 +15,44 @@ from app.schemas import (
 )
 from app.schemas.series_types import SeriesType, get_series_type_from_id, get_series_type_id
 from app.utils.pagination import decode_cursor
+
+
+class ActivityAggregateResult(TypedDict):
+    """Result from daily activity aggregation query."""
+
+    activity_date: date
+    provider_name: str
+    device_id: str | None
+    steps_sum: int
+    active_energy_sum: float
+    basal_energy_sum: float
+    hr_avg: int | None
+    hr_max: int | None
+    hr_min: int | None
+    distance_sum: float | None
+    flights_climbed_sum: int | None
+
+
+class ActiveMinutesResult(TypedDict):
+    """Result from daily active/sedentary minutes query."""
+
+    activity_date: date
+    provider_name: str
+    device_id: str | None
+    active_minutes: int
+    tracked_minutes: int
+    sedentary_minutes: int
+
+
+class IntensityMinutesResult(TypedDict):
+    """Result from daily intensity minutes query."""
+
+    activity_date: date
+    provider_name: str
+    device_id: str | None
+    light_minutes: int
+    moderate_minutes: int
+    vigorous_minutes: int
 
 
 class DataPointSeriesRepository(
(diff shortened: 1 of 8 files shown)


backend/app/repositories/data_point_series_repository.py
```diff
@@ -16,6 +17,44 @@
...
+class ActivityAggregateResult(TypedDict):
+    """Result from daily activity aggregation query."""
+
+    activity_date: date
+    provider_name: str
+    device_id: str | None
+    steps_sum: int
+    active_energy_sum: float
+    basal_energy_sum: float
+    hr_avg: int | None
```
FBegiello, line 29: This assumes rounding up the HR value when counting the `hr_avg` - in most cases should be ok, but there might be some precision levels that require 1 or 2 decimal points.

backend/app/schemas/summaries.py
```diff
@@ -11,17 +11,33 @@ class IntensityMinutes(BaseModel):
     vigorous: int | None = None
 
 
+class HeartRateStats(BaseModel):
+    """Heart rate statistics for a period."""
+
+    avg_bpm: int | None = None
```
FBegiello, line 17: ditto with comments regarding precision - fine for now

backend/app/services/summaries_service.py
```diff
@@ -20,6 +25,16 @@
...
 
+# Activity summary constants
+DEFAULT_MAX_HR = 190  # Assumes ~30 years old when birth_date unavailable
+ACTIVE_STEPS_THRESHOLD = 30  # Steps per minute to be considered "active"
+METERS_PER_FLOOR = 3.0  # Standard floor height for floors_climbed calculation
+
+# HR zone percentages (as fraction of max HR)
+HR_ZONE_LIGHT = (0.50, 0.63)  # 50-63% of max HR
+HR_ZONE_MODERATE = (0.64, 0.76)  # 64-76% of max HR
+HR_ZONE_VIGOROUS = (0.77, 0.93)  # 77-93% of max HR
```
FBegiello, line 36: We can consider moving this in the futere into a general config file

backend/app/services/summaries_service.py
```diff
@@ -28,6 +43,39 @@ def __init__(self, log: Logger):
         self.logger = log
         self.event_record_repo = EventRecordRepository(EventRecord)
         self.data_point_repo = DataPointSeriesRepository(DataPointSeries)
+        self.user_repo = UserRepository(User)
+
+    def _get_user_max_hr(self, db_session: DbSession, user_id: UUID, reference_date: datetime) -> int:
+        """Calculate user's max HR based on age.
+
+        Uses formula: max_hr = 220 - age
+        Falls back to DEFAULT_MAX_HR if birth_date is not available.
```
FBegiello, line 52: We can consider adding multiple calculation methods, especially since the base `220-age` can be innacurate.
We can set this for example as bool with `high_precision` parameter, where current version is considered low precision, and for high precision you use:
- **for male** _208-(0.7age)_ per Tanaka
- **for female** _206-(0.88age)_ per Gulati

PR #240, opened 2026-01-04
feat(mcp-server): Implement Model Context Protocol server for querying user data

healthkowshik:
## Description

- Added `mcp-server` service to `docker-compose.yml` for running the MCP server.
- Introduced new modules for the MCP server and user tools in `users.py`.
- Implemented user-related tools for listing and retrieving user details.
- Updated `pyproject.toml` to include `fastmcp` as a dependency.
- Enhanced `uv.lock` with new package versions and dependencies for the MCP server functionality.

### Related Issue

Ref: https://github.com/the-momentum/open-wearables/issues/239

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Other (please describe):

### General

- [x] My code follows the project's code style
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix/feature works (if applicable)
- [ ] New and existing tests pass locally

### Backend Changes

- [x] `uv run ruff check` passes
- [x] `uv run ruff format --check` passes
- [ ] `uv run ty check` passes

## Testing Instructions

**Steps to test:**
1. Bring up the MCP server together with other services with `docker compose watch`
2. View logs to confirm FastMCP server is up and running with `docker compose logs -f`
3. Come to Cursor or Claude Desktop and add the MCP server details.

```json
{
  "mcpServers": {
    "open-wearables": {
      "url": "http://localhost:8001/sse"
    }
  }
}
```

```bash
# Path to cursor MCP server configs.
vim ~/.cursor/mcp.json

# Path to Claude Desktop MCP server configs.
vim ~/Library/Application Support/Claude/claude_desktop_config.json
```

**Expected behavior:**

**MCP server listed in Cursor settings**
 
image

**Tool 1. List users**

image

**Tool 2. Get user details**

image

Issue #239, opened 2026-01-02
Introduce MCP Server


bartmichalak:
One of the steps we should take to make AI agent integration with Open Wearables easier is to provide an MCP server for Open Wearables.
The MCP server should be created with [FastMCP](https://gofastmcp.com/getting-started/welcome). 

We can try to generate it using [python-ai-kit](https://github.com/the-momentum/python-ai-kit) which has this capability, but I'd be careful since this feature hasn't been tested very well by us.

The server should be easy to deploy - for experiments we can use [fastmcp.cloud](https://fastmcp.cloud/). I've deployed a few servers there in the past and it's super convenient - GitHub integration makes it possible to deploy in just 5 minutes!
Some questions that should be answerable by LLM clients using MCP servers:

**Open-Ended Query**
- What time of day do I feel most energized?
- How do I reduce stress?
- Do I sleep enough?
- How does my sleep compare to a year ago?
- How were my steps affected by sleep last month?

**Objective Query**
- What was my total percentage of light sleep over the past 21 days?
- How many times have I done aerobic workouts in the last 30 days?
- What was the standard deviation of my sleep duration over the past 21 days?

I think we can try to generate a first shot using Claude Code or Codex (with higher-tier models) - from my experience they've handled this pretty well.

A good inspiration could be the [apple-health-mcp](https://github.com/the-momentum/apple-health-mcp-server) project we worked on a while back.

PR #287, opened 2026-01-12
feat(mcp): add initial implementation of Open Wearables MCP server

healthkowshik:
First attempt here: https://github.com/the-momentum/open-wearables/pull/240 was not successful. Over the weekend, put some more time into learning MCPs, apple-health-mcp-server and healthion and this second attempt has come out much better. Thank you @KaliszS for your feedback on the original PR, gave me an opportunity to go deeper into MCP'land 🤖 

---

## Description
- [x] Implemented core application structure with configuration, main entry point, and API client.
- [x] Added tools for querying users and sleep records.
- [x] Included example `.env` file for environment configuration.

### Related Issue

Makes progress on: https://github.com/the-momentum/open-wearables/issues/239

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

### General

### Backend Changes

- [x] `uv run ty check` passes

## Screenshots

> Who can I query health data for?

image

> How much sleep did Laura get in the last few weeks?

image


mcp/app/config.py
```diff
@@ -0,0 +1,51 @@
...
+
+import sys
+from pathlib import Path
+
+from pydantic import Field, SecretStr, ValidationError
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+# Resolve the absolute path to the .env file
+# This ensures it works regardless of the current working directory
+_ENV_FILE = (Path(__file__).parent.parent / "config" / ".env").resolve()
```
FBegiello: I'd move this to the setting class itself, just like the core OW config

```
    model_config = SettingsConfigDict(
        env_file=str(Path(__file__).parent.parent / "config" / ".env"),
        env_file_encoding="utf-8",
        extra="ignore",
        # default env_file solution search .env every time BaseSettings is instantiated
        # dotenv search .env when module is imported, without usecwd it starts from the file it was called
    )
```

mcp/app/main.py
```diff
@@ -0,0 +1,51 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+# Create FastMCP server instance
+mcp = FastMCP(
+    "open-wearables",
+    instructions="""
+    Open Wearables MCP Server - Query wearable health data from multiple providers.
+
```
FBegiello: This is a good start, but in general I'm all for more detailed tool descriptions - you might expand this to avoid implicit assumptions, LLMs usually get them right but not always.
I'd be more explicit in stating something like:
`"""
Enables the model to query data describing user health states and general metrics. Data is acquired from the specific users wereable devices, covering all user related devices and providers, aggregating them into single format ...
"""`
A tip here is to use LLMs to refine prompts for LLMs as they generally form a very solid baseline.

mcp/app/main.py
```diff
@@ -0,0 +1,51 @@
...
+    Available tools:
+    - list_users: Discover users accessible via your API key
+    - get_sleep_records: Get sleep data for a user over the last X days
+
+    Workflow:
+    1. If you don't know the user's ID, call list_users first
+    2. Use the user's ID (or name) to query their health data
+    3. Present the data in a human-friendly format
```
FBegiello: Here you can give examples - LLMs react great to specific examples and patterns (see few-shot or one-shot prompting). This ensures a highl level of control over the desired output

mcp/app/main.py
```diff
@@ -0,0 +1,51 @@
...
+    "open-wearables",
+    instructions="""
+    Open Wearables MCP Server - Query wearable health data from multiple providers.
+
+    Available tools:
+    - list_users: Discover users accessible via your API key
+    - get_sleep_records: Get sleep data for a user over the last X days
+
+    Workflow:
+    1. If you don't know the user's ID, call list_users first
```
FBegiello: If there will be an issue with handling IDs in testing, you can try to be more explicit like "`call list_users tool first and use returned user IDs to determine the current user ID"`. This is only a potential future thing to consider

mcp/app/tools/users.py
```diff
@@ -0,0 +1,65 @@
...
+from app.services.api_client import client
+
+logger = logging.getLogger(__name__)
+
+
+async def list_users(search: str | None = None) -> dict:
+    """
+    List all users accessible via the configured API key.
+
+    Use this tool to discover available users before querying their health data.
```
FBegiello: I'd specify here that the user listing is explicitly for Open Wereables integration - might not be an issue, unless someone uses this in conjunction with some other MCP tool using users

mcp/app/tools/users.py
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    "id": str(user.get("id")),
+                    "first_name": user.get("first_name"),
+                    "last_name": user.get("last_name"),
+                    "email": user.get("email"),
+                }
+                for user in users
+            ],
+            "total": response.get("total", len(users)),
+        }
```
FBegiello, line 68: You might consider handling the user number  - for small services this will be fine, but if you call a database with for example 10k users you might easily burn a lot of token for this tool call.

Maybe a limiting parameter in the tool query or some chunking?
Splitting this into tools like `fetch_top_n_users` and `fetch_all_users` that needs to be added explicitly might be helpful - I'm thinking of the use case when, in generall, user will be asking about his/hers data. Fetching all users might be usefull mostly for some superuser (coach of a team)

mcp/app/tools/sleep.py
```diff
@@ -0,0 +1,238 @@
...
+            duration = record.get("duration_minutes")
+            if duration is not None:
+                durations.append(duration)
+
+            source = record.get("source", {})
+            records.append(
+                {
+                    "date": str(record.get("date")),
+                    "start_time": _format_time(record.get("start_time")),
+                    "end_time": _format_time(record.get("end_time")),
```
FBegiello: I'd be careful with returning start-end times without explicitly stating that they refer to different days. For a human this is obvious, but for an LLM it might not be. I'm worried how this will behave in some specific scenarios and whether there is a risk of mixup when analysing that with smaller LLMs for example

mcp/app/tools/sleep.py
```diff
@@ -0,0 +1,238 @@
...
+                            "first_name": u.get("first_name"),
+                            "last_name": u.get("last_name"),
+                        }
+                        for u in users[:10]  # Limit to first 10
+                    ],
+                    "total_users": len(users),
+                }
+
+        # Step 2: Calculate date range
+        end_date = datetime.now(UTC)
```
FBegiello: This is fine for the tool intention - fetch last n days. What I see as potential addition is a tool to fetch from dat X to date Y - adds more flexibility to usage, especialy for more complex queries

mcp/app/tools/sleep.py
FBegiello, line 1: As per a comment on line 169, I'd be strongly for adding also a tool to retrieve data for specified period from staret to end day, as an addition to the tool for last n days, or possible extension to this one. There might be a lot of cases in which you need to grab data starting from other points in time than from today

This should be quite straightforward but adds a lot of value for the end user

mcp/app/main.py
```diff
@@ -0,0 +1,51 @@
...
+logger.info(f"Open Wearables MCP server initialized. API URL: {settings.open_wearables_api_url}")
+
+
+def main() -> None:
+    """Entry point for the MCP server."""
+    mcp.run()
+
+
+if __name__ == "__main__":
+    main()
```
FBegiello: Keeping in mind future developement, I'd be for separating this into a router based architecture as we have for example in https://github.com/the-momentum/python-ai-kit, for the MCP

What I mean is to keep tools separated with routers, split per the type of data handled.
Something like:

```
from fastmcp import FastMCP

hello_router = FastMCP(name="Hello MCP")


@hello_router.tool
def hello(name: str) -> str:
    """Say hello to user."""
    return f"Hello {name}, what's up!"
```

and then


```
from fastmcp import FastMCP

from app.mcp.tools import hello

mcp_router = FastMCP(name="Main MCP")

mcp_router.mount(hello.hello_router)
```

mcp/app/config.py
healthkowshik: Yes. Made the change here: https://github.com/the-momentum/open-wearables/pull/287/changes/b2d9d285b4b6fda5ee834059133e1830153e1e09

mcp/app/main.py
healthkowshik: Improved the descriptions a bit more in this commit: https://github.com/the-momentum/open-wearables/pull/287/changes/7c4438ce1b33e39db2f16adb6c97cd0b74ccd46e

mcp/app/main.py
healthkowshik: Good feedback. Added a few examples here: https://github.com/the-momentum/open-wearables/pull/287/changes/ca7565e2e20ea1d59cc8be1b9d9fd1af1728e860

mcp/app/main.py
healthkowshik: Made a small update for this here: https://github.com/the-momentum/open-wearables/pull/287/changes/a1d8c0c168fb460c378f2f41f45a89c5c0837aed

mcp/app/tools/users.py
healthkowshik: Added `open wearables` here: https://github.com/the-momentum/open-wearables/pull/287/changes/abe302447dfc72e59aa620ceaec70698efcc916e

mcp/app/tools/users.py
healthkowshik: Added the `limit` parameter here: https://github.com/the-momentum/open-wearables/pull/287/changes/e7d5e001ec726d42ad366b9b84795fe7e98959f8

mcp/app/tools/sleep.py
healthkowshik: Keeping them ISO datetime strings for now then: https://github.com/the-momentum/open-wearables/pull/287/changes/649ddcd9a86cc956744196c6fa0543945076b023

mcp/app/tools/sleep.py
healthkowshik: Removed `n days` and instead switched to `start_date` and `end_date` here: https://github.com/the-momentum/open-wearables/pull/287/changes/049bab636b1c98f01792fd4d9bb9be96e14c64ef

mcp/app/main.py
healthkowshik: Made this change here: https://github.com/the-momentum/open-wearables/pull/287/changes/203fcd2e187725f329af3b01d62096b13d9a34fc Thank you for the reference.

mcp/app/tools/sleep.py
healthkowshik: Removed `n days` and instead switched to `start_date` and `end_date` here: https://github.com/the-momentum/open-wearables/commit/049bab636b1c98f01792fd4d9bb9be96e14c64ef

Issue #281, opened 2026-01-09
HRV Data Integrity: Suunto RMSSD being stored as SDNN series type


Gasiek:
### Description

We have a data integrity issue where Suunto's HRV data (which is RMSSD) is being incorrectly stored under the `heart_rate_variability_sdnn` series type.

### Current Behavior

| Provider | Metric Provided | Stored As |
|----------|-----------------|-----------|
| Apple Health | SDNN | `heart_rate_variability_sdnn` ✅ |
| Suunto | RMSSD | `heart_rate_variability_sdnn` ❌ |

Two clinically different metrics are being mixed into the same series type.

### Evidence

- **Apple Health**: Uses `HKQuantityTypeIdentifierHeartRateVariabilitySDNN` (explicit SDNN)
- **Suunto**: [Official documentation](https://www.suunto.com/pl-pl/sports/News-Articles-container-page/how-to-use-hrv-to-optimize-your-recovery/) confirms they use RMSSD:
  > "Continuous HRV monitoring allows calculating the root mean square of successive differences (RMSSD)"

### Why This Matters

SDNN and RMSSD measure different things:
- **SDNN** (Standard Deviation of NN intervals): Overall HRV, reflects both sympathetic and parasympathetic activity
- **RMSSD** (Root Mean Square of Successive Differences): Short-term HRV, primarily reflects parasympathetic (vagal) activity

Mixing them makes HRV data unreliable for health analysis.

### Proposed Solution

1. Add new series type: `heart_rate_variability_rmssd` (ID: 6)
2. Update Suunto provider to use the new RMSSD type
3. Consider adding both `avg_hrv_sdnn_ms` and `avg_hrv_rmssd_ms` to summary schemas
4. Optionally: migrate existing Suunto HRV data to the correct series type

### Affected Files

- `backend/app/schemas/series_types.py` - Add new enum and definition
- `backend/app/services/providers/suunto/data_247.py` - Line 461, change to use RMSSD type
- `backend/app/schemas/summaries.py` - Consider adding RMSSD field
- `backend/scripts/init/series_type_config.csv` - Add RMSSD config
- Database migration required for new series type

czajkub:
resolved some long time ago

PR #300, opened 2026-01-13
feat: body summary api

### LINKED ISSUE #269: Data summaries endpoint: Personal record and body states summary
**Issue**
Current summary is a draft approach, not connected directly to existing datatypes in the time series (we are working under the assumption that in current version slow changing measurments like body weight are still stored as time series)

- Dedicated method in the service needs to be built to fetch body describing features, with separation into static values (from `personal record` table covering sex and gender, and age) as well as fetching newest recording from time series data (weight, height, etc.)
- No aggregation or filtering for slow-changing measurements (weight, body fat %)
- No compound statistic calculation for body descriptor (i.e. BMI from weight and height)
- No aggregation for highly granular statistics like HR, which need to be fetched as aggregates (i.e. resting HR)
- Currently blood pressure is present as arbitrary blood presure, which is not accurate as this should be an aggregate as well (average of sets of measurments for a period, measure taken once is not representative)

**Recommended solution**
- [ ] Build dedicated body summary method in the `SummaryService`
- [ ] Add basic age calculation from birthdate
- [ ] Add base handling for data that can be fetched directly, from slow changing time series, as last present value  (i.e. weight, body fat %)
- [ ] Add calculating static values from last present mesurments (i.e BMI)
- [ ] Add aggregators for general statistic like HR, resting HR etc. - we should assume an arbitrary period, for example last 30 days, or last 7 days, possibly both for each
- [ ] Rename blood pressure to average and/or add max/min values for blood pressure - current value of generic  "blood_pressure" is not representative of a period of time

Gasiek:
## Description

Implement body summary endpoint that aggregates body composition metrics, vitals, and calculated health indicators into a daily summary format.

**Features:**
- Age calculation from birthdate (PersonalRecord)
- Slow-changing measurements fetched as latest values: weight, height, body fat %, lean mass, temperature
- BMI calculation from weight and height
- 7-day rolling aggregates for vitals: resting heart rate, HRV
- Comprehensive blood pressure stats: avg, min, max for both systolic and diastolic with reading count

**Fixes included:**
- Use `DISTINCT ON` to handle duplicate timestamps deterministically
- Use outer join for WorkoutDetails to include all workouts in aggregates
- Implement proper cursor-based pagination for activity summaries
- Rename HRV field from `rmssd` to `sdnn` to match actual data source
- Fix pagination filtering order issue in body summaries

### Related Issue

Closes #269 

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

### General

### Backend Changes

- [ ] `uv run ty check` passes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Start the development environment: `docker compose up -d`
2. Seed sample data: `make init`
3. Call the body summary endpoint:
   curl "http://localhost:8000/api/v1/users/{user_id}/summaries/body?start_date=2025-01-01T00:00:00Z&end_date=2025-01-07T23:59:59Z"

### Expected behavior:

- Returns daily body summaries with weight, height, BMI, body composition, and vitals
- Age is calculated from user's birthdate
- Blood pressure includes avg/min/max for systolic and diastolic
- Empty days (no data) are excluded from results
- Pagination works correctly with cursor parameter

## Screenshots

N/A - API endpoint only

## Additional Notes

- Sex/gender fields from PersonalRecord deferred for future work
- 30-day aggregates (in addition to 7-day) deferred - can be added later if needed
- Known issue: Suunto stores RMSSD data as SDNN series type (documented with TODO comment and issue #281 )
- TODO added for date range interval standardization in repository methods

### DIFF
--- a/backend/tests/repositories/test_data_point_series_repository.py
+++ b/backend/tests/repositories/test_data_point_series_repository.py
@@ -313,7 +313,10 @@ class TestDataPointSeriesRepository:
             assert sample.series_type_definition_id == expected_type_id
 
     def test_get_samples_by_date_range(self, db: Session, series_repo: DataPointSeriesRepository) -> None:
-        """Test filtering samples by date range."""
+        """Test filtering samples by date range.
+
+        Uses half-open interval [start, end) - start is inclusive, end is exclusive.
+        """
         # Arrange
         user = UserFactory()
         mapping = ExternalDeviceMappingFactory(user=user, device_id="device1")
@@ -336,21 +339,23 @@ class TestDataPointSeriesRepository:
             )
             series_repo.create(db, sample)
 
+        # Query with half-open interval [yesterday, now + 1s) to include both yesterday and now
+        end_datetime = now + timedelta(seconds=1)
         query_params = TimeSeriesQueryParams(
             device_id="device1",
             start_datetime=yesterday,
-            end_datetime=now,
+            end_datetime=end_datetime,
         )
 
         # Act
         results, total_count = series_repo.get_samples(db, query_params, [SeriesType.heart_rate], user.id)
 
-        # Assert
+        # Assert - should get yesterday and now (2 samples), excluding two_days_ago
         assert len(results) == 2
         assert total_count == 2
         for sample, _ in results:
             assert sample.recorded_at >= yesterday
-            assert sample.recorded_at <= now
+            assert sample.recorded_at < end_datetime
 
     def test_get_samples_by_provider(self, db: Session, series_repo: DataPointSeriesRepository) -> None:
         """Test filtering samples by provider ID."""


backend/app/utils/pagination.py
```diff
@@ -134,3 +134,108 @@ def process_paginated_results(
         previous_cursor=previous_cursor,
         has_more=has_more,
     )
+
+
+def encode_date_cursor(cursor_date: date, direction: str = "next") -> str:
```
KaliszS, line 177: Why so many new pagination utils functions?

Gasiek: We needed new cursor functions because summaries don't have row IDs like time-series data does. The existing cursor uses timestamp + UUID, but summaries are aggregated by date - there's no individual row ID to reference.
For body summaries, we just need the date. For activity summaries, we need date + provider + device because a user can have Garmin and Apple data on the same day. If we only used the date, pagination would skip records after the first provider.
The current solution works but it's not ideal - we basically copy-pasted the same encoding logic three times with different parameters. If we add more summary types, we'd keep adding more functions.
A cleaner approach would be a single generic cursor function that takes a list of fields and encodes them. Then each summary type just passes whatever fields it needs without us writing new encode/decode pairs each time.
So instead of adding new util functions, we could change this:
```
def encode_cursor(timestamp: datetime, item_id: UUID, direction: str = "next") -> str:
    cursor_str = f"{timestamp.isoformat()}|{item_id}"
    encoded = base64.urlsafe_b64encode(cursor_str.encode("utf-8")).decode("utf-8")

    if direction == "prev":
        return f"prev_{encoded}"
    return encoded
```
To this:
```
def encode_generic_cursor(fields: list[str], direction: str = "next") -> str:
    cursor_str = "|".join(fields)
    encoded = base64.urlsafe_b64encode(cursor_str.encode()).decode()
    
    if direction == "prev":
        return f"prev_{encoded}"
    return encoded
```


We will still need the new util functions, but they will use the `encode_generic_cursor` so there won't be code duplication.

backend/app/schemas/summaries.py
```diff
@@ -60,27 +60,48 @@ class SleepSummary(BaseModel):
...
 class BloodPressure(BaseModel):
-    systolic_mmhg: int | None = None
-    diastolic_mmhg: int | None = None
+    """Blood pressure statistics aggregated over a period.
+
+    Values are aggregated from multiple readings to provide a more representative measure.
+    """
+
+    avg_systolic_mmhg: int | None = Field(None, description="Average systolic pressure", example=120)
+    avg_diastolic_mmhg: int | None = Field(None, description="Average diastolic pressure", example=80)
```
FBegiello, line 75: Same issue as previously with integer values for averages - while the source value (like HR for example) is registered as integer for sure, I'm still on the fence when it comes to averages. Rounding is a form of data loss that I'm not sure we need to enforce.
Can stay for now, but we need to discuss this as a general approach for the project

backend/app/repositories/data_point_series_repository.py
```diff
@@ -226,6 +226,10 @@ def get_averages_for_time_range(
         """Get average values for specified series types within a time range.
 
         Returns a dict mapping SeriesType to average value (or None if no data).
+
+        TODO: This method uses closed interval [start, end] (<=) while other methods
+        use half-open [start, end) (<). Consider standardizing on half-open intervals
+        for consistency across the repository.
```
FBegiello: This is important for the consistency of use across the platform

backend/app/repositories/data_point_series_repository.py
```diff
@@ -572,3 +576,132 @@ def get_daily_intensity_minutes(
...
+        Used for slow-changing measurements like weight, height, body fat %.
+
+        Args:
+            before_date: Only consider measurements recorded before this datetime
+
+        Returns:
+            Dict mapping SeriesType to tuple of (value, recorded_at, provider_name, device_id)
+        """
+        if not series_types:
+            return {}
```
FBegiello: We could return an error here - empty dict might be misleading as empty query is not always indicative of failed query when retrieving data (i.e. no data for the period != error of execution for the period)

backend/app/repositories/data_point_series_repository.py
```diff
@@ -572,3 +576,132 @@ def get_daily_intensity_minutes(
...
+            .join(ExternalDeviceMapping, self.model.external_device_mapping_id == ExternalDeviceMapping.id)
+            .join(
+                latest_subq,
+                (self.model.series_type_definition_id == latest_subq.c.series_type_definition_id)
+                & (self.model.recorded_at == latest_subq.c.max_recorded_at),
+            )
+            .filter(ExternalDeviceMapping.user_id == user_id)
+            # DISTINCT ON (PostgreSQL) ensures exactly one result per series type
+            # Order by id desc as tiebreaker for deterministic selection
+            .distinct(self.model.series_type_definition_id)
```
FBegiello, line 814: This needs a closer look int he future - we need to implement a provider/device reliability rating and select based on that for deterministic, controlled and predictable selection.

As of now, I'm worried about the predictable  part, as selecting by ID does not guarantee us what type od device is being prioritized. This is a level of opacity. 
While we can accept this in this release, I'd be stronly in favour of addressing that sooner or later

FBegiello (approved):

KaliszS (approved):

PR #344, opened 2026-01-20
feat: generate user sdk token by admin

bartmichalak:
## Description

It adds the ability for administrators to generate SDK tokens for users directly from the user details page. 

The generated token has an infinite expiration time since we currently don't support refresh tokens.

This will help with onboarding users to the demo app that we will be releasing alongside the SDK directly from the developer portal.

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Log into the dashboard and go to the user's details
2. Generate a token
3. Use this token as a `bearer token` in the POST /api/v1/sdk/users/{user_id}/sync/apple/healthion endpoint

## Screenshots

https://github.com/user-attachments/assets/fdf95693-2b3d-4bdd-bb6b-1482d80ee51a

## Additional Notes

NA

### DIFF
--- a/frontend/src/routes/_authenticated/users/$userId.tsx
+++ b/frontend/src/routes/_authenticated/users/$userId.tsx
@@ -179,15 +179,13 @@ function UserDetailPage() {
   };
 
   const handleCopyToken = async () => {
-    if (tokenData?.access_token) {
-      const success = await copyToClipboard(
-        tokenData.access_token,
-        'Token copied to clipboard'
-      );
-      if (success) {
-        setTokenCopied(true);
-        setTimeout(() => setTokenCopied(false), 2000);
-      }
+    const success = await copyToClipboard(
+      tokenData?.access_token || '',
+      'Token copied to clipboard'
+    );
+    if (success) {
+      setTokenCopied(true);
+      setTimeout(() => setTokenCopied(false), 2000);
     }
   };
   if (!userLoading && !user) {


frontend/src/routes/_authenticated/users/$userId.tsx
```diff
@@ -150,6 +171,22 @@ function UserDetailPage() {
...
+  };
+
+  const handleCopyToken = async () => {
+    if (tokenData?.access_token) {
+      await navigator.clipboard.writeText(tokenData.access_token);
+      setTokenCopied(true);
+      toast.success('Token copied to clipboard');
+      setTimeout(() => setTokenCopied(false), 2000);
+    }
+  };
```
Gasiek, line 190: use copyToClipboard utility function from `frontend/src/lib/utils/clipboard.ts`. It handles errors when copying fails.

backend/app/api/routes/v1/sdk_token.py
```diff
@@ -1,46 +1,75 @@
...
 from app.database import DbSession
 from app.schemas.oauth import Token
 from app.schemas.sdk import SDKTokenRequest
 from app.services import application_service, create_sdk_user_token
+from app.utils.auth import DeveloperOptionalDep
 
 router = APIRouter()
 
+# Default app_id for admin-generated tokens
+ADMIN_APP_ID = "admin"
```
KaliszS: maybe it would fit better as part of settings?

KaliszS: What happens when two users try this method and both get default app id?

frontend/src/routes/_authenticated/users/$userId.tsx
```diff
@@ -237,6 +274,23 @@ function UserDetailPage() {
...
+                <Loader2 className="h-4 w-4 animate-spin" />
+                Generating...
+              </>
+            ) : (
+              <>
+                <Key className="h-4 w-4" />
+                Generate Token
+              </>
+            )}
+          </button>
```
Gasiek, line 294: I am not convinced that this blue color is a good choice. Is there a point of it not being white? Blue isn't really coherent with our design. We should switch to using Button with variants, but that's a thing for a different PR. I would change it to white (same as the ones next to it) for now.

backend/app/api/routes/v1/sdk_token.py
bartmichalak: Everything will work, but we're actually losing information about who generated that token (for credentials generated through the Application model, we retain information about who created the key).

To preserve this information, I suggest following:
`app_id = f"admin:{developer.id}"`

WDYT? Then I wouldn't move this to settings, we'll make it a variable. Alternatively we could move the prefix, but there's probably no point

frontend/src/routes/_authenticated/users/$userId.tsx
bartmichalak: Good catch, wasn't aware of that!

frontend/src/routes/_authenticated/users/$userId.tsx
bartmichalak: There was no point, maybe because I've always written backend code haha.

Changed to white. 

Button with variants seem to be a good option, would you create an issue for this change?

backend/app/api/routes/v1/sdk_token.py
KaliszS: Good idea! But after the change I believe it should stay right here ;) Settings are somehow static, so it woould loose information about developer.id

frontend/src/routes/_authenticated/users/$userId.tsx
Gasiek: Already created [https://github.com/the-momentum/open-wearables/issues/347](url)

backend/app/api/routes/v1/sdk_token.py
bartmichalak: Yeah, I only meant moving static prefix (`admin`):
`{settings.ADMIN_PREFIX}:{developer.id}` 

but i don't see a big value from this TBH.

frontend/src/routes/_authenticated/users/$userId.tsx
bartmichalak: image

looks better now 😎

frontend/src/routes/_authenticated/users/$userId.tsx
Gasiek: I think I would remove the `if (tokenData?.access_token)` because we won't get the toast error otherwise.

PR #349, opened 2026-01-20
feat(mcp): initial implementation of MCP server

KaliszS:
Original PR is here: #287 
Added this one because I messed up git history and have to do a soft reset.


KaliszS:
@healthkowshik I have changed a few details:

1. code quality (now it uses pre-commit, so you have to sync with `uv sync --group code-quality` instead of dev group and then you can run all checks with `uv run pre-commit run --all-files`)
2. upgrade python version (3.11 is old enough to be updated, I've set up 3.13, like on backend - you don't need to install anything hovewer, thanks to `.python-version` file, uv will detct and install required python version, that's why I removed python from project requirements at all)

PR #372, opened 2026-01-21
fix: redirect to dashboard after login instead of users page

bartmichalak:
## Description

Redirect authenticated users to the dashboard instead of the users page after login.

### Related Issue

NA

### General

- [ ] My code follows the project's code style

### Backend Changes

NA

### Frontend Changes

- [ ] `pnpm run lint` passes

## Testing Instructions

Steps to test:

**Test login redirect:**
1. Log out (if logged in)
2. Log in via /login
3. Verify redirect to /dashboard instead of /users

**Test registration redirect:**
1. Log out
2. Register a new account via /register
3. Verify redirect to /dashboard after registration

**Test root route redirect:**
1. While logged in, navigate to / (root)
2. Verify redirect to /dashboard

**Test auth page guards:**
1. While logged in, try accessing /login, /forgot-password, or /reset-password
2. Verify redirect to /dashboard (instead of allowing access to those pages)

**Test logout and re-login flow:**
1. Log out
2. Log back in
3. Verify redirect to /dashboard works correctly

## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/frontend/src/routes/users/$userId/pair.index.tsx
+++ b/frontend/src/routes/users/$userId/pair.index.tsx
@@ -92,7 +92,6 @@ function PairWearablePage() {
               size="icon"
               onClick={reset}
               aria-label="Dismiss error"
-              // className="text-red-400/70 hover:text-red-300 hover:bg-transparent"
             >
               <X className="w-5 h-5" />
             </Button>


KaliszS (changes requested):
It requires to change the same information over and over in different placs. Maybe we can create a global function out of that redirect or at least add setttings variable to where we want redirection in that cases, so in the future it would be enough to change that only once?

PR #393, opened 2026-01-26
feat: add request body validation and documentation for Apple Health Sync

bartmichalak:
## Description

The main goal of this PR was to add validation to the /sync/apple endpoint, but I also did a refactor because the code was hard to navigate (schemas scattered across multiple files, types as strings instead of enums). The following changes were made:                                                                     
                                                                                                                                                                  
**Schema consolidation**                                                                                                                      
  - Deleted files: `record_import.py`, `workout_import.py`, `source_info.py`
  - Everything consolidated into a single file: `sync_request.py`
  - Renamed `redis_sleep.py` -> `sleep_state.py` (clearer naming)
                                                                                                                                                                  
**StrEnum instead of Literal/string**                                                                                                         
  - `AppleMetricType` - `HKQuantityTypeIdentifier` types (heart_rate, steps, etc.)
  - `AppleCategoryType` - `HKCategoryTypeIdentifier` types (sleep_analysis)
  - `SDKWorkoutType` - SDK workout types (walking, running, cycling, etc.) - however, it requires additional refactor: https://github.com/the-momentum/open-wearables/issues/394
  - `WorkoutStatisticType` - workout statistic types (duration, distance, averageHeartRate, etc.)
                                                                                                                                                                  
**Separate record types**                                                                                                                    
- HealthRecord replaced with two separate models:                                                                                                               
  - MetricRecord (type: AppleMetricType) - for records array                                                                                                    
  - SleepRecord (type: AppleCategoryType) - for sleep array                                                                                                     
                                                                                                                                                                
Validation                                                                                                                                                 
- SleepRecord.value - range validation 0-5 (sleep phases)                                                                                 
                                                                                                                                                                  
Import service refactor                                                                                                                                                                                                                                                                                                           
- Match statement now uses enums instead of strings (e.g., `WorkoutStatisticType.AVERAGE_HEART_RATE` instead of "averageHeartRate") 

## Checklist

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

NA

## Screenshots

<img width="1151" height="669" alt="image" src="https://github.com/user-attachments/assets/969e8750-6d32-407d-92a1-ec68267e283d" />

## Additional Notes

NA

### DIFF
--- a/backend/app/schemas/apple/healthkit/sync_request.py
+++ b/backend/app/schemas/apple/healthkit/sync_request.py
@@ -6,7 +6,7 @@ from datetime import datetime
 from decimal import Decimal
 from typing import Any
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 from app.constants.series_types import AppleCategoryType, AppleMetricType
 from app.constants.workout_statistics import WorkoutStatisticType
@@ -16,16 +16,17 @@ from app.constants.workout_types import SDKWorkoutType
 class OSVersion(BaseModel):
     """Operating system version info from HealthKit source."""
 
+    model_config = ConfigDict(populate_by_name=True)
+
     major_version: int = Field(alias="majorVersion")
     minor_version: int = Field(alias="minorVersion")
     patch_version: int = Field(alias="patchVersion")
 
-    class Config:
-        populate_by_name = True
-
 
 class SourceInfo(BaseModel):
     """Source/device information for HealthKit records."""
+
+    model_config = ConfigDict(populate_by_name=True)
 
     name: str | None = None
     bundle_identifier: str | None = Field(default=None, alias="bundleIdentifier")
@@ -37,9 +38,6 @@ class SourceInfo(BaseModel):
     device_model: str | None = Field(default=None, alias="deviceModel")
     device_hardware_version: str | None = Field(default=None, alias="deviceHardwareVersion")
     device_software_version: str | None = Field(default=None, alias="deviceSoftwareVersion")
-
-    class Config:
-        populate_by_name = True
 
 
 class MetricRecord(BaseModel):


backend/app/schemas/apple/healthkit/sync_request.py
```diff
@@ -0,0 +1,168 @@
...
+    value: float | int
+
+
+class Workout(BaseModel):
+    """Schema for workout/exercise session from HealthKit."""
+
+    uuid: str | None = None
+    type: SDKWorkoutType | None = None
+    startDate: datetime
+    endDate: datetime
```
bartmichalak, line 87: should we probably use the `Field` type here, similarly to `SourceInfo`?

backend/app/schemas/apple/healthkit/sync_request.py
```diff
@@ -0,0 +1,168 @@
...
+    """Schema for JSON import format from HealthKit.
+
+    Represents a single health measurement (heart rate, steps, distance, etc.).
+    """
+
+    uuid: str | None = None
+    type: AppleRecordType | None = None
+    startDate: datetime
+    endDate: datetime
+    unit: str | None
```
bartmichalak, line 63: IMO there should also be an enum here

KaliszS: Will there be any logic, like converting betweem units? If we are going to implement any flow control (like if or match) on values, enum should be fine, otherwise I woukd just use plan Literal.

backend/app/schemas/apple/healthkit/sync_request.py
KaliszS: Could you expand on that a bit?

backend/app/schemas/apple/healthkit/sync_request.py
```diff
@@ -0,0 +1,168 @@
...
+    product_type: str | None = Field(default=None, alias="productType")
+    operating_system_version: OSVersion | None = Field(default=None, alias="operatingSystemVersion")
+    device_name: str | None = Field(default=None, alias="name")
+    device_manufacturer: str | None = Field(default=None, alias="deviceManufacturer")
+    device_model: str | None = Field(default=None, alias="deviceModel")
+    device_hardware_version: str | None = Field(default=None, alias="deviceHardwareVersion")
+    device_software_version: str | None = Field(default=None, alias="deviceSoftwareVersion")
+
+    class Config:
+        populate_by_name = True
```
KaliszS: It's deprecated since Pydantic v2. Now configuration is resolved within model_config field and ConfigDict class.

backend/app/schemas/apple/healthkit/sync_request.py
bartmichalak: For consistency with Python naming conventions (snake_case) and with SourceInfo in the same file. Fields like startDate would become start_date = Field(alias="startDate"). 

This also lets us remove the # ruff: noqa: N815 at the top.

We should definatelly take one approach here, now we have mix-up 😀

backend/app/schemas/apple/healthkit/sync_request.py
bartmichalak: it has been generated by Claude, which is concerning since we already have a rule for using Pydantic V2 syntax in `agents.md` for backend:
https://github.com/the-momentum/open-wearables/blob/main/backend/AGENTS.md#schemas-layer-appschemas

backend/app/schemas/apple/healthkit/sync_request.py
bartmichalak: I think any conversion logic should be implemented in a separate service rather than at this level. 

However, after thinking about it for a while, I believe it's a much broader topic that we need to consider. I have created a separate issue for tracking it. 

https://github.com/the-momentum/open-wearables/issues/424

For now, I would leave it as it is.

WDYT?

PR #401, opened 2026-01-27
docs: add uv lock instruction to backend AGENTS.md

bartmichalak:
## Description

As you can see in [this PR](https://github.com/the-momentum/open-wearables/pull/398), the agents don't seem to realise that after modifying the `.toml` file, you need to run the uv lock command to update `uv.lock`.

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

NA

## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -217,9 +217,6 @@ make downgrade                             # Rollback
 ```bash
 cd backend
 
-# Add new dependencies (updates pyproject.toml, lockfile, and venv)
-uv add <package-name>
-
 # Lint and format (run after changes)
 uv run ruff check . --fix && uv run ruff format .
 


bartmichalak:
Actually, uv lock without any flags should be 'safe' by default. From the docs:

"With an existing uv.lock file, uv will prefer the previously locked versions of packages when running uv sync and uv lock. Package versions will only change if the project's dependency constraints exclude the previous, locked version."

And also:

"uv will not consider lockfiles outdated when new versions of packages are released — the lockfile needs to be explicitly updated if you want to upgrade dependencies."

So running plain uv lock after adding new packages to pyproject.toml will only add those new packages while preserving existing locked versions. You'd need --upgrade or --upgrade-package to actually change existing versions.

PR #405, opened 2026-01-27
feat: add reset_db target to Makefile and implement reset_database script

bartmichalak:
## Description

During development, I often reset the database, so it's worth having a single command at hand. I haven't found anything built into SQLAlchemy to handle this. There might be a smarter way to do it, but this works fine, and I don't think we need to pay too much attention to the implementation here. :)

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

Just run the script 

`make reset_db`

**Expected behavior:**

Truncates all database tables. 

## Screenshots

image

## Additional Notes

NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,3 +48,6 @@ create_migration:  ## Create a new migration. Use 'make create_migration m="Desc
 
 downgrade:  ## Revert the last migration
 	$(DOCKER_EXEC) $(ALEMBIC_CMD) downgrade -1
+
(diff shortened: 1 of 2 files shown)


KaliszS (commented):
It just deletes all rows, but leaves tables in place. Was that the goal of the PR?

Because when we want to deploy new migration, it often requried to drop all tables (because some sql column types are impossible to convert). If we would try to do this here, we would have to drop them in order, because of relations between them.

PR #390, opened 2026-01-24
fix(backend): batched inserts

elmariachi111:
## Description

we ran into some major issues when importing a >250M apple health XML file. This PR

- batches insertions into smaller chunks (postgres simply failed when celery tried inserting data)
- logs field value transactions a bit better
- fails more gracefully, doesnt fail the full import, when some field cant be read

### General

- [🤷  ] My code follows the project's code style
- [x ] I have performed a self-review of my code (this is Opus 4.5 on Cursor.)
- [- ] I have added tests that prove my fix/feature works (if applicable)
- [- ] New and existing tests pass locally

### Backend Changes

You have to be in `backend` directory to make it work:
- [ ] `uv run pre-commit run --all-files` passes

> too lazy to think about this rn 👇 
> error: Failed to spawn: `pre-commit`
>  Caused by: No such file or directory (os error 2)

### Frontend Changes

- [ ] `pnpm run lint` passes
- [ ] `pnpm run format:check` passes
- [ ] `pnpm run build` succeeds

## Testing Instructions

**Steps to test:**
1. import a really large file
2. it works
3.

PR #376, opened 2026-01-21
refactor(backend): user connection & device data model simplification

KaliszS:
Fixes #375 

Added mapping for Apple devices

PR #410, opened 2026-01-29
refactor(integrations): add trace IDs, batch inserts, and clean up webhook logging for Garmin integration

### LINKED ISSUE #448: Garmin sync: duplicate key violations on data_point_series due to missing ON CONFLICT
## Problem

Garmin webhook/backfill processing throws `duplicate key value violates unique constraint "uq_data_point_series_source_type_time"` when concurrent requests try to insert the same data points.

## Root cause

`data_247.py` saves each data point via individual `DataPointSeriesRepository.create()` → `db.add()` + `db.commit()`, which handles duplicates by catching `IntegrityError` after the fact. Under concurrent webhook deliveries or parallel backfill + webhook, this causes race conditions and logged errors.

Affected methods: `save_dailies_data()`, `save_epochs_data()`, `save_body_composition()`, `save_hrv_data()`.

## Fix suggestion (please verify)

Switch Garmin wellness data to use the existing `DataPointSeriesRepository.bulk_create()` path (batch `INSERT ... ON CONFLICT DO NOTHING`), same as Apple Health already does via `bulk_create_samples()`.

farce1:
## Description

Refactor the Garmin integration backend: add per-request trace IDs for observability, switch to chunked batch inserts for webhook data processing, enforce Garmin's 100 req/min rate-limit budget with backoff, and migrate to structured logging. Removes the `pull_token.py` module (programmatic token generation is forbidden by Garmin). On the frontend, the connection card now shows real-time sync progress and a new source badge component is added.

Fixes #448

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Connect a Garmin account via OAuth and verify the flow completes
2. Trigger a backfill and check logs for trace IDs and rate-limit delays
3. Send a webhook PUSH payload and confirm data is batch-inserted correctly
4. Open the connection card in the frontend and verify sync progress updates in real time

**Expected behavior:**
- Every webhook request and backfill chunk carries a unique trace ID in logs
- Batch inserts use chunked writes and upsert on conflict (no duplicate rows)
- Backfill respects the 100 req/min budget with calculated delays between batches
- Frontend connection card shows live sync progress during a backfill

## Additional Notes

- `pull_token.py` was removed — Garmin forbids programmatic pull-token generation
- Backfill handler moved from `providers/garmin/backfill.py` to `providers/garmin/handlers/backfill.py`
- Backend tests require a running PostgreSQL instance (CI provides this via Docker)
- Frontend has no test files yet, so vitest exits with code 1 (not a code issue)

### DIFF
--- a/frontend/src/hooks/api/use-users.ts
+++ b/frontend/src/hooks/api/use-users.ts
@@ -132,17 +132,17 @@ export function useDeleteUser() {
   });
 }
 
-export function useGenerateUserToken() {
+export function useGenerateInvitationCode() {
   return useMutation({
-    mutationFn: (userId: string) => usersService.generateToken(userId),
+    mutationFn: (userId: string) => usersService.generateInvitationCode(userId),
     onSuccess: () => {
-      toast.success('User token generated successfully');
+      toast.success('Invitation code generated successfully');
     },
     onError: (error: unknown) => {
       const message =
         error instanceof Error
           ? error.message
-          : 'Failed to generate user token';
+          : 'Failed to generate invitation code';
       toast.error(message);
     },
   });
--- a/frontend/src/lib/api/config.ts
+++ b/frontend/src/lib/api/config.ts
@@ -23,7 +23,8 @@ export const API_ENDPOINTS = {
     `/api/v1/users/${userId}/import/apple/xml/direct`,
   userAppleXmlPresignedUrl: (userId: string) =>
     `/api/v1/users/${userId}/import/apple/xml/s3`,
-  userToken: (userId: string) => `/api/v1/users/${userId}/token`,
+  userInvitationCode: (userId: string) =>
+    `/api/v1/users/${userId}/invitation-code`,
 
   // OAuth endpoints
   oauthAuthorize: (provider: string) => `/api/v1/oauth/${provider}/authorize`,
--- a/frontend/src/lib/api/services/users.service.ts
+++ b/frontend/src/lib/api/services/users.service.ts
@@ -9,7 +9,7 @@ import type {
   PaginatedUsersResponse,
   PresignedURLRequest,
   PresignedURLResponse,
-  Token,
+  InvitationCode,
 } from '../types';
 
 export const usersService = {
@@ -97,9 +97,8 @@ export const usersService = {
     }
   },
 
-  async generateToken(userId: string): Promise<Token> {
-    // Send no body - admin authentication via Bearer token will be used
-    const endpoint = API_ENDPOINTS.userToken(userId);
-    return apiClient.post<Token>(endpoint, null);
+  async generateInvitationCode(userId: string): Promise<InvitationCode> {
+    const endpoint = API_ENDPOINTS.userInvitationCode(userId);
+    return apiClient.post<InvitationCode>(endpoint, null);
   },
 };
--- a/frontend/src/lib/api/types.ts
+++ b/frontend/src/lib/api/types.ts
@@ -94,9 +94,12 @@ export interface AuthResponse {
   developer_id: string;
 }
 
-export interface Token {
-  access_token: string;
-  token_type: string;
+export interface InvitationCode {
+  id: string;
+  code: string;
+  user_id: string;
+  expires_at: string;
+  created_at: string;
 }
 
 export interface RegisterRequest {
--- a/frontend/src/routes/_authenticated/users/$userId.tsx
+++ b/frontend/src/routes/_authenticated/users/$userId.tsx
@@ -20,7 +20,7 @@ import {
   useUser,
   useDeleteUser,
   useAppleXmlUpload,
-  useGenerateUserToken,
+  useGenerateInvitationCode,
 } from '@/hooks/api/use-users';
 import { ROUTES } from '@/lib/constants/routes';
 import { copyToClipboard } from '@/lib/utils/clipboard';
@@ -81,13 +81,13 @@ function UserDetailPage() {
   const { mutate: deleteUser, isPending: isDeleting } = useDeleteUser();
   const { handleUpload, isUploading: isUploadingFile } = useAppleXmlUpload();
   const {
-    mutate: generateToken,
-    data: tokenData,
-    isPending: isGeneratingToken,
-  } = useGenerateUserToken();
+    mutate: generateInvitationCode,
+    data: invitationCodeData,
+    isPending: isGeneratingCode,
+  } = useGenerateInvitationCode();
   const [copied, setCopied] = useState(false);
-  const [tokenCopied, setTokenCopied] = useState(false);
-  const [isTokenDialogOpen, setIsTokenDialogOpen] = useState(false);
+  const [codeCopied, setCodeCopied] = useState(false);
+  const [isCodeDialogOpen, setIsCodeDialogOpen] = useState(false);
   const fileInputRef = useRef<HTMLInputElement>(null);
 
   const isUploading = isUploadingFile(userId);
@@ -171,22 +171,22 @@ function UserDetailPage() {
     });
   };
 
-  const handleGenerateToken = () => {
-    generateToken(userId, {
+  const handleGenerateInvitationCode = () => {
+    generateInvitationCode(userId, {
       onSuccess: () => {
-        setIsTokenDialogOpen(true);
+        setIsCodeDialogOpen(true);
       },
     });
   };
 
-  const handleCopyToken = async () => {
+  const handleCopyCode = async () => {
     const success = await copyToClipboard(
-      tokenData?.access_token || '',
-      'Token copied to clipboard'
+      invitationCodeData?.code || '',
+      'Invitation code copied to clipboard'
     );
     if (success) {
-      setTokenCopied(true);
-      setTimeout(() => setTokenCopied(false), 2000);
+      setCodeCopied(true);
+      setTimeout(() => setCodeCopied(false), 2000);
     }
   };
   if (!userLoading && !user) {
@@ -274,10 +274,10 @@ function UserDetailPage() {
           </Button>
           <Button
             variant="secondary"
-            onClick={handleGenerateToken}
-            disabled={isGeneratingToken}
+            onClick={handleGenerateInvitationCode}
+            disabled={isGeneratingCode}
           >
-            {isGeneratingToken ? (
+            {isGeneratingCode ? (
               <>
                 <Loader2 className="h-4 w-4 animate-spin" />
                 Generating...
@@ -285,7 +285,7 @@ function UserDetailPage() {
             ) : (
               <>
                 <Key className="h-4 w-4" />
-                Generate SDK Token
+                Generate Invitation Code
               </>
             )}
           </Button>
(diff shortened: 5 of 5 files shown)


backend/app/schemas/series_types.py
```diff
@@ -111,6 +111,14 @@ class SeriesType(str, Enum):
     time_in_daylight = "time_in_daylight"
     water_temperature = "water_temperature"
 
+    # =========================================================================
+    # GARMIN-SPECIFIC METRICS (IDs 220-239)
+    # =========================================================================
+    stress_level = "stress_level"  # Garmin stress score (1-100)
```
bartmichalak: If those attributes are Garmin-specific, I think they should reflect that fact in their name.

Not sure about body battery right now, but other metrics are definitely monitored by other providers, like Oura Ring, Whoop, or Samsung Galaxy Watch.

I would assume the following:
- for provider-specific metrics, let's include that provider in the name
- for general metrics (those we'll introduce in the future at the OW level - the name will be generic - they will be provider-agnostic)

Those specific values currently shouldn't be returned in any endpoints - but we can keep them in the DB, later we'll figure out how to approach it. 

WDYT?

backend/app/api/routes/v1/garmin_webhooks.py
```diff
@@ -604,17 +622,17 @@ async def garmin_push_notification(
...
         # Sequential flow: each webhook triggers the NEXT data type
-        # sleeps → webhook → dailies → webhook → epochs → ... → next day
+        # sleeps → webhook → dailies → webhook → epochs → ... → complete
         backfill_triggered = []
         for user_id_str in all_users:
-            status = get_backfill_status(user_id_str)
-            if status["in_progress"]:
+            backfill_status = get_backfill_status(user_id_str)
+            if backfill_status["overall_status"] == "in_progress":
                 logger.info(
```
bartmichalak: Can you please replace it and next ones with `log_structured`? 

It makes browsing logs a way easier! 

You can see a reference here:
https://github.com/the-momentum/open-wearables/blob/main/backend/app/api/routes/v1/sdk_sync.py#L42

backend/app/repositories/data_point_series_repository.py
```diff
@@ -101,21 +101,38 @@ def _resolve_mappings(
...
+        from sqlalchemy import and_, or_
+
+        # Build OR conditions to handle NULL device_id properly
+        conditions = []
+        for user_id, provider_name, device_id in identities:
+            if device_id is None:
+                # Use IS NULL for NULL device_id
+                conditions.append(
+                    and_(
+                        ExternalDeviceMapping.user_id == user_id,
```
bartmichalak: This model has been removed in this PR:
https://github.com/the-momentum/open-wearables/pull/376

@KaliszS can provide more information

KaliszS (commented):
We need to resolve conflicts first.

backend/app/api/routes/v1/oauth.py
```diff
@@ -92,6 +92,10 @@ async def oauth_callback(
         providers=[provider.value],
     )
 
+    # For Garmin: Auto-trigger 90-day backfill for all 16 data types
+    if provider.value == "garmin":
```
bartmichalak: Shouldn't we use here enum value? cc @KaliszS

backend/app/schemas/series_types.py
farce1: Done — renamed all four to `garmin_stress_level`, `garmin_skin_temperature`, `garmin_fitness_age`, `garmin_body_battery` in both the enum and `SERIES_TYPE_DEFINITIONS`, plus updated all references in `data_247.py`.

backend/app/api/routes/v1/garmin_webhooks.py
farce1: Done — replaced all ~40 `logger.info/warning/error/debug` calls with `log_structured(logger, level, message, **attrs)`. Dynamic values moved from f-strings to keyword arguments.

backend/app/repositories/data_point_series_repository.py
farce1: This is already resolved — `ExternalDeviceMapping` was removed during the merge with main and no references remain in the current code.

backend/app/api/routes/v1/oauth.py
farce1: Done — changed to `if provider == ProviderName.GARMIN:`.

backend/app/api/routes/v1/garmin_webhooks.py
```diff
@@ -469,152 +601,137 @@ async def garmin_push_notification(
...
                     errors.append(f"Error processing activity {activity_id}: {str(e)}")
 
-        # Process wellness data types (sleeps, dailies, epochs, bodyComps)
+        # Process all wellness data types (batch processing)
         wellness_results: dict[str, Any] = {}
+        users_with_new_success: set[str] = set()  # Track first-time successes only
 
         if hasattr(garmin_strategy, "data_247") and garmin_strategy.data_247:
             garmin_247 = cast(Garmin247Data, garmin_strategy.data_247)
             repo = UserConnectionRepository()
```
czajkub, line 620: Will there ever be a case when this:
`if hasattr(garmin_strategy, "data_247") and garmin_strategy.data_247:`
is not fulfilled? From what i can see, `garmin_strategy.data_247` is initialised when calling
`garmin_strategy = factory.get_provider("garmin")`
Also, I would rename `repo` to `user_repo` or something similar to improve readability (since this is a pretty large file)

backend/app/integrations/celery/tasks/__init__.py
```diff
@@ -1,9 +1,30 @@
...
+from .garmin_backfill_task import (
+    reset_type_status as reset_garmin_type_status,
+)
+from .garmin_backfill_task import (
+    start_full_backfill as start_garmin_full_backfill,
+)
+from .garmin_backfill_task import (
+    trigger_backfill_for_type as trigger_garmin_backfill_for_type,
+)
+from .garmin_backfill_task import (
```
czajkub, line 26: I'd group all of these imports into one like this:
```python
from .garmin_backfill_task import (
    check_triggered_timeout as check_garmin_triggered_timeout,
    get_backfill_status as get_garmin_backfill_status,
    ....
)
```

backend/app/integrations/celery/tasks/garmin_backfill_task.py
```diff
@@ -1,238 +1,953 @@
...
+GARMIN_BACKFILL_BUDGET_PCT = 0.3  # Reserve 30% of rate limit for backfill
+
+# Rate limiting delays (derived from Garmin's 100 req/min limit)
+_backfill_budget = int(GARMIN_RATE_LIMIT_REQUESTS * GARMIN_BACKFILL_BUDGET_PCT)  # 30 req/min
+DELAY_BETWEEN_TYPES = GARMIN_RATE_LIMIT_WINDOW // _backfill_budget  # 2s between types
+DELAY_AFTER_RATE_LIMIT = GARMIN_RATE_LIMIT_WINDOW  # Wait for full window reset (60s)
+
+# Stale request timeout settings
+TRIGGERED_TIMEOUT_SECONDS = 300  # 5 min before skipping a triggered type
+MAX_TYPE_ATTEMPTS = 3  # Total attempts (1 original + 2 retries)
```
czajkub, line 46: Maybe we could move these constants into a config file? The `app/config.py` is already large as is, so maybe to some specific garmin config file?

backend/app/integrations/celery/tasks/garmin_backfill_task.py
```diff
@@ -1,238 +1,953 @@
...
+                type_info["completed_at"] = completed_at
+            success_count += 1
+        elif status == "failed":
+            error = redis_client.get(_get_key(user_id_str, "types", data_type, "error"))
+            if error:
+                type_info["error"] = error
+            failed_count += 1
+        elif status == "skipped":
+            skipped_count += 1
+        else:
```
czajkub, line 162: This is mostly preference so it's ultimately up to you, but we could change these `elif` chains to a `match` statement, which (I believe) is more used in this repo

backend/app/integrations/celery/tasks/garmin_backfill_task.py
```diff
@@ -1,238 +1,953 @@
...
-    redis_client.expire(_get_backfill_key(user_id, "days_completed"), 3600)
 
-    logger.info(f"Completed backfill for user {user_id}")
+def complete_backfill(user_id: str | UUID) -> None:
+    """Mark entire backfill as complete (all types processed)."""
+    redis_client = get_redis_client()
+    user_id_str = str(user_id)
+
+    # Set overall complete marker with shorter TTL (1 day)
+    redis_client.setex(_get_key(user_id_str, "overall_complete"), 86400, "1")
```
czajkub, line 455: This is really nitpicky, but maybe we could change the `86400` here to `24 * 60 * 60` or move it to a separate variable? Though it's not really necessary since it's explained in the comment above. Also, the `"1"` is signifying success (or True), right?

backend/app/integrations/celery/tasks/garmin_backfill_task.py
```diff
@@ -1,238 +1,953 @@
...
+    for data_type in ALL_DATA_TYPES:
+        reset_type_status(user_id, data_type)
 
-    Sequence:
-    1. sleeps for day 1 → webhook → dailies for day 1 → ...
-    2. After all 5 data types → move to day 2
-    3. After 30 days → complete
+    # Trigger the first type (with small delay to avoid immediate burst)
+    first_type = ALL_DATA_TYPES[0]
+    trigger_backfill_for_type.apply_async(args=[user_id, first_type], countdown=1)
```
czajkub, line 511: What changes here with a 1 second delay? Is this because of some weirdness from Garmin's side?

backend/app/integrations/celery/tasks/garmin_backfill_task.py
```diff
@@ -1,238 +1,953 @@
...
+        Dict with trigger status
+    """
+    if data_type not in ALL_DATA_TYPES:
+        return {"error": f"Invalid data type: {data_type}"}
 
     try:
         user_uuid = UUID(user_id)
     except ValueError as e:
-        logger.error(f"[trigger_next_backfill] Invalid user_id: {user_id}")
+        logger.error(f"Invalid user_id: {user_id}")
```
czajkub, line 636: Is there a reason as to why this was kept as a logger error as opposed to `log_structured`?

backend/app/repositories/data_point_series_repository.py
czajkub, line 1: BTW, there is another pull request open from another contributor (#390) that addresses the same issue. So we should decide which approach we should take (although they are mostly similar, the difference being that here we are upserting and the other PR still has `on conflict do nothing`

backend/app/services/providers/garmin/handlers/backfill.py
```diff
@@ -25,20 +26,31 @@ class GarminBackfillService:
...
+    MAX_HEALTH_API_DAYS = 30
+    MAX_ACTIVITY_API_DAYS = 30
+
+    # Constants for compatibility and clarity
+    BACKFILL_CHUNK_DAYS = 30  # Max days per single request (Garmin limit)
+    MAX_REQUEST_DAYS = 30  # Alias for BACKFILL_CHUNK_DAYS
+    SUMMARY_DAYS = 0  # No summary coverage gap (REST endpoints removed)
+
+    # Data types that use the Activity API (same 30-day limit as all types)
+    ACTIVITY_API_TYPES = {"activities", "activityDetails"}
```
czajkub, line 53: I'd also move these to some garmin-specifig config like in one of the comments above

backend/app/services/providers/garmin/data_247.py
```diff
@@ -675,10 +687,825 @@ def save_hrv_data(
...
+        )
+
+        return record, detail
+
+    def save_activity_data(
+        self,
+        db: DbSession,
+        user_id: UUID,
+        raw_activity: dict[str, Any],
+    ) -> int:
```
czajkub, line 775: Why are we returning `int` here as opposed to `bool`?

backend/app/services/providers/garmin/data_247.py
```diff
@@ -675,10 +687,825 @@ def save_hrv_data(
...
+            source=self.provider_name,
+            user_id=user_id,
+        )
+
+    def save_moveiq_data(
+        self,
+        db: DbSession,
+        user_id: UUID,
+        raw_moveiq: dict[str, Any],
+    ) -> int:
```
czajkub, line 1353: same here with int/bool

backend/app/services/providers/garmin/workouts.py
```diff
@@ -81,9 +84,7 @@ def get_workouts_historical(
                     all_activities.extend(activities)
             except Exception as e:
                 # Log error but continue with other chunks
-                import logging
-
-                logging.getLogger(__name__).warning(
+                logger.warning(
```
czajkub, line 87: also here logger / `log_structured`

backend/app/services/providers/api_client.py
czajkub, line 1: also in this file some logger / `log_structured` discrepancies

PR #426, opened 2026-02-03
chore: add code review skill

bartmichalak:
## Description

Adds a code review skill (for now, only for backend) that validates generated code (likely by AI) against project conventions, specifically checking for Pydantic v2 syntax compliance and proper layer separation between repositories and services.

### General

NA

### Backend Changes

NA

### Frontend Changes

NA

## Testing Instructions

NA

## Screenshots

NA

## Additional Notes

NA


.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+| `@validator` | `@field_validator` |
+| `@root_validator` | `@model_validator` |
+| `schema_extra` | `json_schema_extra` |
+| `.dict()` | `.model_dump()` |
+| `.json()` | `.model_dump_json()` |
+| `.parse_obj()` | `.model_validate()` |
+| `.parse_raw()` | `.model_validate_json()` |
+| `Field(regex=...)` | `Field(pattern=...)` |
+| `Optional[X] = None` | `X \| None = None` |
+| `List[X]`, `Dict[K, V]` | `list[X]`, `dict[K, V]` |
```
KaliszS: Those two are not Pydantic related, just Python 3.10+ syntax. I mena it's correct, but wrong category. We can create `modenr python` section with those two + generics syntax from python 3.12+

.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+
+    @field_validator("email")
+    @classmethod
+    def validate_email(cls, v: str | None) -> str | None:
+        if v and "@" not in v:
+            raise ValueError("Invalid email")
+        return v
+```
+
+**Incorrect example (DO NOT USE):**
```
KaliszS: Why do we even need incorect examples? I would say it just enhance tokens usage in context, so it could make agent job worse. Also informations in that md file are pretty basics, I don't think they require "not to do" sort of examples.

.ai/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+        users = db.query(User).filter(User.is_premium == True).all()
+        for user in users:
+            user.discount = self.calculate_discount(user)  # BAD: calculation
+        return users
+```
+
+#### Services (`app/services/`)
+- Contains **ALL** business logic
+- **NEVER** performs database operations directly (use repositories)
+- Coordinates between repositories, external APIs, and other services
```
KaliszS, line 89: mainly between repositories and endpoints routes, the latter is missing here

.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+        for user in users:
+            user.discount = self.calculate_discount(user)  # BAD: calculation
+        return users
+```
+
+#### Services (`app/services/`)
+- Contains **ALL** business logic
+- **NEVER** performs database operations directly (use repositories)
+- Coordinates between repositories, external APIs, and other services
+- Handles validation beyond Pydantic schema validation
```
KaliszS: validtaion and error handling

.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,194 @@
...
+class UserService:
+    def get_user(self, db: DbSession, user_id: UUID) -> User:
+        return db.query(User).filter(User.id == user_id).first()  # BAD: direct query
+
+    def update_user(self, db: DbSession, user_id: UUID, data: dict):
+        db.execute(update(User).where(User.id == user_id).values(**data))  # BAD
+        db.commit()
+```
+
+### Type Hints
```
KaliszS: it can be part of the `modern python` section

.claude/skills/code-review/SKILL.md
bartmichalak: Agreed and removed

If you can think of any more tricky or sophisticated examples of when to use / don't use something, feel free to commit your suggestions directly to this branch

.ai/skills/code-review/SKILL.md
bartmichalak: But it's written from the service perspective - services don't coordinate endpoint routes - it's the opposite dependency. Endpoints call services, not the other way around.                                               
                                                                                                                                                                                                                                                But I'm also curious about your approach when it comes to calling services from another service- do we allow such a behaviour?

.claude/skills/code-review/SKILL.md
bartmichalak: moved 👍

.claude/skills/code-review/SKILL.md
bartmichalak: As you suggested, I've extracted it to a separate section

Please take a look and let me know if anything is still missing

.ai/skills/code-review/SKILL.md
KaliszS: Yes, we do. For example healthkit/import_service is calling event_record_service. It's used as a composition inside init method.

.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+        return user
+
+    def get_premium_users_with_discount(self, db: DbSession) -> list[UserWithDiscount]:
+        users = self.crud.get_premium(db)  # Repository handles query
+        return [self._apply_discount(u) for u in users]  # Logic in service
+```
+
+### Modern Python Syntax
+
+Use modern Python 3.10+ and 3.12+ syntax:
```
KaliszS: Table below is about only 3.10, 3.12+ syntax is about generics in next subsection.

.claude/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,178 @@
...
+| `Optional[X]` | `X \| None` |
+| `Union[X, Y]` | `X \| Y` |
+| `List[X]` | `list[X]` |
+| `Dict[K, V]` | `dict[K, V]` |
+| `Tuple[X, ...]` | `tuple[X, ...]` |
+| `Set[X]` | `set[X]` |
+
+**Generics (Python 3.12+):**
+```python
+# Deprecated - using TypeVar
```
KaliszS: It's not really deprecated. In 90% (or more) use cases, new syntax for generic can be used instead of TypeVar, but in reality that new syntax is creating TypeAliases, that's not exactly the same thing.

So I would just change that text from `Deprecated - using TypeVar` to `redundant in most cases that accepts type aliases - using TypeVar`

Issue #288, opened 2026-01-12
HIPAA compliance support and PHI data handling


bartmichalak:
As healthcare organizations evaluate Open Wearables for production use, we need to address compliance requirements and data protection standards for handling Protected Health Information (PHI).

# Areas to address:

## Access Tracking & Accountability
- Implementation of comprehensive audit logging
- Tracking data access patterns (user, timestamp, authentication method)
- Retention policies for audit records

## Current Logging Practices Review
- Assessment of existing error tracking and monitoring systems
- Identification of potential PHI exposure in logs/telemetry
- Review of what data currently gets captured in error reports

## Secure Configuration Guidelines
- Best practices documentation for healthcare deployments
- Data sanitization/redaction strategies
- Recommendations for production logging levels
- Guidance on disabling verbose logging of sensitive payloads

Additional Considerations:
- Data anonymization/de-identification options
- Business Associate Agreement (BAA) requirements for infrastructure providers

PR #431, opened 2026-02-05
feat(backend): add refresh token support for SDK and developer authentication

bartmichalak:
## Description

Implementation of refresh tokens mechanism for two authentication types:
  - SDK tokens (mobile apps)
  - Developer tokens (admin dashboard)

Key changes
  - Token rotation - each refresh invalidates the old token and issues a new one
  - Opaque tokens with rt-{32 hex chars} format
  - Support for multiple active tokens per user (e.g., different devices)
  - Revoked tokens retained in database for audit purposes (we probably don't need it for now, but might be helpful fot the future [support of HIPAA](https://github.com/the-momentum/open-wearables/issues/288)

**New endpoints:**
- POST /v1/token/refresh                                                                                                                                                                                                                      
- POST /v1/token/revoke  

**Modified endpoints:**
- POST /v1/auth/login                                                                                                                                                                                                                           
  - Added refresh_token in response                                                                                                                                                                                                             
                                                                                                                                                                                                                                                
- POST /v1/sdk/users/{user_id}/token                                                                                                                                                                                                            
  - Added refresh_token in response                                                                                                                                                                                                             
  - Removed infinite parameter from request                                                                                                                                                                                                     
  - Tokens now always expire (previously admin-generated tokens had no expiration)

## Checklist

### General

### Backend Changes

## Testing Instructions

  1. SDK Token Flow

  **Generate SDK token with refresh token**
  POST /api/v1/sdk-token
  {
    "user_id": "<test_user_id>",
    "app_id": "<test_app_id>"
  }
  Expected: Response includes both access_token and refresh_token (rt-...)

  **Refresh the token**
  POST /api/v1/token/refresh
  {
    "refresh_token": "rt-<token_from_step_1>"
  }
  Expected:
  - New access_token returned
  - New refresh_token returned (different from input)
  - Old refresh token is now invalid

  **Verify old token is revoked**
  POST /api/v1/token/refresh
  {
    "refresh_token": "rt-<old_token_from_step_1>"
  }
  Expected: 401 Unauthorized (token revoked)

  ---
  2. Developer Token Flow

  **Login as developer**
  POST /api/v1/auth/login
  {
    "email": "<developer_email>",
    "password": "<password>"
  }
  Expected: Response includes access_token and refresh_token

  **Refresh developer token**
  POST /api/v1/token/refresh
  {
    "refresh_token": "rt-<developer_refresh_token>"
  }
  Expected: New access_token and refresh_token

  ---
  3. Token Revocation

  **Revoke a refresh token**
  POST /api/v1/token/revoke
  {
    "refresh_token": "rt-<valid_refresh_token>"
  }
  Expected: 200 OK

  **Verify revoked token cannot be used**
  POST /api/v1/token/refresh
  {
    "refresh_token": "rt-<revoked_token>"
  }
  Expected: 401 Unauthorized

  ---
  4. Edge Cases

  - Invalid token format (no rt- prefix) → 400 Bad Request
  - Non-existent token → 401 Unauthorized
  - Empty refresh_token → 422 Validation Error
  - Use access_token as refresh_token → 401 Unauthorized
  - Multiple devices (multiple valid refresh tokens) → All should work independently

  ---
  5. Automated Tests
  pytest backend/tests/api/v1/test_token.py -v
  pytest backend/tests/services/test_refresh_token_service.py -v
  pytest backend/tests/repositories/test_refresh_token_repository.py -v



## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/backend/app/services/refresh_token_service.py
+++ b/backend/app/services/refresh_token_service.py
@@ -5,6 +5,7 @@ from uuid import UUID
 
 from fastapi import HTTPException, status
 
+from app.config import settings
 from app.database import DbSession
 from app.models import RefreshToken
 from app.repositories.refresh_token_repository import refresh_token_repository
@@ -133,6 +134,7 @@ class RefreshTokenService:
             access_token=access_token,
             token_type="bearer",
             refresh_token=new_refresh_token,
+            expires_in=settings.access_token_expire_minutes * 60,
         )
 
     def revoke_token(self, db_session: DbSession, refresh_token_str: str) -> bool:
--- a/backend/app/api/routes/v1/auth.py
+++ b/backend/app/api/routes/v1/auth.py
@@ -3,6 +3,7 @@ from typing import Annotated
 from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi.security import OAuth2PasswordRequestForm
 
+from app.config import settings
 from app.database import DbSession
 from app.schemas import DeveloperRead, DeveloperUpdate
 from app.schemas.token import TokenResponse
@@ -44,7 +45,12 @@ async def login(
     access_token = create_access_token(subject=str(developer.id))
     refresh_token = refresh_token_service.create_developer_refresh_token(db, developer.id)
 
-    return TokenResponse(access_token=access_token, token_type="bearer", refresh_token=refresh_token)
+    return TokenResponse(
+        access_token=access_token,
+        token_type="bearer",
+        refresh_token=refresh_token,
+        expires_in=settings.access_token_expire_minutes * 60,
+    )
 
 
 @router.post("/logout")
--- a/backend/app/api/routes/v1/sdk_token.py
+++ b/backend/app/api/routes/v1/sdk_token.py
@@ -3,6 +3,7 @@ from uuid import UUID
 
 from fastapi import APIRouter, Body, HTTPException, status
 
+from app.config import settings
 from app.database import DbSession
 from app.schemas.sdk import SDKTokenRequest
 from app.schemas.token import TokenResponse
@@ -70,4 +71,9 @@ async def create_user_token(
 
     refresh_token = refresh_token_service.create_sdk_refresh_token(db, user_id, app_id)
 
-    return TokenResponse(access_token=access_token, token_type="bearer", refresh_token=refresh_token)
+    return TokenResponse(
+        access_token=access_token,
+        token_type="bearer",
+        refresh_token=refresh_token,
+        expires_in=settings.access_token_expire_minutes * 60,
+    )
--- a/backend/tests/api/v1/test_auth.py
+++ b/backend/tests/api/v1/test_auth.py
@@ -14,6 +14,7 @@ Tests cover:
 from fastapi.testclient import TestClient
 from sqlalchemy.orm import Session
 
+from app.config import settings
 from tests.factories import DeveloperFactory
 from tests.utils import developer_auth_headers
 
@@ -37,6 +38,7 @@ class TestLogin:
         data = response.json()
         assert "access_token" in data
         assert data["token_type"] == "bearer"
+        assert data["expires_in"] == settings.access_token_expire_minutes * 60
         assert isinstance(data["access_token"], str)
         assert len(data["access_token"]) > 0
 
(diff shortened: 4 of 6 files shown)


backend/app/models/refresh_token.py
```diff
@@ -0,0 +1,28 @@
...
+    """Generic refresh token for SDK and Developer tokens.
+
+    Stores opaque refresh tokens in the database for secure token refresh.
+    The token_type field indicates whether this is an SDK token or Developer token.
+    """
+
+    __tablename__ = "refresh_token"
+
+    id: Mapped[PrimaryKey[str_64]]  # rt-{32 hex chars}
+    token_type: Mapped[str_64]  # "sdk" | "developer"
```
KaliszS: If there are only two values possible, maybe Enum or Literal at least? After the potential change, it has to be checked, how alembic reacts to this with generating migration.

backend/app/models/refresh_token.py
```diff
@@ -0,0 +1,28 @@
...
+
+    # For SDK tokens
+    user_id: Mapped[Indexed[FKUser] | None]
+    app_id: Mapped[str_64 | None]
```
KaliszS, line 22: isn't that related to application table? If so, it might require foreign key.

backend/app/schemas/token.py
```diff
@@ -0,0 +1,16 @@
+from pydantic import BaseModel
+
+
+class TokenResponse(BaseModel):
+    """Token response with optional refresh token."""
+
+    access_token: str
+    token_type: str = "bearer"
```
KaliszS, line 8: it was intended to be `str | None = "bearer"` probably

backend/app/services/refresh_token_service.py
```diff
@@ -0,0 +1,162 @@
...
+
+from app.database import DbSession
+from app.models import RefreshToken
+from app.repositories.refresh_token_repository import refresh_token_repository
+from app.schemas.token import TokenResponse
+from app.services.sdk_token_service import create_sdk_user_token
+from app.utils.security import create_access_token
+
+
+def _generate_refresh_token_id() -> str:
```
KaliszS: Why not to add this as `@staticmethod` inside service?

backend/app/services/sdk_token_service.py
```diff
@@ -5,31 +5,28 @@
 from app.config import settings
 
 
-def create_sdk_user_token(app_id: str, user_id: str, infinite: bool = False) -> str:
+def create_sdk_user_token(app_id: str, user_id: str) -> str:
```
KaliszS, line 8: Do we really need it to be separated service from refresh_token_service?

backend/app/models/refresh_token.py
bartmichalak: Yeah, that's a very good idea. I observed, that LLMs tend to not using enums in such cases and I agree with you it's a bad decision. Maybe it's worth adding it to `AGENTS.md`, we don't anything about enums there.

I added TokenType(StrEnum). Alembic doesn't generate a new migration since the DB column stays String(64).

I only needed to add the mapping to type_annotation_map in `database.py`

backend/app/models/refresh_token.py
bartmichalak: You're right that this looks like a missing FK. I even started implementing this change, however then I realised that `app_id` in RefreshToken serves a dual role - for app-generated tokens it stores `Application.app_id`, but for admin-generated tokens it stores a synthetic string `"admin:{developer_id}"` which doesn't exist in the application table. 

Adding a FK would break the admin flow...so I created the issue to address this problem:
https://github.com/the-momentum/open-wearables/issues/443

As a separate PR, because I think it could require from us a little bit refactor here.

backend/app/services/refresh_token_service.py
bartmichalak: Yeah, it makes sense, moved.

backend/app/services/sdk_token_service.py
bartmichalak: You're right, it doesn't need to be a separate service. 

However, rather than merging it into `RefreshTokenService` I'd approach it differently - we could extend the existing `create_access_token` in utils/security.py to support extra claims via **kwargs. 

That way all stateless JWT generation lives in one place and `create_sdk_user_token` becomes just `create_access_token(subject=user_id, scope="sdk", app_id=app_id)`

I think it can be addressed separately, as it won't introduce any breaking changes in the contract between SDK and backend.  

Created #444 to track this.

PR #446, opened 2026-02-08
feat: add Strava provider integration

piotrsedzik:
## Summary

Adds full Strava provider support to Open Wearables, enabling users to connect their Strava account and sync all workout activities. This follows the existing provider pattern (Strategy + Factory + OAuth/Workouts templates) used by Garmin, Polar, Suunto, and Whoop.

### What is included

**Strava Provider (9 new files):**
- **OAuth 2.0 flow** - Authorize, token exchange, and athlete profile retrieval (no PKCE, BODY auth method)
- **Activity sync** - Fetches activities from Strava API with page-based pagination (up to 200/page), maps 30+ Strava sport types (Run, TrailRun, Ride, MountainBikeRide, Swim, Yoga, etc.) to the unified WorkoutType enum
- **Metrics extraction** - Heart rate (avg/max), distance, speed, power (watts), elevation gain, calories, moving time
- **Historical backfill** - Configurable date range, supports syncing years of data in a single request
- **Webhook endpoints** - GET verification (echoes hub.challenge) and POST event handler for real-time activity sync; always returns 200 to prevent Strava retries
- **Provider icon** - Strava SVG logo

**Bug fixes (4 changes across backend + frontend):**
- **Moving time over elapsed time** - Workout duration now uses moving_time (excludes pauses) instead of elapsed_time
- **Calorie calculation** - Fixed energy conversion from kJ x 0.239 (pure physics) to kcal = kJ (standard exercise science approximation where human metabolic efficiency ~25% cancels the unit conversion factor)
- **Raised pagination limits** - Workouts endpoint limit increased from 100 to 1000 for larger date ranges (365d)
- **Activity tab sort order** - Added sort_order query parameter (asc/desc) to the activity summaries endpoint; frontend now requests desc so the Activity Days list shows today first instead of the oldest record

### Files changed

| Type | File | Purpose |
|------|------|---------|
| Modified | schemas/oauth.py | Add STRAVA to ProviderName enum |
| Modified | config.py | Add Strava OAuth settings (client ID/secret, redirect URI, scopes, webhook verify token) |
| Modified | config/.env.example | Add Strava environment variables |
| Modified | schemas/__init__.py | Export StravaActivityJSON |
| Modified | providers/factory.py | Register StravaStrategy in provider factory |
| Modified | routes/v1/__init__.py | Mount Strava webhook router |
| Modified | routes/v1/events.py | Raise workout limit from 100 to 1000 |
| Modified | routes/v1/summaries.py | Add sort_order param to activity summaries |
| Modified | summaries_service.py | Implement descending sort with correct cursor pagination |
| Modified | activity-section.tsx | Use sort_order desc for days list, remove .reverse() hack |
| Modified | workout-section.tsx | Raise summary fetch limit to 1000 |
| Modified | health.service.ts | Add sort_order to SummaryParams interface |
| **New** | strava_webhooks.py | Webhook verification + event handler endpoints |
| **New** | workout_types/strava.py | Map 30+ Strava sport types to unified WorkoutType |
| **New** | schemas/strava/ | Pydantic model for Strava activity JSON |
| **New** | providers/strava/oauth.py | OAuth 2.0 implementation |
| **New** | providers/strava/workouts.py | Activity fetch, normalization, backfill, webhook processing |
| **New** | providers/strava/strategy.py | Provider strategy (no 24/7 data - Strava only has workouts) |
| **New** | provider-icons/strava.svg | Provider icon |

## Test plan

- [x] OAuth flow: redirects to Strava login, callback exchanges token, athlete profile saved
- [x] Activity sync: 2,430 historical activities fetched and normalized (2018-2026)
- [x] Calorie values verified against Strava API raw data (kilojoules to kcal)
- [x] Moving time correctly used for duration (excludes pauses)
- [x] Activity tab displays most recent days first with correct pagination
- [x] Workout tab shows up to 1000 activities for 365-day range
- [ ] Webhook endpoint: verify with POST /api/v1/strava/webhooks/webhook (requires ngrok for local dev)
- [ ] Run make test to verify no regressions

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/frontend/src/components/user/workout-section.tsx
+++ b/frontend/src/components/user/workout-section.tsx
@@ -336,7 +336,7 @@ export function WorkoutSection({
     {
       start_date: dateToTimestamp(startDate),
       end_date: dateToTimestamp(endDate),
-      limit: 1000,
+      limit: 100,
       sort_order: 'desc',
     }
   );
--- a/frontend/src/lib/api/services/health.service.ts
+++ b/frontend/src/lib/api/services/health.service.ts
@@ -41,7 +41,7 @@ export interface SummaryParams {
   start_date: string; // ISO date string (e.g., "2025-01-01T00:00:00Z")
   end_date: string; // ISO date string
   cursor?: string;
-  limit?: number; // 1-400, default 50
+  limit?: number; // 1-100, default 50
   sort_order?: 'asc' | 'desc';
   [key: string]: string | number | undefined;
 }


backend/app/constants/workout_types/strava.py
```diff
@@ -0,0 +1,106 @@
+from app.schemas.workout_types import WorkoutType
+
+# Strava API sport_type values mapped to unified WorkoutType.
+# Strava uses PascalCase strings for sport_type (preferred over the legacy `type` field).
+# Source: https://developers.strava.com/docs/reference/#api-models-SportType
```
bartmichalak, line 5: In the list below, I see values ​​such as:
- football
- baseball
- basketball
- volleyball
- hockey

and a few more that I can't find in this documentation:
https://developers.strava.com/docs/reference/#api-models-SportType

What is the reason? Where do the values ​​in this list come from?

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+from app.repositories import UserConnectionRepository
+from app.schemas import StravaActivityJSON
+from app.services.providers.factory import ProviderFactory
+from app.services.providers.strava.workouts import StravaWorkouts
+
+router = APIRouter()
+logger = getLogger(__name__)
+
+
+@router.get("/webhook")
```
bartmichalak, line 18: I see that the LLM probably followed what we have in the Garmin webhook implementation here, but it's a bad pattern. The logic should be moved to the separate service (please take a look at the rest of the endpoints). 

For the Garmin implementation, I'll create a separate ticket to refactor this

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+    {
+        "object_type": "activity",
+        "object_id": 12345678,
+        "aspect_type": "create",
+        "owner_id": 87654321,
+        "subscription_id": 999,
+        "event_time": 1234567890
+    }
+
+    Must always return 200 to prevent Strava from retrying.
```
bartmichalak, line 52: Does this always return 200 despite errors on our end? 
Doesn't that mean if our backend throws an error and we fail to save the data, we'll never receive it again since Strava won't retry?

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+        strava_workouts: StravaWorkouts = strava_strategy.workouts
+
+        try:
+            # Fetch full activity detail from Strava API
+            activity_data = strava_workouts.get_workout_detail_from_api(
+                db, internal_user_id, str(object_id)
+            )
+
+            if not activity_data:
+                logger.warning(f"No data returned for Strava activity {object_id}")
```
bartmichalak: Please use `log_structured` method. 

You can find example here:
https://github.com/the-momentum/open-wearables/blob/main/backend/app/integrations/celery/tasks/process_apple_upload_task.py#L47

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+        # Use sport_type for more specific mapping, fallback to type
+        workout_type = get_unified_workout_type(raw_workout.sport_type)
+
+        # Prefer moving_time (excludes pauses) over elapsed_time
+        duration_seconds = raw_workout.moving_time or raw_workout.elapsed_time
+        start_date, end_date = self._extract_dates(raw_workout.start_date, duration_seconds)
+
+        metrics = self._build_metrics(raw_workout)
+
+        device_name = raw_workout.device_name or "Strava"
```
bartmichalak: I don't think we should use Strava as a device name - it doesn't make sense to me.

backend/app/api/routes/v1/events.py
```diff
@@ -26,7 +26,7 @@ async def list_workouts(
     _api_key: ApiKeyDep,
     record_type: str | None = None,
     cursor: str | None = None,
-    limit: Annotated[int, Query(ge=1, le=100)] = 50,
+    limit: Annotated[int, Query(ge=1, le=1000)] = 50,
```
bartmichalak: What is the reason for this change? Why do we need to increase that value?

backend/app/services/providers/strava/oauth.py
```diff
@@ -0,0 +1,53 @@
...
+
+class StravaOAuth(BaseOAuthTemplate):
+    """Strava OAuth 2.0 implementation."""
+
+    @property
+    def endpoints(self) -> ProviderEndpoints:
+        """OAuth endpoints for authorization and token exchange."""
+        return ProviderEndpoints(
+            authorize_url="https://www.strava.com/oauth/authorize",
+            token_url="https://www.strava.com/api/v3/oauth/token",
```
bartmichalak: Why don't we use `api_base_url` method defined in strategy here?

@property
    def api_base_url(self) -> str:
        """Base URL for the provider's API."""
        return "https://www.strava.com"

backend/app/services/providers/strava/oauth.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # OAuth configuration
+    use_pkce: bool = False  # Strava doesn't require PKCE
+    auth_method: AuthenticationMethod = AuthenticationMethod.BODY  # Strava expects credentials in body
+
+    def _get_provider_user_info(self, token_response: OAuthTokenResponse, user_id: str) -> dict[str, str | None]:
+        """Fetches Strava athlete ID and username via API."""
+        try:
+            response = httpx.get(
+                "https://www.strava.com/api/v3/athlete",
```
bartmichalak: Same question as above

backend/app/services/providers/strava/strategy.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    def name(self) -> str:
+        """Unique identifier for the provider (lowercase)."""
+        return "strava"
+
+    @property
+    def api_base_url(self) -> str:
+        """Base URL for the provider's API."""
+        return "https://www.strava.com"
```
bartmichalak: Shouldn't we set https://www.strava.com/api/v3 here?

backend/app/schemas/strava/activity_import.py
```diff
@@ -0,0 +1,60 @@
...
+    Based on Strava API v3 DetailedActivity schema.
+    """
+
+    model_config = ConfigDict(populate_by_name=True)
+
+    # Required fields
+    id: int
+    name: str
+    type: str  # e.g. "Run", "Ride", "Swim"
+    sport_type: str  # More specific, e.g. "TrailRun", "MountainBikeRide"
```
bartmichalak, line 28: Shouldn't we consider extracting the known Strava sport types ("Run", "TrailRun", "Ride", etc.) into a StravaSportType enum? Right now these values are magic strings duplicated between this schema and the mapping in constants/workout_types/strava.py.

The field itself should stay str to gracefully accept unknown types from the API, but an enum would give us a single source of truth for the known ones

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def _normalize_workout(
+        self,
+        raw_workout: StravaActivityJSON,
+        user_id: UUID,
+    ) -> tuple[EventRecordCreate, EventRecordDetailCreate]:
+        """Normalize Strava activity to EventRecordCreate and EventRecordDetailCreate."""
+        workout_id = uuid4()
+
+        # Use sport_type for more specific mapping, fallback to type
```
bartmichalak, line 187: Where is this fallback implemented? 🤔

backend/app/api/routes/v1/events.py
FBegiello: +1, 1000 seems to be a high number - if this is needed for MCP it might be better to implement querying multiple times to cover fuller histrory access

backend/app/config.py
```diff
@@ -94,6 +94,13 @@ class Settings(BaseSettings):
     whoop_redirect_uri: str = "http://localhost:8000/api/v1/oauth/whoop/callback"
     whoop_default_scope: str = "offline read:cycles read:sleep read:recovery read:workout"
 
+    # STRAVA OAUTH SETTINGS
+    strava_client_id: str | None = None
+    strava_client_secret: SecretStr | None = None
+    strava_redirect_uri: str = "http://localhost:8000/api/v1/oauth/strava/callback"
+    strava_default_scope: str = "activity:read_all,profile:read_all"
+    strava_webhook_verify_token: str = "open-wearables-strava-verify"
```
FBegiello, line 102: If this is a token provided by strava I'd change it for `SecretStr` as to avoid sharing your personal tokens

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+
+        object_type = payload.get("object_type")
+        object_id = payload.get("object_id")
+        aspect_type = payload.get("aspect_type")
+        owner_id = payload.get("owner_id")  # Strava athlete ID
+
+        # Only process activity events
+        if object_type != "activity":
+            logger.info(f"Ignoring non-activity event: {object_type}")
+            return {"status": "ok", "message": f"Ignoring {object_type} event"}
```
FBegiello: This might require extension in the future, if non-activity features cover anything that we would want to record i.e. activities being limited only to workours and not registering HR or any such geanular data

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+            return {"status": "ok", "message": "Missing required fields"}
+
+        # Map Strava athlete ID to internal user
+        repo = UserConnectionRepository()
+        connection = repo.get_by_provider_user_id(db, "strava", str(owner_id))
+
+        if not connection:
+            logger.warning(f"No connection found for Strava athlete {owner_id}")
+            return {"status": "ok", "message": "User not connected"}
+
```
FBegiello: I'd raise an error here, returning status "ok" might be misleading

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+        internal_user_id: UUID = connection.user_id
+        logger.info(f"Mapped Strava athlete {owner_id} to internal user {internal_user_id}")
+
+        # Get Strava workouts service via factory
+        factory = ProviderFactory()
+        strava_strategy = factory.get_provider("strava")
+        if not isinstance(strava_strategy.workouts, StravaWorkouts):
+            logger.error("Strava workouts service not available")
+            return {"status": "ok", "message": "Service unavailable"}
```
FBegiello: ^ditto

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+            return {
+                "status": "ok",
+                "activity_id": object_id,
+                "record_ids": [str(rid) for rid in created_ids],
+            }
+
+        except IntegrityError:
+            db.rollback()
+            logger.info(f"Strava activity {object_id} already exists, skipping")
+            return {"status": "ok", "message": "Duplicate activity"}
```
FBegiello: ^as with errors

backend/app/api/routes/v1/strava_webhooks.py
```diff
@@ -0,0 +1,158 @@
...
+            return {"status": "ok", "message": "Validation error"}
+
+        except Exception as e:
+            logger.error(f"Error processing Strava activity {object_id}: {e}")
+            return {"status": "ok", "message": "Processing error"}
+
+    except Exception as e:
+        logger.error(f"Error processing Strava webhook: {e}")
+        # Always return 200 to prevent Strava retries
+        return {"status": "ok", "message": "Error processing event"}
```
FBegiello: ^ditto

backend/app/constants/workout_types/strava.py
```diff
@@ -0,0 +1,106 @@
...
+        >>> get_unified_workout_type("Run")
+        WorkoutType.RUNNING
+        >>> get_unified_workout_type("TrailRun")
+        WorkoutType.TRAIL_RUNNING
+        >>> get_unified_workout_type("MountainBikeRide")
+        WorkoutType.MOUNTAIN_BIKING
+        >>> get_unified_workout_type("UnknownSport")
+        WorkoutType.OTHER
+    """
+    return STRAVA_TO_UNIFIED.get(strava_sport_type, WorkoutType.OTHER)
```
FBegiello: I'd roll this into one function, not to leave the static mapping like that

backend/app/schemas/strava/activity_import.py
FBegiello: This is in line with how we import other providers (for example garmin) - this schema I assume is representative of the incoming json, that will only later be mapped into enums on our end

backend/app/services/providers/strava/oauth.py
```diff
@@ -0,0 +1,53 @@
...
+            authorize_url="https://www.strava.com/oauth/authorize",
+            token_url="https://www.strava.com/api/v3/oauth/token",
+        )
+
+    @property
+    def credentials(self) -> ProviderCredentials:
+        """OAuth credentials from environment variables."""
+        return ProviderCredentials(
+            client_id=settings.strava_client_id or "",
+            client_secret=(settings.strava_client_secret.get_secret_value() if settings.strava_client_secret else ""),
```
FBegiello, line 29: Do we need to handle this like that? Is `strava_client_secret` optional?

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+        end_date: datetime,
+    ) -> list[Any]:
+        """Get activities from Strava API with page-based pagination.
+
+        Strava API uses epoch timestamps for after/before parameters
+        and supports up to 200 activities per page.
+        """
+        all_activities: list[Any] = []
+        page = 1
+        per_page = 200  # Strava API max
```
FBegiello: I'd considetr having this as a default value somwhere, in case it changes later - it'll be easier to grab

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+                    self.logger.warning(f"Returning partial activity data due to error: {e}")
+                    break
+                raise
+
+        return all_activities
+
+    def get_workouts_from_api(self, db: DbSession, user_id: UUID, **kwargs: Any) -> Any:
+        """Get activities from Strava API with specific options."""
+        page = kwargs.get("page", 1)
+        per_page = min(kwargs.get("per_page", 30), 200)
```
FBegiello: if this shares values with the ones before then it's antoher argument for it to be stored in a variable

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+
+        params: dict[str, Any] = {
+            "page": page,
+            "per_page": per_page,
+        }
+
+        after = kwargs.get("after")
+        before = kwargs.get("before")
+        if after:
+            params["after"] = int(after) if not isinstance(after, int) else after
```
FBegiello: we can leave only `int(after) `i think if the goal is to ensure that it always must be an int

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+        if before:
+            params["before"] = int(before) if not isinstance(before, int) else before
```
FBegiello: ^ditto

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+        end_date = start_date + timedelta(seconds=elapsed_time)
+        return start_date, end_date
+
+    def _build_metrics(self, raw_workout: StravaActivityJSON) -> EventRecordMetrics:
+        """Build metrics from Strava activity data."""
+        metrics: EventRecordMetrics = {}
+
+        # Heart rate
+        if raw_workout.average_heartrate is not None:
+            metrics["heart_rate_avg"] = Decimal(str(raw_workout.average_heartrate))
```
FBegiello: Why are we casting the value here through a str?

backend/app/services/providers/strava/workouts.py
```diff
@@ -0,0 +1,280 @@
...
+            metrics["elev_low"] = Decimal(str(raw_workout.elev_low))
+
+        # Energy: prefer calories (if available and non-zero), fallback to kilojoules.
+        # Strava's list endpoint often returns calories=None, so we fall back to kilojoules.
+        # Standard exercise approximation: kcal ≈ kJ (human efficiency ~25% cancels the unit factor).
+        if raw_workout.calories is not None and raw_workout.calories > 0:
+            metrics["energy_burned"] = Decimal(str(raw_workout.calories))
+        elif raw_workout.kilojoules is not None:
+            metrics["energy_burned"] = Decimal(str(raw_workout.kilojoules))
+
```
FBegiello, line 172: This needs to be handled in another way - we cannot store mesures in different units in one schema. We will be unable to read fromt his realiably - either recalculate it to shared unit or create a new field

backend/app/services/providers/strava/oauth.py
czajkub: no, all fields in `ProviderCredentials` are required so this way of doing it is necessary

backend/app/services/providers/strava/oauth.py
czajkub: Base url is used in the `strategy`, in oauth its these two

czajkub: Nevermind, i see what you meant

backend/app/services/providers/strava/strategy.py
czajkub: yes

backend/app/services/providers/strava/workouts.py
czajkub: I believe it's fine this way, this is the max from Strava API and we want to get the entire backfill, so this way of doing it is optimal

backend/app/services/providers/strava/workouts.py
czajkub: okay, now we can store it separately

backend/app/services/providers/strava/workouts.py
czajkub: 👍

backend/app/services/providers/strava/workouts.py
czajkub:
```python
>>> a = 3 / 7
>>> from decimal import Decimal
>>> Decimal(a)
Decimal('0.428571428571428547638078043746645562350749969482421875')
>>> Decimal(str(a))
Decimal('0.42857142857142855')
```

No idea, it loses some precision (which is the whole point of Decimal), i think the AI followed the code from somewhere else, i recall some other castring through str elswhere in the code

backend/app/services/providers/strava/workouts.py
czajkub: fallback implemented, also fixed up the workout types

backend/app/services/providers/strava/handlers/webhook_helpers.py
```diff
@@ -0,0 +1,223 @@
...
+        action="webhook_verified",
+    )
+    return {"hub.challenge": hub_challenge}
+
+
+async def handle_webhook_event(request: Request, db: DbSession) -> dict:
+    """Handle Strava webhook event."""
+    try:
+        payload = await request.json()
+        log_structured(
```
bartmichalak, line 51: It's something to address in another PR, but we should add `provider` field here and everywhere when something is related to integrations to make navigating through logs easier

backend/app/api/routes/v1/strava_webhooks.py
bartmichalak: @czajkub what about this? How does this behave?

backend/app/constants/workout_types/strava.py
```diff
@@ -0,0 +1,90 @@
...
+    "WeightTraining": WorkoutType.STRENGTH_TRAINING,
+    "Yoga": WorkoutType.YOGA,
+    "Pilates": WorkoutType.PILATES,
+    "Crossfit": WorkoutType.CARDIO_TRAINING,
+    "Elliptical": WorkoutType.ELLIPTICAL,
+    "StairStepper": WorkoutType.STAIR_CLIMBING,
+    "HighIntensityIntervalTraining": WorkoutType.CARDIO_TRAINING,
+    # Racket Sports
+    "Pickleball": WorkoutType.PICKLEBALL,
+    "Racquetball": WorkoutType.OTHER,
```
bartmichalak, line 51: I would consider adding new, missing types to our own dictionary.

bartmichalak: Added an issue:
https://github.com/the-momentum/open-wearables/issues/458

backend/app/schemas/strava/activity_import.py
bartmichalak: yeah, but shouldn't we have control over what's coming?

I mean, how will we get notified when Strava (it actually applies to other integrations too) adds a new type?

Maybe an enum isn't the best option here, but we definitely should have a consistent strategy between providers to handle that :)

@czajkub what's the current behaviour?

frontend/src/components/user/workout-section.tsx
```diff
@@ -336,7 +336,7 @@ export function WorkoutSection({
     {
       start_date: dateToTimestamp(startDate),
       end_date: dateToTimestamp(endDate),
-      limit: 100,
+      limit: 1000,
```
bartmichalak: Isn't that a leftover from the change that we finally reverted on the backend side?

frontend/src/lib/api/services/health.service.ts
```diff
@@ -41,7 +41,8 @@ export interface SummaryParams {
   start_date: string; // ISO date string (e.g., "2025-01-01T00:00:00Z")
   end_date: string; // ISO date string
   cursor?: string;
-  limit?: number; // 1-100, default 50
+  limit?: number; // 1-400, default 50
```
bartmichalak: 1-400? 🤔

frontend/src/components/user/workout-section.tsx
czajkub: yes, forgot to check the frontend side

backend/app/services/providers/strava/handlers/webhook_helpers.py
czajkub: yes, ill add an issue for that

PR #447, opened 2026-02-09
feat(backend): add missing statistics from Apple Health SDK

czajkub:
## Description

- adapt new statistics (`WorkoutStatistics`) and metrics (`SeriesType`) received from workouts / records coming from new version of Apple SDK
- refactor Apple SDK import services 
- update seeding config
- granularize `series_type` constants definitions

### General

### Backend Changes

### DIFF
--- a/backend/app/schemas/series_types.py
+++ b/backend/app/schemas/series_types.py
@@ -142,7 +142,6 @@ class SeriesType(str, Enum):
     garmin_fitness_age = "garmin_fitness_age"  # Garmin fitness age estimate
     garmin_body_battery = "garmin_body_battery"  # Garmin body battery (0-100)
 
-
     # =========================================================================
     # OTHER (IDs 500-)
     # =========================================================================


backend/app/constants/series_types/apple/workout_statistics.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+WORKOUT_STATISTIC_TYPE_TO_SERIES_TYPE: dict[str, SeriesType] = {
+    # Running Metrics
+    "averageRunningStrideLength": SeriesType.running_stride_length,
+    "averageVerticalOscillation": SeriesType.running_vertical_oscillation,
+    "averageGroundContactTime": SeriesType.running_ground_contact_time,
+    # Other Metrics
+    "averageMETs": SeriesType.physical_effort,
+    "lapLength": SeriesType.distance_other,
```
bartmichalak: are we sure here?

backend/app/constants/series_types/apple/workout_statistics.py
```diff
@@ -0,0 +1,113 @@
...
+}
+
+
+WORKOUT_STATISTIC_TYPE_TO_SERIES_TYPE: dict[str, SeriesType] = {
+    # Running Metrics
+    "averageRunningStrideLength": SeriesType.running_stride_length,
+    "averageVerticalOscillation": SeriesType.running_vertical_oscillation,
+    "averageGroundContactTime": SeriesType.running_ground_contact_time,
+    # Other Metrics
+    "averageMETs": SeriesType.physical_effort,
```
bartmichalak: This mapping doesn't look right. 

METs (Metabolic Equivalent of Task) is an objective physiological unit (1 MET = resting energy expenditure), with typical values ranging from ~1.0 to 15+. Our physical_effort is an abstract score on a 1-10 scale with unit score. 
Different semantics, different scale - storing raw MET values under this type would result in data that doesn't make sense in the context of how physical_effort is defined in our system.

I'd suggest either adding a dedicated SeriesType (e.g. average_mets with unit METs), or explicitly converting METs to our score (but that requires defining a conversion formula and deciding whether we're okay with losing the precision of the original data).

backend/app/constants/series_types/apple/category_types.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Apple HealthKit category type identifiers (HKCategoryTypeIdentifier...).
+
+    These represent categorical health data like sleep analysis.
+    """
+
+    SLEEP_ANALYSIS = "HKCategoryTypeIdentifierSleepAnalysis"
+
+    """
+    EVENTS
```
bartmichalak: I'm not sure we want to add all those categories right now since we don't really handle them in practice (needs more in-depth planning). 

I'd probably just create a few separate issues to address these topics.

bartmichalak (commented):
I haven't verified this super carefully, but general feeling:

let's not introduce new types while we don't actually support them. I think this PR should:
- add missing series types, since they (probably?) don't require any additional logic to make them work
- refactor for sleep stages

And for the rest - let's create issues with proper descriptions

backend/app/constants/series_types/apple/workout_statistics.py
```diff
@@ -0,0 +1,113 @@
...
+    "lapLength": SeriesType.distance_other,
+    "swimmingLocationType": SeriesType.distance_other,
+    "indoorWorkout": SeriesType.physical_effort,
+    # Weather
+    "weatherTemperature": SeriesType.water_temperature,
+    "weatherHumidity": SeriesType.environmental_sound_reduction,
```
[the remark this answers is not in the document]
czajkub: yea fixed

backend/app/constants/series_types/apple/workout_statistics.py
czajkub: No, in this case everything was pinned down to some series type (even if it made no sense), just added some missing series types and left missing ones unhandled with a comment

backend/app/constants/series_types/apple/workout_statistics.py
czajkub: That makes sense, added avg METs as a series type

backend/app/constants/series_types/apple/category_types.py
czajkub: Sure, I'll add an issue

backend/app/schemas/series_types.py
```diff
@@ -110,6 +128,22 @@ class SeriesType(str, Enum):
     environmental_sound_reduction = "environmental_sound_reduction"
     time_in_daylight = "time_in_daylight"
     water_temperature = "water_temperature"
+    uv_exposure = "uv_exposure"
+    inhaler_usage = "inhaler_usage"
+    weather_temperature = "weather_temperature"
+    weather_humidity = "weather_humidity"
+
+    # =========================================================================
+    # OTHER (IDs 220-)
```
czajkub: NOTE: need to adjust this, 220-239 reserved for garmin (not yet merged pr)

czajkub:
also tested apple sync process via `ngrok`-hosted instance

backend/app/constants/series_types/apple/sleep_types.py
```diff
@@ -0,0 +1,17 @@
...
+    ASLEEP_CORE = 3
+    ASLEEP_DEEP = 4
+    ASLEEP_REM = 5
+
+
+def get_apple_sleep_phase(apple_sleep_phase: int) -> SleepPhase | None:
+    try:
+        return SleepPhase(apple_sleep_phase)
+    except ValueError:
+        return None
```
KaliszS, line 17: Just to make it clear - it won't be logged to sentry.

PR #452, opened 2026-02-10
fix(backend): change "/" → "" on the developers and invitations routers

farce1:
## Description

The Team tab on Settings showed an infinite skeleton loading state in production. The root cause was FastAPI's 307 trailing-slash redirects generating `http://` URLs behind the HTTPS reverse proxy, which browsers blocked as mixed content.

**Changes:**

1. **Add trailing slashes to collection endpoints** (`config.ts`) — `/api/v1/developers/` and `/api/v1/invitations/` now match the backend canonical paths, eliminating the 307 redirect entirely
2. **Add error state to TeamTab** (`team-tab.tsx`) — follows the same pattern as CredentialsTab: shows "Failed to load team data" with a Retry button instead of infinite skeleton
3. **Remove duplicate network error retries** (`client.ts`) — the fetch-level catch block retried network errors 3 times, compounding with TanStack Query's 3 retries for 12 total attempts. Now only TanStack Query handles network retries
4. **Use `API_CONFIG.baseUrl` in provider-item** (`provider-item.tsx`) — replaced raw `import.meta.env.VITE_API_URL` with the centralized config
5. **Document trailing slash convention** (`frontend/AGENTS.md`, `backend/AGENTS.md`) — prevents future regressions

### General

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Deploy to a staging environment behind HTTPS, or locally set `VITE_API_URL` to an unreachable URL and visit Settings → Team tab
2. Verify the Team tab shows "Failed to load team data" with a Retry button (not infinite skeleton)
3. With correct config, verify Team tab loads developers and invitations without any 307 redirects in the Network tab
4. Verify Settings → Credentials, Providers, and Priorities tabs still work correctly

**Expected behavior:**
- No mixed content errors in browser console
- No 307 redirects for `/api/v1/developers/` or `/api/v1/invitations/`
- Team tab shows error state with retry on failure, data on success

## Additional Notes

The 307 redirect was the root cause: FastAPI redirects `/developers` → `/developers/`, and behind a reverse proxy the redirect URL used `http://` instead of `https://`, triggering mixed content blocks. Adding the trailing slash to the frontend endpoints avoids the redirect entirely.

### DIFF
--- a/frontend/src/lib/api/config.ts
+++ b/frontend/src/lib/api/config.ts
@@ -57,11 +57,11 @@ export const API_ENDPOINTS = {
   testAutomation: (id: string) => `/api/v1/automations/${id}/test`,
 
   // Developers endpoints
-  developers: '/api/v1/developers/',
+  developers: '/api/v1/developers',
   developerDetail: (id: string) => `/api/v1/developers/${id}`,
 
   // Invitations endpoints (authenticated)
-  invitations: '/api/v1/invitations/',
+  invitations: '/api/v1/invitations',
   invitationDetail: (id: string) => `/api/v1/invitations/${id}`,
   invitationResend: (id: string) => `/api/v1/invitations/${id}/resend`,
 
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -343,7 +343,7 @@ app/api/routes/
 - Define functions as `async` by default
 - Use **kebab-case** for paths: `/heart-rate`, `/import-data`
 - Keep route code minimal, delegate to services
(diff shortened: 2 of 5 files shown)


frontend/src/lib/api/config.ts
```diff
@@ -57,11 +57,11 @@ export const API_ENDPOINTS = {
   testAutomation: (id: string) => `/api/v1/automations/${id}/test`,
 
   // Developers endpoints
-  developers: '/api/v1/developers',
```
bartmichalak, line 60: I see that only those endpoints from the developer and invitations namespace have / at the end.

I'd rather change them there to make it consistent with the rest of the API endpoints.

This way we wouldn't even need to add those specific agent instructions.

farce1: - Backend: changed "/" → "" on the developers and invitations routers so they match every other route (no trailing slash)                                                                                      
  - Frontend: config.ts stays without trailing slashes (consistent)                                                                                                                                              
  - AGENTS docs: updated backend guidance to say "use "" not "/"", removed the frontend trailing-slash section since it's no longer needed

Issue #430, opened 2026-02-05
feat(sdk): invitation code for demo iOS / Android app onboarding


bartmichalak:
Currently demo app users need to enter `user_id` and `access_token`. With the upcoming refresh token support, 
they would need to enter 3 values: user_id, access_token, refresh_token - which is cumbersome.

## Solution
Single invitation code that can be exchanged for all auth credentials.

### Endpoints
- `POST /api/v1/users/{user_id}/invitation-code` - generate code (auth required)
- `POST /api/v1/invitation-code/redeem` - exchange code for tokens (public) 
  - **Note**: namespace TBD

### Spec
- Expires: 7 days
- Single-use
- Generator: developer via dashboard/API

Issue #443, opened 2026-02-06
RefreshToken.app_id stores mixed data - no referential integrity


bartmichalak:
\`RefreshToken.app_id\` stores two different types of values:

- \`Application.app_id\` - for tokens created by application credentials
- \`"admin:{developer_id}"\` - for tokens generated by the admin

This prevents the foreign key from being added to the \`application\` table, and the field acts as an "origin identifier" instead of a reference.

PR #455, opened 2026-02-10
feat(backend): add invitation code for smoother demo app onboarding 

bartmichalak:
## Description

Thanks to this functionality, developers will be able to generate an invitation code from the dashboard (frontend implementation in the next PR), which the user can then enter into the application during its first launch.

This replaces the need to enter `user_id` and `access_token`, which can be confusing for end users.

Furthermore, after adding the refresh token in the backend, we would have to add this as a third field, which significantly complicates onboarding.

New endpoints:
- `POST /api/v1/users/{user_id}/invitation-code` - generate code (developer auth required)
- `POST /api/v1/invitation-code/redeem` - exchange code for user_id, access_token, refresh_token (public) 

Fixes: https://github.com/the-momentum/open-wearables/issues/430 

## Checklist

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

  1. Login as developer, get Bearer token                                                                                                                                                                                                         
  2. `POST /api/v1/users/{user_id}/invitation-code` with Bearer → get `code`                                                                                                                                                                      
  3. `POST /api/v1/invitation-code/redeem` with `{"code": "..."}` → get `user_id`, `access_token`, `refresh_token`                                                                                                                                
  4. `POST /api/v1/sync/apple` with received `access_token` as Bearer → verify token works                                                                                                                                                        
  5. Redeem same code again → 404                                           
  6. Generate new code for same user → redeem old code → 404 (auto-revoked)

## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/backend/app/api/routes/v1/user_invitation_code.py
+++ b/backend/app/api/routes/v1/user_invitation_code.py
@@ -28,6 +28,8 @@ async def generate_invitation_code(
     The code can be shared with a mobile app user who enters it to receive
     SDK access and refresh tokens without manually entering user_id and tokens.
 
+    Previously generated codes for this user are marked as expired.
+
     Requires developer authentication.
     """
     return user_invitation_code_service.generate(db, user_id, developer.id)


.ai/specs/003-invitation-code.md
```diff
@@ -0,0 +1,150 @@
...
+## Existing Code Reused
+
+| What | File |
+|------|------|
+| `create_sdk_user_token()` | `backend/app/services/sdk_token_service.py` |
+| `refresh_token_service.create_sdk_refresh_token()` | `backend/app/services/refresh_token_service.py` |
+| `TokenResponse` schema | `backend/app/schemas/token.py` |
+| `user_service.get()` for user validation | `backend/app/services/user_service.py` |
+| `DeveloperDep` auth dependency | `backend/app/utils/auth.py` |
+| `settings.invitation_expire_days` (7 days) | `backend/app/config.py` |
```
bartmichalak, line 131: invitation_expire_days belongs to the developer email invitation feature (defined under # EMAIL SETTINGS) - user onboarding codes should have their own setting so the two TTLs can evolve independently.

bartmichalak: It's been addresses.

backend/app/services/user_invitation_code_service.py
```diff
@@ -0,0 +1,90 @@
...
+
+        if not invitation_code:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Invalid or expired invitation code",
+            )
+
+        self.crud.mark_redeemed(db_session, invitation_code)
+
+        app_id = f"invite:{invitation_code.created_by_id}"
```
bartmichalak, line 68: I don't think this is a good approach TBH - we rather should bind this to an actual application.id instead.

However, this will be addressed separately, together with the existing admin:{developer_id} pattern in SDK token generation (sdk_token.py:58), which has the same issue

Both flows synthesize a fake app_id string instead of referencing a real Application entity.

Ref issue: https://github.com/the-momentum/open-wearables/issues/443

PR #479, opened 2026-02-17
feat(backend): auto seed admin on backend app startup

bartmichalak:
## Description

Rationale: enable zero-touch deployment on Railway - after clicking "Deploy" the user immediately has an admin account with no manual steps (like running manually python seed script via Railway CLI)

Changes: 
- `seed_admin.py` - reads credentials from env vars (ADMIN_EMAIL/ADMIN_PASSWORD) with fallback to defaults. Check changed from "does admin with this email exist" to "does any developer exist" - seed only creates an account in an empty database.
- `app.sh` - seed runs automatically on startup (after migrations and init scripts, before FastAPI). Idempotent — safe on restarts.

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

run `docker compose up`

## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/backend/scripts/init/seed_admin.py
+++ b/backend/scripts/init/seed_admin.py
@@ -10,14 +10,7 @@ from app.services import developer_service
 def seed_admin(email: str, password: str) -> None:
     """Create default admin developer if it doesn't exist."""
     with SessionLocal() as db:
-        existing = developer_service.crud.get_all(
-            db,
-            filters={},
-            offset=0,
-            limit=1,
-            sort_by=None,
-        )
-        if existing:
+        if developer_service.crud.exists_any(db):
             print("A developer account already exists, skipping admin seed.")
             return
 
--- a/backend/app/repositories/repositories.py
+++ b/backend/app/repositories/repositories.py
@@ -1,6 +1,7 @@
 from uuid import UUID
 
 from pydantic import BaseModel
+from sqlalchemy import exists
 from sqlalchemy.orm import Query
 
 from app.database import BaseDbModel, DbSession
@@ -27,6 +28,9 @@ class CrudRepository[
         db_session.commit()
         db_session.refresh(creation)
         return creation
+
+    def exists_any(self, db_session: DbSession) -> bool:
+        return db_session.query(exists().where(self.model.id.isnot(None))).scalar()
 
     def get(self, db_session: DbSession, object_id: UUID | int) -> ModelType | None:
         return db_session.query(self.model).filter(getattr(self.model, "id") == object_id).one_or_none()


backend/app/config.py
```diff
@@ -59,6 +59,10 @@ class Settings(BaseSettings):
     redis_password: SecretStr | None = None
     redis_username: str | None = None  # Redis 6.0+ ACL
 
+    # ADMIN ACCOUNT SEED
+    admin_email: str = "admin@admin.com"
+    admin_password: str = "secret123"
```
czajkub: id change this to `SecretStr`

backend/scripts/init/seed_admin.py
```diff
@@ -1,37 +1,29 @@
...
-            filters={"email": email},
+            filters={},
             offset=0,
             limit=1,
             sort_by=None,
         )
         if existing:
-            print(f"Admin developer {email} already exists, skipping.")
+            print("A developer account already exists, skipping admin seed.")
             return
```
czajkub, line 15: One question here -  do we always want exactly one admin account to exist? As it stands (in this PR and currently in the repo), we do not have an endpoint for creating a developer account, and here if there is any dev account in the database this is skipped

backend/app/config.py
czajkub: by `this` i meant the password, the email can probably stay as plain `str`

backend/scripts/init/seed_admin.py
```diff
@@ -1,37 +1,29 @@
...
 from app.services import developer_service
 
-DEFAULT_EMAIL = "admin@admin.com"
-DEFAULT_PASSWORD = "secret123"
-
 
 def seed_admin(email: str, password: str) -> None:
     """Create default admin developer if it doesn't exist."""
     with SessionLocal() as db:
         existing = developer_service.crud.get_all(
```
KaliszS: `get_all` repository fetch all columns and because of `all` method, it creates entire `Developer` object even if there's no rows.

The best solution would be to use method `exists()`. After adding import `from sqlalchemy import exists` we would need a new function in developer repository, something like this: 
```py
def exists_any(self, db_session: DbSession) -> bool:
    return db_session.query(exists().where(True)).scalar()
```

But it probably doesn't matter, since it's run once per deployment. So if you prefer to leave it as it is, go ahead. But I recommend to implement that new repository method.

KaliszS (approved):

backend/scripts/init/seed_admin.py
bartmichalak: >we do not have an endpoint for creating a developer account, and here if there is any dev account in the database this is skipped

we do! :)

https://docs.openwearables.io/api-reference/invitations/create-invitation

and it's even implemented in the dashboard:

image

czajkub: Oh, i stand corrected. well, ignore this comment then hahah

backend/scripts/init/seed_admin.py
bartmichalak: Implemented 👍

backend/app/config.py
bartmichalak: Yeah, good suggestion - I've changed this

PR #485, opened 2026-02-18
feat: remove redundant admin seeding from make file and update docs

bartmichalak:
## Description

Admin account is already auto-created on startup via app.sh, so running it again in make init was redundant. 
- Removed the duplicate `seed_admin.py` call, 
- renamed the command from `make init` to `make seed` to better reflect its purpose, 
- updated all docs accordingly.

### General

### Backend Changes

### Frontend Changes

NA

## Testing Instructions

`docker compose up -d` 

## Screenshots

NA

## Additional Notes

NA

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,7 +37,7 @@ open-wearables/
 # Start all services
 docker compose up -d
 
-# Admin account is auto-created on startup (admin@admin.com / secret123)
+# Admin account and series type definitions are auto-created on startup (admin@admin.com / secret123)
 # Seed sample test data (optional)
 make seed
 
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,6 @@ migrate:  ## Apply all migrations
 
 seed:  ## Seed sample data (test users and activity data)
 	$(DOCKER_EXEC) uv sync --group dev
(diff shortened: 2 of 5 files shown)


docs/sdk/flutter/example-app.mdx
```diff
@@ -170,7 +170,6 @@ await HealthBgSync.startBackgroundSync();
     git clone https://github.com/the-momentum/open-wearables.git
     cd open-wearables
     docker compose up -d
-    make init
```
KaliszS, line 173: Why not leaving here just `make seed`?

bartmichalak: I wanted to remove it mostly because the admin account is already created.

However, I completely forgot that `$(DOCKER_EXEC) uv run python scripts/init/seed_series_types.py` is also called there and it's necessary to make the backend work. 

Now I think we should run this `seed_series_types` during app start (`app.sh`) and it'll even make more sense, because once a new type is added to the backend it'll be automatically populated in the database during deployment. 

WDYT?

KaliszS: Good idea!

bartmichalak: Made the changes

PR #487, opened 2026-02-18
docs: add multi-provider SDK architecture plan

bartmichalak:
Architecture plan for expanding the SDK from Apple Health-only to multi-provider (Apple Health, Google Health Connect, Samsung Health) across Flutter and React Native.

Feel free to leave comments and join the discussion!

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive specifications for mobile SDK architecture and unified health payload format across health providers.


.ai/specs/004-mobile-sdk-architecture-plan.mdx
```diff
@@ -0,0 +1,444 @@
...
+- **One schema, all providers**: Same JSON structure regardless of whether data comes from Apple Health, Google Health Connect, or Samsung Health. Field names and nesting are identical.
+- **Types remain provider-specific strings**: The `type` field carries the provider's native identifier (`"HKQuantityTypeIdentifierHeartRate"` for Apple, `"HeartRateRecord"` for Google HC). Backend interprets per provider.
+- **Sleep supports optional `parentId`**: Providers with native sleep sessions (Google HC, Samsung) send a parent session record + child stage records linked via `parentId`. Providers without sessions (Apple) send flat phase records without `parentId` -- backend reconstructs sessions via existing Redis state machine.
+
+**Backend structure**:
+```
+backend/app/
+  schemas/
+    sdk/
+      sync_request.py               # UNIFIED schema for all providers
```
bartmichalak, line 41: not sure about that. Since we have specific types for each provider, shouldn't we have separate schemas (for records, sleep, workouts) - to be able to define custom enums for each provider?

Alternatives:
- not validate types completely (don't think that's a good idea)
- enums aggregating types from all providers (also doesn't seem to be a good idea)

.ai/specs/004-mobile-sdk-architecture-plan.mdx
```diff
@@ -0,0 +1,444 @@
...
+
+#### sleep (sessions and phases)
+
+Flat records with optional `parentId`. Sessions and phases coexist at the same level in the array.
+
+**Why payloads differ per provider -- and why `parentId` is optional:**
+
+The difference stems from how each platform's health API models sleep:
+
+- **Apple HealthKit** has no concept of a "sleep session". It stores sleep as individual `HKCategorySample` records, each representing a single phase transition (e.g., "deep sleep from 22:00 to 22:45"). All records share the same type (`HKCategoryTypeIdentifierSleepAnalysis`) and differ only by `value` (0=IN_BED, 1=ASLEEP_UNSPECIFIED, 2=AWAKE, 3=LIGHT, 4=DEEP, 5=REM). There is no parent record linking them -- HealthKit simply returns a flat list of time-stamped phase samples. Our backend already reconstructs sessions from these flat phases using a Redis-based state machine (`sleep_service.py`) that groups consecutive phases and detects session boundaries via gap analysis (>1 hour gap = session end).
```
bartmichalak: I would add that it's not only in the case of sleep sessions that it differs.

Samsung and Google have the concept of a 'master record' also for health metrics, example:

```json
      // ----- BLOOD_PRESSURE -----
      {
        "uid": "bp-uuid-001",
        "dataType": "BLOOD_PRESSURE",
        "startTime": 1706774400000,
        "endTime": null,
        "dataSource": { "appId": "com.sec.android.app.shealth", "deviceId": "phone-001" },
        "device": { /* ... */ },
        "values": [
          {"type": "SYSTOLIC", "value": 120.0},    // mmHg - systolic pressure (Float)
          {"type": "DIASTOLIC", "value": 80.0},    // mmHg - diastolic pressure (Float)
          {"type": "PULSE", "value": 72.0}         // BPM - pulse during measurement (Float, optional)
        ]
      },
```

bartmichalak (commented):
very preliminary review and just two thoughts off the top of my head, will come back to it tomorrow

.ai/specs/004-mobile-sdk-architecture-plan.mdx
KaliszS: Or we can just create mapping (provider type -> our unified type), which will do validation anyway (we can raise exception if mapping won't recognize type).

KaliszS (changes requested):
Would be useful to have here comparison of what we can get from all 3 providers (in form of table).

.ai/specs/004-mobile-sdk-architecture-plan.mdx
bartmichalak: Yeah, that might be a good approach too 👍

I wanted to use an enum in the API spec mostly for documentation purposes, to explicitly show what we accept on the BE side:

image

But maybe just mentioning that we accept native types and linking to the Apple / Samsung / Google documentation would be enough?

.ai/specs/005-unified-health-payload.mdx
kmlpiekarz, line 1: I attached a document with full explanation, examples, and comparisons.

Main idea:
- We normalize data from Health Connect and Samsung to match the Apple Health structure.
- Sleep data: In Health Connect and Samsung, one record contains stages[]. We split them into separate records to match Apple’s flat structure.
- Workouts: If a workout contains sessions[], we also split them into separate records (like in Apple).
- Other records - If a record contains nested samples or seriesData, we split them into separate records too.

In short, we flatten all nested arrays into individual records to keep one consistent internal structure across all providers.

PR #495, opened 2026-02-19
perf(sleep): eliminate N+1 query in sleep summaries endpoint

### LINKED ISSUE #493: N+1 query in GET /api/v1/users/{user_id}/summaries/sleep
The sleep summary endpoint fires a separate `SELECT avg(value) FROM data_point_series` query **for each sleep record** in the result set. With the default page size of 50, this means up to 50 nearly identical DB queries per request.

The N+1 happens in `summaries_service.py` (~line 250) — the loop over sleep results calls `data_point_repo.get_averages_for_time_range()` individually for each record to fetch physiology averages (currently only heart rate) for that sleep session's time window.

```python
for result in results:           # up to 50 iterations
    ...
    physio_averages = self.data_point_repo.get_averages_for_time_range(
        db_session, user_id, sleep_start, sleep_end, SLEEP_PHYSIO_SERIES_TYPES
    )
```

Each call issues:
```sql
SELECT series_type_definition_id, avg(value)
FROM data_point_series JOIN data_source ON ...
WHERE user_id = ? AND recorded_at >= ? AND recorded_at < ?
  AND series_type_definition_id IN (?)
GROUP BY series_type_definition_id
```

## Suggested fix

Batch the heart rate averaging into a single query by pre-fetching averages for all sleep windows at once — e.g. using a single query with `CASE WHEN` bucketing, a lateral join, or by fetching raw data for the full date range and computing per-window averages in Python.

## Impact

- ~19 occurrences detected on our internal Sentry instance; actual volume will be higher in external/customer deployments
- Transaction duration ~573ms, of which ~74ms spent on repeated queries (currently low, but scales linearly with page size and will worsen with more series types)

Alm0stSurely:
## Problem

Closes #493

The sleep summary endpoint fires a separate SQL query for each sleep record to fetch physiological averages (heart rate). With the default page size of 50, this results in up to 50 nearly identical database queries per request.

## Analysis

The N+1 occurs in `get_sleep_summaries()` where a loop iterates over sleep results and calls `get_averages_for_time_range()` individually for each record.

## Solution

1. **Added batch fetching method**: New `get_averages_for_time_ranges()` method in `DataPointSeriesRepository` that fetches averages for multiple time windows in a single query using PostgreSQL's UNNEST and LATERAL JOIN.

2. **Updated service layer**: Modified `get_sleep_summaries()` to pre-fetch all physiological averages in one batched query.

## Impact

- **Before**: 1 + N queries (1 for sleep records, N for physiological averages)
- **After**: 2 queries total
- **Reduction**: From ~51 queries to 2 queries with default page size

## Notes

- Maintains backward compatibility
- Graceful fallback if batch query fails
- No API contract changes

The existing test `test_get_sleep_summary_with_physiological_metrics` validates the functionality.

### DIFF
--- a/backend/app/repositories/data_point_series_repository.py
+++ b/backend/app/repositories/data_point_series_repository.py
@@ -395,6 +395,122 @@ class DataPointSeriesRepository(
 
         return averages
 
+    def get_averages_for_time_ranges(
+        self,
+        db_session: DbSession,
+        user_id: UUID,
+        time_ranges: list[tuple[datetime, datetime]],
+        series_types: list[SeriesType],
+    ) -> list[dict[SeriesType, float | None]]:
+        """Get average values for multiple time ranges in a single query.
+
+        This method batches the N queries required for N time ranges into a single
+        efficient query using a lateral join, eliminating the N+1 query problem.
+
+        Args:
+            db_session: Database session
+            user_id: User ID to filter by
+            time_ranges: List of (start_time, end_time) tuples (half-open interval)
+            series_types: List of series types to get averages for
+
+        Returns:
+            List of dicts mapping SeriesType to average value, in same order as time_ranges
+        """
+        if not time_ranges or not series_types:
+            return [{t: None for t in series_types} for _ in time_ranges]
+
+        type_ids = [get_series_type_id(t) for t in series_types]
+
+        # Build a VALUES clause for time ranges to use with LATERAL join
+        # This allows us to compute averages for all time windows in one query
+        from sqlalchemy import text
+        from sqlalchemy.dialects.postgresql import ARRAY
+
+        # Create arrays of start and end times for the VALUES clause
+        start_times = [tr[0] for tr in time_ranges]
+        end_times = [tr[1] for tr in time_ranges]
+
+        # Use UNNEST to expand arrays into rows, then lateral join to compute averages
+        # This is more efficient than multiple separate queries
+        range_subq = (
+            db_session.query(
+                func.unnest(start_times).label("range_start"),
+                func.unnest(end_times).label("range_end"),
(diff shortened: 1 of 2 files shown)


backend/app/repositories/data_point_series_repository.py
```diff
@@ -395,6 +395,122 @@ def get_averages_for_time_range(
...
+        """
+        if not time_ranges or not series_types:
+            return [{t: None for t in series_types} for _ in time_ranges]
+
+        type_ids = [get_series_type_id(t) for t in series_types]
+
+        # Build a VALUES clause for time ranges to use with LATERAL join
+        # This allows us to compute averages for all time windows in one query
+        from sqlalchemy import text
+        from sqlalchemy.dialects.postgresql import ARRAY
```
KaliszS, line 427: Definitely we don't want to have imports in the middle of the functions

backend/app/repositories/data_point_series_repository.py
```diff
@@ -395,6 +395,122 @@ def get_averages_for_time_range(
...
+                current_range_idx = range_idx
+                current_averages = {t: None for t in series_types}
+
+            if type_id is not None:
+                try:
+                    series_type = get_series_type_from_id(type_id)
+                    if series_type in current_averages:
+                        current_averages[series_type] = float(avg_value) if avg_value is not None else None
+                except KeyError:
+                    pass
```
KaliszS, line 502: Why pass? It won;t be handled by sentry then

KaliszS (changes requested):
The most important thing is to check (in flower for example) what's the change in celery task execution time. You have added a few joins, our performance may get worse after that. Can you provide comparison (before vs after)?

PR #503, opened 2026-02-19
fix: handle UUID serialization in structured logging (#496)

### LINKED ISSUE #496: UUID serialization error in structured logging masks real API failures
## Description

When any provider API request fails (e.g. timeout, 401, 5xx), the error logging in `api_client.py` passes a `uuid.UUID` object directly to `log_structured()`, which calls `json.dumps()` without a custom encoder. This raises `TypeError: Object of type UUID is not JSON serializable`, masking the original API error.

Currently observed for Whoop in our internal instance (267 occurrences since 2026-02-12, escalating), but the issue affects **all providers** that use `make_authenticated_request()` - including Garmin, Suunto, and Polar - since they all pass `user_id: UUID` through the same code path.

Image

## Impact

- Real API errors are **never logged** - they are replaced by this serialization error
- Debugging actual API issues (expired tokens, rate limits, server errors) is impossible without the original error context
- Affects every user whose sync encounters any API error, regardless of provider

## Root cause

`log_structured()` in `structured_logging.py:64` calls `json.dumps(log_entry)` with no custom JSON encoder. When `make_authenticated_request()` in `api_client.py` catches an API error, it passes `user_id` (a `uuid.UUID` object) as a keyword argument to `log_structured()` (lines 210 and 231). `json.dumps` cannot serialize UUID objects, so the logging itself throws before the original error can be recorded.

## Reproduction path

1. Any Celery sync task calls a provider API via `make_authenticated_request()` with a `UUID` user_id
2. The provider API returns an error (any non-2xx response or network failure)
3. `api_client.py` exception handler calls `log_structured(..., user_id=user_id, ...)`
4. `json.dumps` fails on the UUID object → `TypeError`

## Affected call sites

- `backend/app/services/providers/whoop/data_247.py`
- `backend/app/services/providers/garmin/data_247.py`
- `backend/app/services/providers/garmin/handlers/backfill.py`
- `backend/app/services/providers/suunto/data_247.py`
- `backend/app/services/providers/templates/base_workouts.py` (Polar and others)

## Relevant files

- `backend/app/utils/structured_logging.py:64`
- `backend/app/services/providers/api_client.py:210, 231`

kaifcodec:
Fixed the UUID serialization bug in structured_logging.py. 

The logger was crashing whenever it encountered a UUID object (like user_id), which masked the actual API failures developers needed to see. I've added a helper to handle the conversion and passed it to `json.dumps` so that UUIDs are now correctly logged as strings.

This fixes the issue across all providers (Whoop, Garmin, etc.) since they all rely on this utility.

Fixes #496

### DIFF
--- a/backend/app/utils/structured_logging.py
+++ b/backend/app/utils/structured_logging.py
@@ -2,8 +2,16 @@
 
 import json
 import sys
+import uuid
 from logging import Logger
 from typing import Any
+
+
+def json_serial(obj: Any) -> str:
+    """JSON serializer for objects not serializable by default json code"""
+    if isinstance(obj, uuid.UUID):
+        return str(obj)
+    raise TypeError(f"Object of type {obj.__class__.__name__} is not JSON serializable")
 
 
 def log_structured(
@@ -61,7 +69,7 @@ def log_structured(
     # Emit as single-line JSON directly to stdout
     # This bypasses logger formatters (like Celery's) that add prefixes
     # Platforms will parse this JSON string correctly
-    json_str = json.dumps(log_entry)
+    json_str = json.dumps(log_entry, default=json_serial)
 
     # Always use stdout to avoid Railway's automatic level conversion
     # Platforms can convert stderr logs to level.error automatically, which creates


healthkowshik:
Logs were getting polluted because of this error. Although multiple places used `str()` to avoid this, I did find a few places where this was missed. The approach in this PR looks to be a better solution instead of adding the `str` in every call of the `log_structured` function.

PR #582, opened 2026-03-09
fix(backend): improve Apple XML data source extraction

### LINKED ISSUE #577: AWS Secret Key read as masked string
`settings.aws_secret_access_key` is typed as Pydantic's SecretStr, which masks the value as `*******` when passed
directly. This causes a 403 error from AWS S3 when trying to download the export file.

### LINKED ISSUE #578: HKDevice memory addresses create duplicate data sources during XML import
Apple Health XML stores the device attribute as a full HKDevice description string including a runtime memory address,
e.g.:
`<<HKDevice: 0xb0f577c00>, name:iPhone, manufacturer:Apple Inc., model:iPhone, hardware:iPhone16,1, ...>`
Since memory addresses differ for every record, each unique string creates a separate data_source row in the database.
A single import of one user's data produced 73 distinct data sources for the same physical iPhone.

The _filter_by_priority logic in summaries_service.py selects only one data source per day. With 73 "devices" per day,
up to 98% of step data is silently discarded in the activity summary response, resulting in near-zero step counts for
most days — even though the raw data is correctly stored in the DB.

czajkub:
## Description

- Ensure AWS SecretStr gets extracted correctly to ensure connection to s3 bucket
- Extract device info from XML device, instead of passing entire device string to `device_model`

### General

### Backend Changes

## Screenshots

Data source mappings from export XML file

image

## Additional Notes

Resolves #577 
Resolves #578 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device metadata parsing to reliably extract model and software version.
  * Presigned URL flows now validate cloud storage availability early and return clear 503/403 responses instead of failing later.

* **Chores**
  * Safer S3 client initialization with structured logging and non-throwing credential warnings.

* **Tests**
  * Updated tests to accept 403/503 outcomes when cloud storage is not configured.

### DIFF
--- a/backend/app/integrations/celery/tasks/process_aws_upload_task.py
+++ b/backend/app/integrations/celery/tasks/process_aws_upload_task.py
@@ -4,7 +4,6 @@ from logging import getLogger
 from pathlib import Path
 from uuid import UUID
 
-from fastapi import HTTPException, status
 from sqlalchemy.orm import Session
 
 from app.database import SessionLocal
@@ -13,6 +12,7 @@ from app.services.apple.apple_xml.aws_service import get_s3_client
 from app.services.apple.apple_xml.xml_service import XMLService
 from app.services.timeseries_service import timeseries_service
 from app.services.user_service import user_service
+from app.utils.sentry_helpers import log_and_capture_error
 from app.utils.structured_logging import log_structured
 from celery import shared_task
 
@@ -31,7 +31,14 @@ def process_aws_upload(bucket_name: str, object_key: str, user_id: str | None = 
 
     s3_client = get_s3_client()
     if not s3_client:
-        raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail="S3 client not configured")
+        err = RuntimeError("S3 client not configured — cannot process AWS upload")
+        log_and_capture_error(
+            err,
+            logger,
+            "S3 client unavailable in process_aws_upload task",
+            extra={"bucket_name": bucket_name, "object_key": object_key, "user_id": user_id},
+        )
+        raise err
 
     with SessionLocal() as db:
         temp_xml_file = None


backend/app/integrations/celery/tasks/process_aws_upload_task.py
```diff
@@ -28,6 +29,10 @@ def process_aws_upload(bucket_name: str, object_key: str, user_id: str | None =
         object_key: S3 object key (path)
     """
 
+    s3_client = get_s3_client()
+    if not s3_client:
+        raise HTTPException(status_code=status.HTTP_503_SERVICE_UNAVAILABLE, detail="S3 client not configured")
```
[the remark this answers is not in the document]
czajkub: changed it to a RuntimeError and added log_and_capture_error

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -101,6 +102,35 @@ def _parse_decimal_value(self, raw_value: str | None, metric_type: str) -> Decim
...
+            raw_fields[key.strip()] = value.strip()
+
+        return SourceInfo(
+            name=raw_fields.get("name"),
+            device_id=raw_fields.get("device"),
+            device_model=raw_fields.get("model"),
+            device_manufacturer=raw_fields.get("manufacturer"),
+            device_hardware_version=raw_fields.get("hardware"),
+            device_software_version=raw_fields.get("software"),
+        )
```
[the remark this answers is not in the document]
czajkub: device in XML export always has this sort of structure:
`device="&lt;&lt;HKDevice: 0x66aabba20&gt;, name:Apple Watch, manufacturer:Apple Inc., model:Watch, hardware:Watch6,12, software:26.2, creation date:2026-01-15 22:56:09 +0000&gt;"`
Otherwise its an ill-formed XML file, for example manually tampered with

backend/app/integrations/celery/tasks/process_aws_upload_task.py
```diff
@@ -28,6 +29,17 @@ def process_aws_upload(bucket_name: str, object_key: str, user_id: str | None =
         object_key: S3 object key (path)
     """
 
+    s3_client = get_s3_client()
+    if not s3_client:
```
KaliszS, line 33: Looks like this variable is not used elsewhere, so we could just do `if not get_s3_client():`

backend/app/services/apple/apple_xml/aws_service.py
```diff
@@ -1,17 +1,24 @@
...
-try:
-    s3_client = boto3.client(
-        "s3",
-        region_name=AWS_REGION,
-        aws_access_key_id=settings.aws_access_key_id,
-        aws_secret_access_key=settings.aws_secret_access_key,
-    )
-except NoCredentialsError:
-    raise Exception("AWS credentials not configured")
+def get_s3_client():  # noqa: ANN201
```
KaliszS, line 14: Why not just add returned type `whatever boto3.client returns | None`?

backend/app/integrations/celery/tasks/process_aws_upload_task.py
czajkub: the client is used below but the git diff doesn't show it, before the instance of the client was imported which made error handling harder, also initalising the client this way felt more in line with the rest of the codebase (e.g. redis client or celery worker)

backend/app/services/apple/apple_xml/aws_service.py
czajkub: I wanted to, but the boto3 (aws) module doesn't export the class for the client or use type annotations (at least i couldn't get to it)
image

czajkub: here is the return annotation of the `.client` for clarity

image

PR #583, opened 2026-03-09
feat(backend): Add sleep record handling in Apple Health XML export

### LINKED ISSUE #573: Apple Health XML Import: Multiple bugs preventing correct data display
## Summary

  During testing of the Apple Health XML import feature, several bugs were identified that prevent correct data import
  and display. This report summarizes the findings.

  ---

  ## Bug 1: AWS Secret Key read as masked string (SecretStr)

  **File:** `backend/app/services/apple/apple_xml/aws_service.py`, line 14

  **Problem:**
  `settings.aws_secret_access_key` is typed as Pydantic's `SecretStr`, which masks the value as `*******` when passed
  directly. This causes a 403 error from AWS S3 when trying to download the export file.

  **Fix:**

  Before:  aws_secret_access_key=settings.aws_secret_access_key,

  After: aws_secret_access_key=settings.aws_secret_access_key.get_secret_value()

  ---
  ## Bug 2: HKDevice memory addresses create duplicate data sources

  File: backend/app/services/apple/apple_xml/xml_service.py

  Problem:
  Apple Health XML stores the device attribute as a full HKDevice description string including a runtime memory address,
   e.g.:
  <<HKDevice: 0xb0f577c00>, name:iPhone, manufacturer:Apple Inc., model:iPhone, hardware:iPhone16,1, ...>
  Since memory addresses differ for every record, each unique string creates a separate data_source row in the database.
   A single import of one user's data produced 73 distinct data sources for the same physical iPhone.

  Consequence (Bug 3):
  The _filter_by_priority logic in summaries_service.py selects only one data source per day. With 73 "devices" per day,
   up to 98% of step data is silently discarded in the activity summary response, resulting in near-zero step counts for
   most days — even though the raw data is correctly stored in the DB.

  Suggested fix:
  Normalize the device string when parsing XML records — extract a stable identifier (e.g. hardware field like
  iPhone16,1) instead of storing the full HKDevice description.

  ---
  ## Bug 4: Sleep data not imported

  Problem:
  Sleep data in Apple Health exports is stored as HKCategoryTypeIdentifierSleepAnalysis records with text values (e.g.
  HKCategoryValueSleepAnalysisAsleepDeep). The current XML parser only handles HKQuantityType records with numeric
  values. Sleep records are silently skipped during import.

  Sleep data would need to be parsed as time-range events (start/end datetime + sleep phase) and stored as event_record
  entries with category sleep, similar to how workouts are handled.

  ---
  Environment

  - Backend: FastAPI + Celery
  - Import method: Apple Health Export (XML via S3/SQS)
  - All bugs reproduced on local Docker setup

### LINKED ISSUE #579: Sleep data not imported via Apple Health export XML
Sleep data in Apple Health exports is stored as `HKCategoryTypeIdentifierSleepAnalysis` records with text values (e.g.
`HKCategoryValueSleepAnalysisAsleepDeep`). The current XML parser only handles `HKQuantityType` records with numeric
values. Sleep records are silently skipped during import.

Sleep data would need to be parsed as time-range events (start/end datetime + sleep phase) and stored as event_record
entries with category sleep, similar to how workouts are handled.

czajkub:
## Description

- Add handling of sleep record coming from Apple XML export file
- fix: Add calories burned to workout details 

### General

### Backend Changes

## Screenshots

image

## Additional Notes

The sleep records are passed to the `sleep_service` which normally accepts data from the mobile SDK for apple / android - the information passed is the same, so it's sent there to keep things stable across both ways of importing Apple Health data.

Now, importing via the SDK and XML brings the same results in the database

Resolves #579 
Resolves #573 

## Summary by CodeRabbit

* **New Features**
  * Apple Health sleep data is now supported and automatically imported and synchronized during health file uploads.

* **Improvements**
  * Sleep records are imported alongside heart rate, workouts, and other metrics.
  * Import progress tracking now includes sleep-specific counters and skip-reason logs.
  * Cleanup of stale sleep entries is scheduled after sleep processing.

* **Tests**
  * Tests updated for the extended XML parsing output and reduced mocking of the cleanup task.

### DIFF
--- a/backend/app/services/apple/healthkit/sleep_service.py
+++ b/backend/app/services/apple/healthkit/sleep_service.py
@@ -1,7 +1,6 @@
 import json
 from datetime import datetime
 from logging import getLogger
-from typing import TYPE_CHECKING
 from uuid import UUID, uuid4
 
 from app.config import settings
@@ -20,9 +19,6 @@ from app.schemas.apple.healthkit.sleep_state import SLEEP_START_STATES, SleepSta
 from app.services.apple.healthkit.device_resolution import extract_device_info
 from app.services.event_record_service import event_record_service
 from app.utils.structured_logging import log_structured
-
-if TYPE_CHECKING:
-    from app.integrations.celery.tasks.finalize_stale_sleep_task import finalize_stale_sleeps
 
 redis_client = get_redis_client()
 
@@ -196,6 +192,9 @@ def handle_sleep_data(
             device_model,
         )
         save_sleep_state(user_id, current_state)
+
+    # import not at module level in order to avoid circular import
+    from app.integrations.celery.tasks.finalize_stale_sleep_task import finalize_stale_sleeps
 
     # if we dont call the task, last sleep session in payload will stay
     # in redis until celery beat task runs


backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -19,7 +19,12 @@
     TimeSeriesSampleCreate,
 )
 from app.schemas.apple.apple_xml.stats import XMLParseStats
-from app.schemas.apple.healthkit.sync_request import SourceInfo
+from app.schemas.apple.healthkit.sync_request import (
```
czajkub, line 22: would be changed in the schema refactor pr (#560)

backend/app/integrations/celery/tasks/process_aws_upload_task.py
```diff
@@ -102,11 +103,13 @@ def _import_xml_data(db: Session, xml_path: str, user_id: str) -> None:
...
+    for time_series_records, workouts, sync_request in xml_service.parse_xml(user_id):
         for record, detail in workouts:
             created_record = event_record_service.create(db, record)
             detail_for_record = detail.model_copy(update={"record_id": created_record.id})
             event_record_service.create_detail(db, detail_for_record)
         if time_series_records:
             timeseries_service.bulk_create_samples(db, time_series_records)
             db.commit()
+        if sync_request:
+            handle_sleep_data(db, sync_request, user_id)
```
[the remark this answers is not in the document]
czajkub: yes, youre right, fixing it in previous commit

backend/app/integrations/celery/tasks/process_xml_upload_task.py
```diff
@@ -122,4 +123,7 @@ def _import_xml_data(db: Session, xml_path: str, user_id: str) -> XMLParseStats:
             timeseries_service.bulk_create_samples(db, time_series_records)
             db.commit()
 
+        if sync_request:
+            handle_sleep_data(db, sync_request, user_id)
+
```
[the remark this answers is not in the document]
czajkub: yes, youre right, fixing it in previous commit

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -131,6 +136,33 @@ def _extract_device_info(self, raw_source: str | None) -> SourceInfo:
...
+            "HKCategoryValueSleepAnalysisInBed": SleepPhase.IN_BED,
+        }
+        stage = value_to_stage.get(document.get("value"))
+        if stage is None:
+            return None
+
+        start_date = datetime.fromisoformat(str(document.get("startDate")))
+        end_date = datetime.fromisoformat(str(document.get("endDate")))
+
+        source_info = self._extract_device_info(document.get("source", ""))
```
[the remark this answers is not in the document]
czajkub: yes, youre right, fixing it in previous commit

as for startDate and endDate: theyre required fields in the XML export file, theyll always be there

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -332,7 +398,8 @@ def parse_xml(
                         self.stats.records_skipped,
                         self.stats.workouts_skipped,
                     )
-                    yield time_series_records, workouts
+                    sync_request = self._wrap_sleep_data(sleep_records)
+                    yield time_series_records, workouts, sync_request
                     time_series_records = []
                     workouts = []
```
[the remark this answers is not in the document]
czajkub: yes, youre right, fixing it in previous commit

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -286,27 +333,46 @@ def parse_xml(
...
+
+                    # Handle sleep records
+                    if record.get("type") == "HKCategoryTypeIdentifierSleepAnalysis":
+                        sleep_record = self._normalize_sleep_record(record)
+                        if sleep_record is not None:
+                            sleep_records.append(sleep_record)
+                            self.stats.sleep_records_processed += 1
+                        else:
+                            self.stats.sleep_record_skip(f"unknown_sleep_stage:{record.get('value')}")
+                        elem.clear()
```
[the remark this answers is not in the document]
czajkub: already addressed

backend/app/schemas/apple/apple_xml/stats.py
```diff
@@ -20,3 +22,8 @@ def workout_skip(self, reason: str) -> None:
         """Record a skipped workout with its reason."""
         self.workouts_skipped += 1
         self.skip_reasons[reason] = self.skip_reasons.get(reason, 0) + 1
+
+    def sleep_record_skip(self, reason: str) -> None:
+        """Record a skipped sleep record with its reason."""
+        self.sleep_records_skipped += 1
+        self.skip_reasons[reason] = self.skip_reasons.get(reason, 0) + 1
```
KaliszS: We've got a few issues here. First one - we break DRY rule, because we have 3 methods doing basically the same thing. The second one is how we handle categories of processes and skipped rows. Now everything (records, workouts and sleep) is handled in one container, while it's worth to consider taking everything into more abstract level to distinguish all entities.

My attempt to solve those issue:
```py
from dataclasses import dataclass, field
from collections import Counter

@dataclass
class ParseMetric:
    processed: int = 0
    skipped: int = 0
    reasons: Counter[str] = field(default_factory=Counter)

    def process(self) -> None:
        self.processed += 1

    def skip(self, reason: str) -> None:
        self.skipped += 1
        self.reasons[reason] += 1

@dataclass
class XMLParseStats:
    records: ParseMetric = field(default_factory=ParseMetric)
    workouts: ParseMetric = field(default_factory=ParseMetric)
    sleep: ParseMetric = field(default_factory=ParseMetric)

    def record_skip(self, reason: str) -> None:
        self.general.skip(reason)

    def workout_skip(self, reason: str) -> None:
        self.workouts.skip(reason)

    def sleep_skip(self, reason: str) -> None:
        self.sleep.skip(reason)
```
I haven't tested that code, so if you want to use that, you have to test it first with real data. What has changed:
- imported `collections.Counter`, because it handles potential missing data and start counting from 0 (so we can simplify `self.skip_reasons[reason] = self.skip_reasons.get(reason, 0) + 1` to `self.skip_reasons[reason] += 1`)
- separate `ParseMetric` class to handle processed and skipped rows like a pro ;)
- now all entities have separated skip reasons, but if you prefer, we can go back to shared ones (or we can do something like 
```py
@property 
all_skip_reasons(self) -> Counter:
    return self.general.reasons + self.workouts.reasons + self.sleep.reasons
```

backend/app/services/apple/apple_xml/xml_service.py
```diff
@@ -131,6 +136,35 @@ def _extract_device_info(self, raw_source: str | None) -> SourceInfo:
...
+        """Normalize a sleep record."""
+        value_to_stage = {
+            "HKCategoryValueSleepAnalysisAsleepCore": SleepPhase.ASLEEP_LIGHT,
+            "HKCategoryValueSleepAnalysisAsleepDeep": SleepPhase.ASLEEP_DEEP,
+            "HKCategoryValueSleepAnalysisAsleepREM": SleepPhase.ASLEEP_REM,
+            "HKCategoryValueSleepAnalysisAwake": SleepPhase.AWAKE,
+            "HKCategoryValueSleepAnalysisInBed": SleepPhase.IN_BED,
+            "HKCategoryValueSleepAnalysisAsleep": SleepPhase.SLEEPING,
+            "HKCategoryValueSleepAnalysisAsleepUnspecified": SleepPhase.SLEEPING,
+        }
```
KaliszS: How many times this function is called? Because if we use that frequently, we are going to waste our resources on building that dictionary over and over. It might be better to move that dict outside function definition (as `_SLEEP_VALUE_MAP` for example).

backend/app/services/apple/healthkit/sleep_service.py
```diff
@@ -193,6 +193,13 @@ def handle_sleep_data(
         )
         save_sleep_state(user_id, current_state)
 
+    # import not at module level in order to avoid circular import
+    from app.integrations.celery.tasks.finalize_stale_sleep_task import finalize_stale_sleeps
```
KaliszS: We can still have this at module level to avoid confusion. It's enough to add `if TYPE_CHECKING` from typing library.

KaliszS: I mean:
```py
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from app.integrations.celery.tasks.finalize_stale_sleep_task import finalize_stale_sleeps
```

backend/app/schemas/apple/apple_xml/stats.py
KaliszS: Also we had names:
- `workout_skip`
- `sleep_record_sleep`

We should choose one convention. In the example above I hve normalised it to not use `record` inside both methods names.

KaliszS: @coderabbitai ad to your knowledge lesson to use `collections.Counter` or `collections.defaultdict` if we need to handle missing keys. Also add another note to check whether methods break DRY rule, like it happened here.

backend/app/services/apple/apple_xml/xml_service.py
czajkub: Its called for each sleep stage (so about 20 stages per night), so definitely moving it outside will help. moved it alongside other constants in this file

backend/app/services/apple/healthkit/sleep_service.py
czajkub: added it

backend/app/schemas/apple/apple_xml/stats.py
czajkub: I went with your approach, generalising it even further, also improved logging from the parser (it now uses log_structured, it will look better on railway or any other hosting but this is a lazydocker screenshot)


image

backend/app/services/apple/healthkit/sleep_service.py
czajkub: okay - i tried adding this, and pre-commit just changed it to
```python
# also imported below to avoid circular import, adding it here to avoid confusion
if TYPE_CHECKING:
    pass
```

it looks like it doesnt want to allow it this way

czajkub: probably because it was imported twice, and the other import (the one lower) shadowed the first import, so it showed it as unused and removed it

backend/app/services/apple/apple_xml/xml_service.py
KaliszS: @coderabbitai - another note to save. While doing review, always check how often function/method is called and whether we can make it process faster by moving steps like creating constant data structures out of the function scope.

backend/app/services/apple/healthkit/sleep_service.py
czajkub: nevermind - it worked, I just didnt remove the import in the middle

PR #608, opened 2026-03-16
fix(backend): include timezone offset in datetime saved to database

### LINKED ISSUE #607: Backend ignores zoneOffset from SDK sync payload
The SDK sync endpoint accepts a `zoneOffset` field on `MetricRecord`, `SleepRecord`, and `Workout` schemas, but the import service silently discards it - it's never stored or used in any processing.

Additionally, the field has been coming in as `null` from the mobile SDK in all payloads so far.

The backend should store and use `zoneOffset` when it's present (Apple Watch does provide it for sleep and some workout records - see #606).

czajkub:
## Description

As of now, (at least in the case of the mobile SDK), timezone offset is not saved into the database and is ignored (more about timezones in #606). This PR aims to include the information so that the data can be interpreted correctly.

Resolved #607

### General

### Backend Changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Imported Apple HealthKit workouts, statistics, and time-series now preserve and surface zone offset metadata so timestamps and metric timing remain accurate.

* **New Features**
  * Time-series samples, event records, and workout/sleep responses include an optional zone offset field in import payloads and API responses for clearer timeline context.

* **Chores**
  * Database migration added to persist zone offset for relevant records.


backend/app/services/apple/healthkit/import_service.py
```diff
@@ -44,6 +44,19 @@ def __init__(
         self.timeseries_service = timeseries_service
         self.user_connection_repo = UserConnectionRepository()
 
+    @staticmethod
+    def _parse_zone_offset(offset: str) -> tzinfo:
```
czajkub: if necessary, this method might be moved to some other place if used for other providers, e.g. `app/utils`

backend/app/schemas/events.py
```diff
@@ -22,6 +22,7 @@ class Workout(BaseModel):
     name: str | None = Field(None, example="Morning Run")
     start_time: datetime
     end_time: datetime
+    zone_offset: str | None = None
```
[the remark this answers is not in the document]
czajkub: I'm pretty sure this Workout model is only used for retrieving the data that's already in the database and expose it via an endpoint. Since we already validate the offset on input, it shouldn't be necessary to validate it again and it will improve readibility this way, no?

backend/app/repositories/data_point_series_repository.py
```diff
@@ -145,6 +145,7 @@ def _insert_data_points(
                     "external_id": creator.external_id,
                     "data_source_id": source_id,
                     "recorded_at": creator.recorded_at,
+                    "zone_offset": creator.zone_offset,
```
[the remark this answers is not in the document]
czajkub: resolved with last commits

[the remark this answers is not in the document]
czajkub: actually pushed it just now

backend/app/repositories/event_record_repository.py
```diff
@@ -132,6 +132,7 @@ def bulk_create(
                     "duration_seconds": creator.duration_seconds,
                     "start_datetime": creator.start_datetime,
                     "end_datetime": creator.end_datetime,
+                    "zone_offset": creator.zone_offset,
```
czajkub, line 135: resolved with last commits

backend/app/schemas/event_record.py
```diff
@@ -47,6 +47,7 @@ class EventRecordBase(BaseModel):
     duration_seconds: int | None = None
     start_datetime: datetime
     end_datetime: datetime
+    zone_offset: str | None = None
```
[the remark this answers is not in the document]
czajkub: added validation, good idea

backend/app/schemas/events.py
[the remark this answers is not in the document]
czajkub: i already added the validation on input

backend/app/services/apple/healthkit/import_service.py
KaliszS: yes, it should have been in utils imo

backend/app/schemas/timeseries.py
```diff
@@ -69,9 +83,23 @@ class TimeSeriesSampleBase(BaseModel):
...
+        pattern=r"^[+-]\d{2}:\d{2}$",
+        examples=["+01:00", "-05:30"],
+        max_length=10,
+    )
     value: Decimal | float | int
     series_type: SeriesType
 
+    @field_validator("zone_offset", mode="before")
+    @classmethod
+    def normalize_zone_offset(cls, v: str | None) -> str | None:
```
KaliszS: I've seen th same function implemented 3 times. Can we make this to be defined once? Or it's not so easy with pydantic validators?

czajkub: Yes, i learned that this can be handled with `Annotated` from typing, to annotate a type with corresponding pydantic field and validator:

```python
def _normalize_zone_offset(v: str | None) -> str | None:
    if v == "Z":
        return "+00:00"
    return v


ZoneOffset = Annotated[
    str | None,
    Field(
        None,
        description="Timezone offset in the format '+01:00' or '-05:30'",
        pattern=r"^[+-]\d{2}:\d{2}$",
        examples=["+01:00", "-05:30"],
        max_length=10,
    ),
    BeforeValidator(_normalize_zone_offset),
]

```

KaliszS: Nice finding! @coderabbitai you should also learn that as a pydantic trick, might be useful for reviews.

PR #617, opened 2026-03-16
chore: add production docker-compose

bartmichalak:
## Description

Production-ready docker-compose that uses the prod Dockerfile for frontend (Nitro SSR build instead of vite dev server), drops dev-only `develop.watch` configs, and sets `restart: unless-stopped` on all services.

Usage: `docker compose -f docker-compose.prod.yml up -d --build`

## Testing Instructions

**Steps to test:**
1. Deploy to server with `docker compose -f docker-compose.prod.yml up -d --build`
2. Verify frontend serves built assets (not vite dev server)
3. Verify all services restart after failure

**Expected behavior:**
All services come up with production configs, frontend runs Nitro SSR server on port 3000.

## Summary by CodeRabbit

* **Chores**
  * Configured production deployment infrastructure for the complete application stack, including database services, backend services, caching layer, task processing, monitoring capabilities, and frontend components.

### DIFF
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -102,6 +102,8 @@ services:
     build:
       context: ./frontend
       dockerfile: Dockerfile
+      args:
+        VITE_API_URL: ${VITE_API_URL:?Set VITE_API_URL in your environment}
     container_name: frontend__open-wearables
     image: open-wearables-frontend:latest
     pull_policy: never


bartmichalak:
You're right that `develop.watch` is ignored with `docker compose up` - for the backend that approach works well. 

The issue is mainly with the frontend - in dev we use `Dockerfile.dev` (vite dev server with HMR), while in prod we use `Dockerfile` (Nitro SSR build). This can't be solved with a single file without profiles/overrides, since it's a different Dockerfile and different build args (`VITE_API_URL` must be passed as a build arg in production).

This could be done with `docker-compose.override.yml` or compose profiles, but a separate file is simpler and more readable. 

@kczpl any thougths on that?

Issue #169, opened 2025-12-10
Stress testing for high-frequency data access


bartmichalak:
We need stress tests for high-volume data scenarios - especially heart rate and steps samples, which can generate thousands of records per day per user.

What to test:
**Database:**
- Query performance with 1M+ samples for a single user (typical for several months of HR data at 1 sample/min)
- Index efficiency on (user_id, timestamp) for range queries
- Concurrent reads/writes - simulating multiple users syncing data simultaneously

**Endpoints:**
- GETs with various date ranges (day/week/month)
- Response time when paginating through large datasets
- Memory usage when serializing large responses

⚠️ Endpoint testing is blocked by this issue: https://github.com/the-momentum/open-wearables/issues/98

lavneethora:
Happy to work on this. Can I be assigned?

PR #619, opened 2026-03-16
feat: add S3 raw payload replay script

bartmichalak:
## Description

Standalone script to replay S3-stored raw payloads back through the SDK sync endpoint - for reproducing issues, re-processing data after fixes, or load testing. Supports filtering by provider/source/date, dry-run mode, configurable delay, and replaying to a different user than the source. Also added docs section to the raw payload storage guide.

### General

### Backend Changes

## Testing Instructions

**Steps to test:**
1. Enable raw payload storage (`RAW_PAYLOAD_STORAGE=s3`) and send some SDK syncs to populate S3
2. Run with `--dry-run` to verify listing/filtering works:
   ```bash
   cd backend && set -a && source config/.env && set +a
   uv run --with boto3,httpx python scripts/replay_raw_payloads.py --user-id <uuid> --dry-run
   ```
3. Run against local Docker backend to verify replay:
   ```bash
   uv run --with boto3,httpx python scripts/replay_raw_payloads.py --user-id <uuid> --api-key sk-... --delay 0.5
   ```

**Expected behavior:**
- Dry run lists payloads sorted by upload timestamp
- Replay sends each payload to the SDK sync endpoint and reports success/fail per payload

https://github.com/user-attachments/assets/fd33523e-e7a6-4422-90ae-b2c3a568ecac

## Summary by CodeRabbit

* **New Features**
  * Added a standalone payload replay tool to resend stored payloads through the SDK sync endpoint with filtering by user ID, provider, source, and date range; supports delay, limit, and dry-run modes.

* **Documentation**
  * Added a "Replaying Stored Payloads" guide with setup, configuration, usage examples, filtering options, and a detailed options table.

### DIFF
--- a/docs/dev-guides/raw-payload-storage.mdx
+++ b/docs/dev-guides/raw-payload-storage.mdx
@@ -216,6 +216,11 @@ uv run --with boto3,httpx python scripts/replay_raw_payloads.py \
 | `--api-url` | `http://localhost:8000` | Backend API base URL |
 | `--api-key` | `OPEN_WEARABLES_API_KEY` env var | API key for authentication |
 | `--s3-bucket` | `RAW_PAYLOAD_S3_BUCKET` or `AWS_BUCKET_NAME` env var | S3 bucket name |
+| `--s3-prefix` | `raw-payloads` | S3 key prefix |
+| `--s3-endpoint-url` | `RAW_PAYLOAD_S3_ENDPOINT_URL` env var | Custom S3 endpoint for compatible providers |
+| `--aws-region` | `AWS_REGION` env var or `eu-north-1` | AWS region |
+| `--aws-access-key-id` | `AWS_ACCESS_KEY_ID` env var | AWS access key ID |
+| `--aws-secret-access-key` | `AWS_SECRET_ACCESS_KEY` env var | AWS secret access key |
 | `--provider` | all | Filter by provider (`apple`, `samsung`, etc.) |
 | `--source` | all | Filter by source (`sdk`, `webhook`) |
 | `--date-from` | none | Start date filter (YYYY-MM-DD) |


backend/scripts/replay_raw_payloads.py
czajkub, line 1: A few comments on this script:
- When AWS credentials are not set via flags and are not in environmental variables, the script throws an uncaught `botocore.exceptions.NoCredentialsError` - we may want to verify that there are credentials present before initalizing boto3 client
- When creds are set, but the user has non-AWS s3-compatible storage and doesnt specify endpoint URL via flag, `botocore.exceptions.ClientError: An error occurred (InvalidAccessKeyId) when calling the ListObjectsV2 operation: The AWS Access Key Id you provided does not exist in our records` will be thrown - we could wrap the `s3_client` init in a try-except block with error message printed in case of error, so the user doesn't have to see the whole traceback
- For stress tests - even when delay is set to 0, the requests are sent synchronously - we wait for the response each time before sending the next one. If this is intended behaviour then this is fine, but if we want to send it async we can use `httpx.AsyncClient()`
- On successful sync message the backend sends back `OK 200` - this is not connected to the script, but from what I recall the status code response was supposed to be 202, this may have to do with how the sync endpoint is configured in fastapi

Other than that, the script works fine :D

bartmichalak: Thanks for the thorough review!

First two points are addressed. 

As for concurrent/async requests - this is intentionally synchronous. The script is meant to mimic mobile SDK behavior, which sends a payload, waits for the server response, and only then sends the next one. 
For stress tests I think we could figure our another approach, let's consider it as part of https://github.com/the-momentum/open-wearables/issues/169

Re: 200 vs 202 - good catch, you're right. The endpoint returns status_code: 202 in the JSON body, but the actual HTTP status code is 200 because FastAPI defaults to that. We're missing status_code=202 on the route decorator. I think for now it's not very important to change it.

PR #631, opened 2026-03-18
fix(backend): add password length validation to PasswordChange schema

### LINKED ISSUE #605: feat: Add baseline password length validation to PasswordChange schema
## Summary

The `PasswordChange` schema in `backend/app/schemas/developer.py` currently accepts empty or arbitrarily short passwords at the request-schema level, which is inconsistent with the validation enforced elsewhere (e.g., `DeveloperCreate` enforces `min_length=8` on its `password` field).

## Suggested Change

Update the `PasswordChange` Pydantic model to enforce minimum length constraints on `new_password` and `confirm_password` using Pydantic v2+ syntax:

```python
class PasswordChange(BaseModel):
    current_password: str = Field(..., min_length=1)
    new_password: str = Field(..., min_length=8)
    confirm_password: str = Field(..., min_length=8)
```

This ensures validation is enforced at the schema boundary, consistent with coding guidelines that require schemas to implement validation rather than delegating it to database models.

## Motivation

- Keeps validation consistent with existing schemas (e.g., `DeveloperCreate`).
- Prevents empty/short password values from reaching the service layer.
- Aligns with the project coding guideline: *"backend/app/schemas/**/*.py: Schemas layer must use Pydantic 2+ syntax exclusively with validation implemented in schemas, not database models"*.

## References

- Raised during review of PR #550: https://github.com/the-momentum/open-wearables/pull/550
- Original review comment: https://github.com/the-momentum/open-wearables/pull/550#discussion_r2929252520
- Requested by @bartmichalak

bartmichalak:
## Description

`PasswordChange` schema had no length validation on passwords - frontend already enforces min 8 chars but the backend would happily accept empty or 1-char passwords. This aligns the schema with `DeveloperCreate`/`DeveloperUpdate` which already had `min_length=8`. 

Also extracted `MIN_PASSWORD_LENGTH` constant so the value lives in one place.

Closes #605

### General

### Backend Changes

## Testing Instructions

**Steps to test:**
1. `POST /api/v1/auth/change-password` with `new_password` shorter than 8 chars
2. Should get a 400 validation error instead of proceeding to the service layer

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized password validation across account flows (creation, updates, invitation acceptance, and password changes). Minimum password length is now configurable (default: 8). Current password remains required for changes; confirmation matching still enforced.

* **Tests**
  * Added test ensuring password-change requests reject passwords shorter than the configured minimum.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -52,6 +52,8 @@ class Settings(BaseSettings):
     algorithm: str = "HS256"
     access_token_expire_minutes: int = 60
     token_lifetime: int = 3600
+
+    # VALIDATION SETTINGS
     min_password_length: int = 8
 
     # REDIS SETTINGS


backend/app/schemas/developer.py
```diff
@@ -48,9 +50,9 @@ class DeveloperUpdateInternal(BaseModel):
 
 
 class PasswordChange(BaseModel):
-    current_password: str
-    new_password: str
-    confirm_password: str
+    current_password: str = Field(..., min_length=1)
```
KaliszS: why minimum length=1 here?

backend/app/schemas/invitation.py
```diff
@@ -4,6 +4,8 @@
 
 from pydantic import BaseModel, ConfigDict, EmailStr, Field
 
+from app.schemas.developer import MIN_PASSWORD_LENGTH
```
bartmichalak: @KaliszS 
Not sure about importing the constant cross-schema like this (invitation.py -> developer.py). We could extract it to something like `backend/app/constants.py` instead - what's the convention we're going with for shared constants?

backend/app/schemas/developer.py
bartmichalak: Good catch, thanks! I initially went with min_length=1 because Claude Code suggested we might have allowed shorter passwords in the past and it'd be safer to keep backward compatibility. 

Checked the full git history after your comment of this file and turns out we've always enforced min_length=8 on passwor creation - never allowed anything shorter. 

So there's no backward compat concern here. Changed to `MIN_PASSWORD_LENGTH`.

backend/app/schemas/invitation.py
KaliszS: We already have constants directory, but the best place to got for such variables would be just `app/backend/`config.py` with default value imo.

And in the `min=1` comment I meant I think we don't need to limit that field, because all we need is to just check whether password is correct.

Issue #622, opened 2026-03-17
Migrate database adapter to async version


KaliszS:
Continuation of #571 

It will require to drop psycopg probably.

Issue #571, opened 2026-03-08
Sync database calls inside `async def` endpoints block the event loop


sLeeNguyen:
### Summary

Throughout the codebase, `async def` route handlers use a synchronous SQLAlchemy Session (`DbSession`) and call synchronous service/repository methods directly on the event loop thread. This blocks the uvicorn event loop on every request, degrading concurrency across the entire API rather than just the affected endpoint. This creates a scalability bottleneck under load.

### Problem

`DbSession` is a **synchronous** SQLAlchemy Session:

```python
# database.py
DbSession = Annotated[Session, Depends(_get_db_dependency)]
```

This is injected into `async def` handlers across many routes (e.g. `oauth.py`, `garmin_webhooks.py`, `sdk_sync.py`, and others). Inside these async handlers, synchronous database operations are executed directly:

```python
@router.post("/push")
async def garmin_push_notification(db: DbSession, ...) -> dict:
    ...
    # Blocking call executed on the event loop thread
    connection = user_connection_repo.get_by_provider_user_id(db, "garmin", garmin_user_id)
    ...
```

Since this function is synchronous and not awaited or offloaded to a worker thread, it blocks the event loop until the database call completes.

### Root Cause

FastAPI execution behavior differs depending on the handler type:
- `def` endpoints → automatically executed in a **thread pool**
- `async def` endpoints → executed **directly on the event loop**

Therefore, any blocking I/O inside `async def` (such as synchronous database queries) will stall the event loop and delay all concurrent requests.

### Impact

This affects all `async def` endpoints that use synchronous database access. The impact is especially severe for endpoints that process large batched payloads, such as: `/garmin/push`, `/garmin/ping`. These endpoints may perform many database reads/writes per request, significantly increasing the time the event loop is blocked.

In our deployment environment, the service runs on Kubernetes with health/readiness probes. When /garmin/push is triggered, the blocking database operations can stall the event loop, preventing the health check endpoint from responding in time. This causes Kubernetes to mark the pod as unhealthy and restart it.

### Proposed Fix
- Option 1: Convert endpoints to synchronous handlers. 
- Option 2: Migrate to fully async database access (`AsyncSession`). This is the most scalable approach but requires a broader refactor.

PR #632, opened 2026-03-19
feat(api): add provider disconnect endpoint

### LINKED ISSUE #615: Implement provider disconnect endpoint
We don't have a way for developers to programmatically disconnect a user from a provider.

**Scenario:** A wellbeing app uses Open Wearables as middleware to pull health data from wearables. When a user goes to the app's settings and taps "Disconnect Garmin" (or any other provider), the app's backend needs to tell Open Wearables to revoke that connection. Currently there's no endpoint for that.

## What's needed

- `DELETE /api/v1/users/{user_id}/connections/{provider}` - revokes a user's connection to a given provider
  - For OAuth providers: revoke tokens at the provider (if supported), clear stored tokens, mark connection as `REVOKED`
  - For SDK-based providers: mark connection as `REVOKED` (client-side cleanup is handled by the SDK's `signOut()`)
- Document the endpoint so developers know how to integrate disconnect into their apps

## What already exists
- `mark_as_revoked()` in `UserConnectionRepository`
- Garmin deregistration webhook (provider-initiated disconnect) - this issue covers the developer-initiated direction
- Frontend "Disconnect" button (currently a no-op with a TODO)

## Frontend follow-up
- Wire the dashboard's "Disconnect" button to call the new endpoint

bartmichalak:
## Description

Adds `DELETE /api/v1/users/{user_id}/connections/{provider}` so apps can programmatically disconnect a user from a provider. Revokes the connection, clears tokens, and is idempotent (204 for nonexistent/already-revoked). Frontend disconnect button on connection card now works with a confirmation dialog.

Closes #615

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. `curl -X DELETE http://localhost:8000/api/v1/users/{user_id}/connections/garmin -H "X-Open-Wearables-API-Key: KEY"` - expect 204
2. Same DELETE again - still 204 (idempotent)
3. Frontend: click Disconnect on a connection card, confirm dialog, card shows "Revoked" badge

**Expected behavior:**
- Active/expired connections get revoked, tokens cleared
- Already-revoked or nonexistent connections return 204 silently
- Invalid provider returns 400

## Other notes

We need to implement disconnect API calls separately, here is the issue for Garmin:
https://github.com/the-momentum/open-wearables/issues/633

## Summary by CodeRabbit

* **New Features**
  * Add provider disconnect flow: confirmation dialog triggers server-backed DELETE, returns 204 on success
  * Disconnect option hidden for revoked connections and disabled while processing
  * Clears stored auth tokens and expiry for disconnected connections
  * Shows success/error toast and refreshes connections list after action

* **Tests**
  * Comprehensive disconnect endpoint tests: success, idempotency, validation, auth, token-clearing, and cross-user isolation

### DIFF
--- a/backend/app/schemas/developer.py
+++ b/backend/app/schemas/developer.py
@@ -4,6 +4,8 @@ from uuid import UUID, uuid4
 from pydantic import BaseModel, ConfigDict, EmailStr, Field, model_validator
 
 from app.config import settings
+
+
 class DeveloperRead(BaseModel):
     model_config = ConfigDict(from_attributes=True)
 


backend/app/api/routes/v1/connections.py
```diff
@@ -20,3 +24,21 @@ def get_connections_endpoint(
...
+) -> Response:
+    """Disconnect a user from a provider, revoking the connection and clearing tokens."""
+    connection = connection_repo.get_by_user_and_provider(db, UUID(user_id), provider.value)
+
+    if connection is None or connection.status == ConnectionStatus.REVOKED:
+        return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+    connection_repo.disconnect(db, connection)
+    logger.info("Disconnected user %s from provider %s", user_id, provider.value)
+    return Response(status_code=status.HTTP_204_NO_CONTENT)
```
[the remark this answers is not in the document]
KaliszS: @coderabbitai you are wrong here. Repo method is sync, so if we use async we would block event loop and gain nothing. Check out issue #571. There is already plan to convert adapter to async version, but it waits to be implemented in issue #622.

backend/app/api/routes/v1/connections.py
```diff
@@ -20,3 +24,21 @@ def get_connections_endpoint(
...
+def disconnect_provider_endpoint(
+    user_id: str,
+    provider: ProviderName,
+    db: DbSession,
+    _api_key: ApiKeyDep,
+) -> Response:
+    """Disconnect a user from a provider, revoking the connection and clearing tokens."""
+    connection = connection_repo.get_by_user_and_provider(db, UUID(user_id), provider.value)
+
+    if connection is None or connection.status == ConnectionStatus.REVOKED:
```
KaliszS: `get_by_user_and_provider` repo method always returns `UserConnection | None`, so we can just do simply `if not connection or connection.status == ConnectionStatus.REVOKED`

backend/app/api/routes/v1/connections.py
```diff
@@ -1,12 +1,16 @@
...
 from app.repositories import UserConnectionRepository
-from app.schemas import UserConnectionRead
+from app.schemas import ConnectionStatus, UserConnectionRead
+from app.schemas.oauth import ProviderName
 from app.services import ApiKeyDep
 
+logger = logging.getLogger(__name__)
+
 router = APIRouter()
 connection_repo = UserConnectionRepository()
```
KaliszS: We already have `user_connection_service`, so we should move logic with using repository and logging things right there (for both endpoints) and here we should just call that service.

backend/app/api/routes/v1/connections.py
```diff
@@ -20,3 +24,21 @@ def get_connections_endpoint(
...
+
+@router.delete("/users/{user_id}/connections/{provider}", status_code=status.HTTP_204_NO_CONTENT)
+def disconnect_provider_endpoint(
+    user_id: str,
+    provider: ProviderName,
+    db: DbSession,
+    _api_key: ApiKeyDep,
+) -> Response:
+    """Disconnect a user from a provider, revoking the connection and clearing tokens."""
+    connection = connection_repo.get_by_user_and_provider(db, UUID(user_id), provider.value)
```
KaliszS: It is redundant, because now we doing two database calls:
1. fetch object as ORM object
2. update that object in db

We could just run one query and avoid race condition (someone might update the connection between our fetch and update), read more in the repository comment.

backend/app/repositories/user_connection_repository.py
```diff
@@ -127,6 +127,18 @@ def get_expiring_tokens(self, db_session: DbSession, minutes_threshold: int = 5)
...
+        connection.status = ConnectionStatus.REVOKED
+        connection.access_token = None
+        connection.refresh_token = None
+        connection.token_expires_at = None
+        connection.updated_at = datetime.now(timezone.utc)
+        db_session.add(connection)
+        db_session.commit()
+        db_session.refresh(connection)
+        return connection
+
```
KaliszS: We could just build one update query with `WHERE` statement to check whether it exists and is not revoked already. Also at the end we might return something like `result.rowcount > 0` if we need to know whether we have updated something or not. But after all, 204 code used in route when we didn't update anything, wasn't verbose enough anyway and didn't indicate something was updated.

backend/app/schemas/developer.py
```diff
@@ -3,6 +3,8 @@
 
 from pydantic import BaseModel, ConfigDict, EmailStr, Field, model_validator
 
+MIN_PASSWORD_LENGTH = 8
```
KaliszS: It is already added in another PR, so there is no need to duplicate logic here. Especially it's already outdated and you might accidently override correct solution with this one.

KaliszS (changes requested):

backend/app/schemas/developer.py
bartmichalak: Yeah, committed accidentely

backend/app/api/routes/v1/connections.py
bartmichalak: @coderabbitai, why didn't you catch this? 

In backend/agents.md we have following rules:
- "Keep route code minimal, delegate to services"                                                                                                  
- `Flow: request → main.py → head_router → version_router → router → endpoint → service`

Should we specify something more in Agents.md?

bartmichalak: @KaliszS what do you think? 

Are we gonna add this?

bartmichalak: https://github.com/the-momentum/open-wearables/pull/634

backend/app/api/routes/v1/connections.py
bartmichalak: It's just a semantic matter, right? Or am I missing something more important?

backend/app/api/routes/v1/connections.py
bartmichalak: Changed ✅

backend/app/services/user_connection_service.py
```diff
@@ -23,5 +25,23 @@ def get_active_count_in_range(self, db_session: DbSession, start_date: datetime,
         """Get count of active connections created within a date range."""
         return self.crud.get_active_count_in_range(db_session, start_date, end_date)
 
+    def get_connections_by_user(self, db_session: DbSession, user_id: UUID) -> list[UserConnectionRead]:
+        """Get all connections for a user."""
+        connections = self.crud.get_by_user_id(db_session, user_id)
+        return [UserConnectionRead.model_validate(conn) for conn in connections]
```
KaliszS: In our architecture services usually should return SQLALchemy model, so no `.model_validate` method is required. Instead in FastAPI endpoint we should add `response_model` which will handle converting ORM model to Pydantic schema.

@coderabbitai that's a lesson for you.

backend/app/services/user_connection_service.py
```diff
@@ -23,5 +25,23 @@ def get_active_count_in_range(self, db_session: DbSession, start_date: datetime,
...
+        """Disconnect a user from a provider. Raises 404 if connection not found."""
+        updated = self.crud.disconnect(db_session, user_id, provider)
+        if updated:
+            self.logger.info("Disconnected user %s from provider %s", user_id, provider)
+            return
+
+        # Nothing updated - check if connection exists (already revoked) or not found
+        connection = self.crud.get_by_user_and_provider(db_session, user_id, provider)
+        if not connection:
+            raise ResourceNotFoundError("connection", user_id)
```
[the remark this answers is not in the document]
bartmichalak: We intentionally distinguish between two cases: - Connection exists but already revoked - returns 204 (true idempotency, tested in test_disconnect_already_revoked_is_idempotent)

Connection never existed - returns 404 (bad user_id, typo in provider name, etc.)

backend/app/repositories/user_connection_repository.py
bartmichalak: Changed ✅

backend/tests/api/v1/test_connections.py
```diff
@@ -278,3 +280,180 @@ def test_get_connections_excludes_sensitive_data(self, client: TestClient, db: S
...
+        headers = api_key_headers(api_key.id)
+
+        # Act
+        client.delete(f"/api/v1/users/{user.id}/connections/garmin", headers=headers)
+
+        # Assert
+        conn = db.query(UserConnection).filter_by(user_id=user.id, provider="garmin").one()
+        assert conn.access_token is None
+        assert conn.refresh_token is None
+        assert conn.token_expires_at is None
```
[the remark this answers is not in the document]
bartmichalak: Fixed ✅

backend/app/services/user_connection_service.py
KaliszS: @coderabbitai do you also have a note to always try to use only services inside fastapi routes? Repositories should be used only within services if possible. There might be exceptions, but they should be well justified.

PR #643, opened 2026-03-20
fix: hide Sync Now button for SDK-based providers

### LINKED ISSUE #642: Hide "Sync Now" button for SDK-based providers (Apple, Samsung, Google)
## Problem

SDK-based providers (Apple Health, Samsung Health, Google Health Connect) currently display a "Sync Now" button on the connection card - identical to API-based providers like Whoop or Oura.

Image

This is misleading because SDK-based providers work fundamentally differently from API-based ones:

- **API-based providers** (Whoop, Oura, Polar, etc.) - we can pull data on demand from a cloud API, so "Sync Now" makes sense
- **SDK-based providers** (Apple, Samsung, Google) - data is pushed from the mobile device whenever the SDK delivers it. There is no cloud API to pull from

We have no control over the mobile device's sync schedule. The device OS has full authority over background activity and may limit it due to low battery, locked screen (Apple), battery optimization settings (Android), or other system-level restrictions.

Showing "Sync Now" implies we can force a sync, which we can't. At best it's a no-op, at worst it confuses users who expect it to trigger their device to send data.

Deepak8858:
Fixes #642

### Changes
- Hide the "Sync Now" button on the connection card for SDK-based providers: `apple`, `samsung`, `google`, and `auto-health-export`.
- Maintained existing behavior where the button is hidden for `garmin` (due to special backfill handling).

### Why
- SDK-based providers rely on data being pushed from the mobile device. Showing a "Sync Now" button is misleading as we cannot force a sync from the cloud side.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated "Sync Now" button availability for health data providers. The button no longer appears for Apple, Samsung, Google, Garmin, and auto-health-export connections.

### DIFF
--- a/frontend/src/components/user/connection-card.tsx
+++ b/frontend/src/components/user/connection-card.tsx
@@ -373,8 +373,10 @@ export function ConnectionCard({ connection, className }: ConnectionCardProps) {
             </div>
           )}
 
-        {/* Sync button - only for non-Garmin providers */}
-        {connection.provider !== 'garmin' && (
+        {/* Sync button - only for cloud API-based providers (excludes SDK and Garmin) */}
+        {!['garmin', 'apple', 'samsung', 'google', 'auto-health-export'].includes(
+          connection.provider
+        ) && (
           <div className="flex gap-2">
             <Button
               variant="outline"


frontend/src/components/user/connection-card.tsx
```diff
@@ -373,8 +373,10 @@ export function ConnectionCard({ connection, className }: ConnectionCardProps) {
             </div>
           )}
 
-        {/* Sync button - only for non-Garmin providers */}
-        {connection.provider !== 'garmin' && (
+        {/* Sync button - only for cloud API-based providers (excludes SDK and Garmin) */}
+        {!['garmin', 'apple', 'samsung', 'google', 'auto-health-export'].includes(
```
bartmichalak, line 377: @Deepak8858 do you think it's a good idea to reference our providers by strings instead of enums?

PR #666, opened 2026-03-24
refactor: replace per-provider redirect URI settings with shared API_BASE_URL

bartmichalak:
Replace 8 individual `*_REDIRECT_URI` env vars with a single `API_BASE_URL` - redirect URIs are now built dynamically via `settings.oauth_redirect_uri(ProviderName.X)`. There's no reason to set this individually per provider since the only difference is the provider name in the path. The previous setup was error-prone.

Per-provider `*_REDIRECT_URI` env vars still work but emit a `DeprecationWarning` pointing users to `API_BASE_URL`, so existing deployments won't break on upgrade.

## Summary by CodeRabbit

* **New Features**
  * Added a centralized API base URL setting and a helper to compute provider callback URLs.

* **Deprecations**
  * Per-provider redirect URI settings/env vars are deprecated; legacy values remain honored but emit deprecation warnings.

* **Refactor**
  * OAuth redirect URI resolution unified into a single lookup used by all providers; provider integrations now use the centralized lookup.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -178,7 +178,7 @@ class Settings(BaseSettings):
         # This should never be reached given the type annotation, but ensures type safety
         raise ValueError(f"Unexpected type for cors_origins: {type(v)}")
 
-    def oauth_redirect_uri(self, provider: "ProviderName") -> str:
+    def oauth_redirect_uri(self, provider: ProviderName) -> str:
         """Build OAuth redirect URI for a provider.
 
         Uses the legacy per-provider *_REDIRECT_URI env var if set,


backend/app/config.py
```diff
@@ -169,6 +178,23 @@ def assemble_cors_origins(cls, v: str | list[str]) -> list[str] | str:
...
+        legacy_value = getattr(self, legacy_attr, None)
+        if legacy_value is not None:
+            warnings.warn(
+                f"{legacy_attr.upper()} is deprecated, use API_BASE_URL instead.",
+                DeprecationWarning,
+                stacklevel=2,
+            )
+            return legacy_value
+        return f"{self.api_base_url}/api/v1/oauth/{provider.value}/callback"
+
```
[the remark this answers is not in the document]
bartmichalak: @coderabbitai doesn't Pydantic default these to None, not "", so the empty-string guard is unnecessary?

backend/config/.env.example
```diff
@@ -62,26 +62,25 @@ SYNC_INTERVAL_SECONDS=3600  # How often to run automatic sync (default: 1 hour)
 
 #--- Providers ---#
 
+# Public base URL of the API (used for OAuth redirect URIs, etc.)
+API_BASE_URL=http://localhost:8000
+
```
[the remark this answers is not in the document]
bartmichalak: @coderabbitai good catch, removed

backend/app/config.py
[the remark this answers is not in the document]
bartmichalak: @coderabbitai Ok, got it, but I don't think we should protect ourselves against such a circumstance. I don't see a scenario when someone sets empty string for oauth url having api_base_url set in the same time.

I would leave it as it is.

backend/app/config.py
```diff
@@ -169,6 +178,23 @@ def assemble_cors_origins(cls, v: str | list[str]) -> list[str] | str:
         # This should never be reached given the type annotation, but ensures type safety
         raise ValueError(f"Unexpected type for cors_origins: {type(v)}")
 
+    def oauth_redirect_uri(self, provider: "ProviderName") -> str:
```
KaliszS: It can just be `ProviderName` without `"` when usingi mport with TYPE_CHECKING.

backend/app/config.py
```diff
@@ -169,6 +178,23 @@ def assemble_cors_origins(cls, v: str | list[str]) -> list[str] | str:
...
+        """Build OAuth redirect URI for a provider.
+
+        Uses the legacy per-provider *_REDIRECT_URI env var if set,
```
KaliszS: Why to keep that legacy? I would remove all legacy environment variables, it won't break anything.

backend/app/config.py
```diff
@@ -77,48 +83,51 @@ class Settings(BaseSettings):
     sync_interval_seconds: int = 3600  # Default: 1 hour (3600 seconds)
     sleep_sync_interval_seconds: int = 3600  # Default: 1 hour (3600 seconds)
 
+    # API SETTINGS
+    api_base_url: str = "http://localhost:8000"
```
KaliszS, line 87: I would consider using Request library to fetch apo_base_url from endpoint query. But that would require to propagate that environment from the api router to the provider service. Might not be that easy, but worth to discuss that possibility with agent which can take a look at the code.

bartmichalak: So according to Claude it's technically possible, with something like that:

```
 A middleware captures request.base_url into a context variable. The oauth_redirect_uri method reads it - no call chain changes needed.
                                                                                                                                                                                                                                  
  # new contextvar          
  _request_base_url: ContextVar[str | None] = ContextVar("request_base_url", default=None)                                                                                                                                        
                            
  # middleware sets it                                                                                                                                                                                                            
  @app.middleware("http")   
  async def capture_base_url(request, call_next):                                                                                                                                                                                 
      _request_base_url.set(str(request.base_url).rstrip("/"))
      return await call_next(request)                                                                                                                                                                                             
  
  # config reads it with fallback                                                                                                                                                                                                 
  def oauth_redirect_uri(self, provider: ProviderName) -> str:
      base = _request_base_url.get() or self.api_base_url                                                                                                                                                                         
      return f"{base}/api/v1/oauth/{provider.value}/callback"
```


but considering this comment:
>Both share the same fundamental risk: behind a reverse proxy, request.base_url may return the internal address unless forwarded headers are configured correctly. So either way you'd likely want to keep api_base_url as a fallback and only use the request-derived value if a flag is set or the config value is empty.

So I'll leave it as is for now, this PR reduces the need to set a callback per provider and requires setting it only once, I think it's quite a good improvement :D

backend/app/config.py
bartmichalak: Changed.

PR #681, opened 2026-03-27
feat: implement deregister on user deletion

czajkub:
## Description

When `/delete` endpoint on user is called, first, a disconnect (and deregistration if possible) happens (now it only happens manually, and when user was deleted, the connection in database was removed but access in provider api was still granted

### General

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Users can now delete their accounts with automatic cleanup of external provider connections.

* **Bug Fixes**
  * Improved clarity of connection revocation logging messages.

### DIFF
--- a/backend/app/services/user_service.py
+++ b/backend/app/services/user_service.py
@@ -73,7 +73,7 @@ class UserService(AppService[UserRepository, User, UserCreateInternal, UserUpdat
                 log_structured(
                     self.logger,
                     "warning",
-                    f"Failed to deregister user from provider '{connection.provider}'",
+                    "Failed to deregister user",
                     user_id=user.id,
                     provider=connection.provider,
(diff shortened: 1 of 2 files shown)


backend/app/services/user_service.py
```diff
@@ -53,6 +56,30 @@ def update(
...
+                continue
+            try:
+                strategy = provider_factory.get_provider(connection.provider)
+                if oauth := strategy.oauth:
+                    oauth.deregister_user(connection.access_token)
+            except Exception as e:
+                log_structured(
+                    self.logger,
+                    "warning",
+                    f"Failed to deregister user from provider '{connection.provider}'",
```
KaliszS: There is already information about provider below, so no need to repeat that information here.

PR #690, opened 2026-03-30
feat: save sleep stages from Oura

czajkub:
## Description

sleep stages are now saved coming from Oura

will update full coverage matrix later after verification

### General

^^^^ will update docs in another pr

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Enhanced Oura sleep tracking: added optional 5-minute sleep phase data (deep, light, REM, awake).
  * Phase data is converted into ordered stage intervals and added as precise stage timestamps in normalized sleep records.

### DIFF
--- a/backend/app/services/providers/oura/data_247.py
+++ b/backend/app/services/providers/oura/data_247.py
@@ -8,6 +8,7 @@ from typing import Any
 from uuid import UUID, uuid4
 
 from app.config import settings
+from app.constants.series_types.oura.sleep_phase import SLEEP_PHASE_MAP
 from app.constants.sleep import SleepStageType
 from app.database import DbSession
 from app.models import EventRecord
@@ -142,23 +143,16 @@ class Oura247Data(Base247DataTemplate):
         }
         return self._paginate(db, user_id, "/v2/usercollection/sleep", params)
 
-    SLEEP_PHASE_MAP = {
-        "1": SleepStageType.DEEP,
-        "2": SleepStageType.LIGHT,
-        "3": SleepStageType.REM,
-        "4": SleepStageType.AWAKE,
-    }
-
     def _extract_sleep_stages(self, sleep_phase_5_min: str | None, sleep_start: str | None) -> list[SleepStage]:
         """Convert Oura's 5-minute sleep phase string into list of SleepStage."""
-        if not sleep_phase_5_min or not sleep_start:
+        if not (sleep_phase_5_min and sleep_start):
             return []
 
         stages: list[SleepStage] = []
 
         phase_start = datetime.fromisoformat(sleep_start.replace("Z", "+00:00"))
 
-        for stage, group in groupby(sleep_phase_5_min, lambda x: self.SLEEP_PHASE_MAP.get(x, SleepStageType.UNKNOWN)):
+        for stage, group in groupby(sleep_phase_5_min, lambda x: SLEEP_PHASE_MAP.get(x, SleepStageType.UNKNOWN)):
             occurrences = len(list(group))
             stages.append(
                 SleepStage(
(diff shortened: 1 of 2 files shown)


backend/app/services/providers/oura/data_247.py
```diff
@@ -139,6 +142,33 @@ def get_sleep_data(
...
         return self._paginate(db, user_id, "/v2/usercollection/sleep", params)
 
+    def _extract_sleep_stages(self, sleep_phase_5_min: str | None, sleep_start: str | None) -> list[SleepStage]:
+        """Convert Oura's 5-minute sleep phase string into list of SleepStage."""
+        if not sleep_phase_5_min or not sleep_start:
+            return []
+
+        stages: list[SleepStage] = []
+
+        phase_map = {
```
czajkub: could be moved outside of the method? Not sure how much overhead that would remove, I feel like initialising this dict wouldn't be that costly

czajkub: just moved it

backend/app/services/providers/oura/data_247.py
```diff
@@ -139,6 +142,33 @@ def get_sleep_data(
         }
         return self._paginate(db, user_id, "/v2/usercollection/sleep", params)
 
+    SLEEP_PHASE_MAP = {
```
KaliszS: Isn't that so,ething that should belong to schemas or constants?

backend/app/services/providers/oura/data_247.py
```diff
@@ -139,6 +142,33 @@ def get_sleep_data(
...
+        "1": SleepStageType.DEEP,
+        "2": SleepStageType.LIGHT,
+        "3": SleepStageType.REM,
+        "4": SleepStageType.AWAKE,
+    }
+
+    def _extract_sleep_stages(self, sleep_phase_5_min: str | None, sleep_start: str | None) -> list[SleepStage]:
+        """Convert Oura's 5-minute sleep phase string into list of SleepStage."""
+        if not sleep_phase_5_min or not sleep_start:
```
KaliszS: Maybe that would be more compact?
`if not (sleep_phase_5_min and sleep_start)`

backend/app/services/providers/oura/data_247.py
```diff
@@ -139,6 +142,33 @@ def get_sleep_data(
...
+            return []
+
+        stages: list[SleepStage] = []
+
+        phase_start = datetime.fromisoformat(sleep_start.replace("Z", "+00:00"))
+
+        for stage, group in groupby(sleep_phase_5_min, lambda x: self.SLEEP_PHASE_MAP.get(x, SleepStageType.UNKNOWN)):
```
KaliszS: Interesting choice, but why groupby? Couldn't it be done a little bit simpler?

KaliszS (changes requested):

backend/app/services/providers/oura/data_247.py
KaliszS: @coderabbitai could you note to always check logical conditions? Often they can be rewritten into something easier. Wdyt about this case?

backend/app/services/providers/oura/data_247.py
czajkub: fair point, moved to constants

backend/app/services/providers/oura/data_247.py
czajkub: yeah, feels a bit better

backend/app/services/providers/oura/data_247.py
czajkub: at first i did manual grouping, but felt it was a bit too verbose and maybe un-pythonic (like keeping 2 additional variables for last phase and occurences of last phase, iterating manually, and saving the last one because saves happened only after change of phase), and I think this might be a bit more readable

PR #748, opened 2026-04-01
fix(backend): derive Garmin sleep end_datetime from stage timeline

bartmichalak:
## Summary

Two changes to Garmin sleep data processing:

**1. `end_datetime` derived from stage timeline**

We were computing `end_datetime = start + durationInSeconds`, but Garmin's `durationInSeconds` only counts asleep time (deep + light + rem). Now uses the last stage end from `sleepLevelsMap` instead, so `end_datetime` reflects when the session actually ended.

**2. `sleep_total_duration_minutes` decoupled from `duration_seconds`**

Since `duration_seconds` now covers the full session (including awake), `sleep_total_duration_minutes` is explicitly computed as `deep + light + rem` instead of relying on `duration_seconds // 60`. The output value doesn't change - this just keeps it correct after fix 1.

**Example from real payload (night of March 2):**

| | Before | After |
|---|---|---|
| `end_datetime` | `05:03 UTC` (start + 462min) | `06:09 UTC` (last stage end) |
| `duration_seconds` | `27720` (462min) | `31680` (528min, full session) |
| `sleep_total_duration_minutes` | `462` | `462` (unchanged, asleep only) |

The 66-minute gap was exactly `awakeDurationInSeconds` - consistent across all 30 records in the dataset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of sleep end times and total duration calculations for Garmin devices by utilizing actual sleep stage interval data when available.

* **Tests**
  * Added test coverage for sleep duration calculations with detailed sleep stage data.

### DIFF
--- a/backend/tests/providers/garmin/test_garmin_247.py
+++ b/backend/tests/providers/garmin/test_garmin_247.py
@@ -240,10 +240,10 @@ class TestGarmin247Data:
 
         normalized = garmin_247.normalize_sleep(sleep_data, user_id)
 
-        expected_end = datetime(2024, 1, 14, 22, 0, 0, tzinfo=timezone.utc).timestamp() + 7200 + 1800
-        actual_end = datetime.fromisoformat(normalized["end_time"]).timestamp()
+        actual_end = datetime.fromisoformat(normalized["end_time"])
+        expected_end = datetime.fromtimestamp(awake_end_ts, tz=timezone.utc)
         assert actual_end == expected_end
-        assert normalized["duration_seconds"] == 7200 + 1800  # includes awake
+        assert normalized["duration_seconds"] == 9000  # 7200 asleep + 1800 awake
 
     def test_normalize_sleep_missing_stages(self, garmin_247: Garmin247Data) -> None:
         """Test normalizing sleep with missing stage data."""


backend/tests/providers/garmin/test_garmin_247.py
```diff
@@ -214,6 +214,37 @@ def test_extract_sleep_stages_from_map(self, garmin_247: Garmin247Data) -> None:
...
+                "awake": [{"startTimeInSeconds": start_ts + 7200, "endTimeInSeconds": awake_end_ts}],
+            },
+        }
+
+        normalized = garmin_247.normalize_sleep(sleep_data, user_id)
+
+        expected_end = datetime(2024, 1, 14, 22, 0, 0, tzinfo=timezone.utc).timestamp() + 7200 + 1800
+        actual_end = datetime.fromisoformat(normalized["end_time"]).timestamp()
+        assert actual_end == expected_end
+        assert normalized["duration_seconds"] == 7200 + 1800  # includes awake
```
bartmichalak: Not sure if we want to have this awake value included - I need to dive deeper into our approach for other providers and standardize across them

bartmichalak: OK, here's how it looks (consistent across all providers):

- `sleep_total_duration_minutes` (`EventRecordDetail` model) -> asleep only
- `duration_seconds` (`EventRecord` model) -> asleep + awake

We need to rename these columns so they clearly communicate what they represent, or get rid of one of them entirely - I'm not convinced both are needed.

bartmichalak: @corerabbitai please create an issue describing this problem.

bartmichalak: @coderabbitai please create an issue describing this problem

PR #793, opened 2026-04-07
feat: add health score model

### LINKED ISSUE #762: feat: introduce health score model
We want a dedicated model for health scores - both from providers (Garmin, Oura, Whoop, etc.) and our own computed scores.

**Why a separate model?**

- Easily extensible - if we later want to store algorithm version, confidence, or input completeness, we just add columns
- Supports both numeric and qualitative values - some scores are integers (Garmin overall sleep score: 75), some are strings (Garmin sleep sub-scores: "FAIR"), some are both:

```json
"overallSleepScore": {
  "value": 75,
  "qualifierKey": "FAIR"
},
"sleepScores": {
  "totalDuration": { "qualifierKey": "FAIR" },
  "stress": { "qualifierKey": "FAIR" },
  "awakeCount": { "qualifierKey": "EXCELLENT" },
  "remPercentage": { "qualifierKey": "FAIR" },
  "restlessness": { "qualifierKey": "EXCELLENT" },
  "lightPercentage": { "qualifierKey": "FAIR" },
  "deepPercentage": { "qualifierKey": "FAIR" }
}
```

- Clean separation between raw measurements (`data_point_series`) and composite scores
- Allows storing both provider scores and internally-computed OW scores in one place

**Open question: score range configuration**

Providers use different scales (Garmin sleep: 0-100, Whoop strain: 0-21, qualitative only, etc.). Frontend needs to know the range to display scores properly (e.g. "78/100" vs just "78"). Where do we store this config - in code or in the database? TBD.

czajkub:
## Description

Add a new model for health scores (e.g. sleep score) for scores either from provider API (e.g. Garmin, Whoop) or scores calculated within Open Wearables (planned sleep scores)

JSONB field contains score constituents - could be seen as sub-scores, from which main score is calculated

Resolves #762 

### General

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Added health score tracking with categories: Sleep, Recovery, Readiness, Activity, Stress, Body Battery, and Strain
  * Health scores include per-component values, provider metadata, and API schemas for create/update/response
  * Introduced health score ranges for validation/display

* **Behavior Change**
  * An internal provider type exists but is excluded from provider matching and provider listings

* **Tests**
  * Updated tests to reflect the internal provider exclusion

### DIFF
--- a/backend/app/schemas/enums/health_score_category.py
+++ b/backend/app/schemas/enums/health_score_category.py
@@ -7,6 +7,5 @@ class HealthScoreCategory(StrEnum):
     READINESS = "readiness"
     ACTIVITY = "activity"
     STRESS = "stress"
-    HRV = "hrv"
     BODY_BATTERY = "body_battery"
     STRAIN = "strain"


backend/app/schemas/model_crud/activities/health_score.py
```diff
@@ -0,0 +1,41 @@
...
+    """A single constituent of a health score (e.g. deep sleep percentage)."""
+
+    value: Decimal | None = None
+    qualifier: str | None = None
+
+
+class HealthScoreBase(BaseModel):
+    category: str
+    value: Decimal | None = None
+    qualifier: str | None = None
```
bartmichalak: It would valueable adding Field(description=...) to value and qualifier - from the schema alone it's not clear what these fields represent (e.g. is qualifier a textual rating like "good"/"poor"? what does value score?).

Also, these descriptions end up in the generated OpenAPI spec

backend/app/schemas/model_crud/activities/health_score.py
```diff
@@ -0,0 +1,41 @@
...
+
+class ScoreComponent(BaseModel):
+    """A single constituent of a health score (e.g. deep sleep percentage)."""
+
+    value: Decimal | None = None
+    qualifier: str | None = None
+
+
+class HealthScoreBase(BaseModel):
+    category: str
```
bartmichalak: Shouldn't this be an enum? I think we can easily determine the possible values.

backend/app/models/health_score.py
```diff
@@ -0,0 +1,36 @@
...
+    provider: Mapped[ProviderName | None]
+
+    category: Mapped[str_32]
+    value: Mapped[numeric_5_2 | None]
+    qualifier: Mapped[str_32 | None]
+
+    recorded_at: Mapped[datetime]
+    zone_offset: Mapped[str_10 | None]
+
+    constituents: Mapped[json_binary | None]
```
KaliszS: maybe `components` would be a better name?

backend/app/models/health_score.py
```diff
@@ -0,0 +1,36 @@
...
+            "recorded_at",
+            name="uq_health_score_source_category_time",
+        ),
+    )
+
+    id: Mapped[PrimaryKey[UUID]]
+    data_source_id: Mapped[FKDataSource | None]
+    provider: Mapped[ProviderName | None]
+
+    category: Mapped[str_32]
```
KaliszS: As @bartmichalak this could be an enum. But if you use enum here, you have to also add mapping that enum to String(32) in `database.py`.

backend/app/schemas/model_crud/activities/health_score.py
czajkub: yes, added fields

backend/app/schemas/model_crud/activities/health_score.py
czajkub: good idea, changed to enum

backend/app/models/health_score.py
czajkub: makes sense, a bit easier to understand

backend/app/schemas/model_crud/activities/health_score.py
```diff
@@ -0,0 +1,41 @@
...
+    provider: ProviderName | None = None
+
+
+class HealthScoreUpdate(HealthScoreBase): ...
+
+
+class HealthScoreResponse(HealthScoreBase):
+    id: UUID
+    data_source_id: UUID | None
+    provider: ProviderName | None
```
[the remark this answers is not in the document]
czajkub: yea, added it

backend/migrations/versions/2026_04_07_1515-5643d2210b58_health_score_table.py
```diff
@@ -0,0 +1,44 @@
...
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "health_score",
+        sa.Column("id", sa.UUID(), nullable=False),
+        sa.Column("data_source_id", sa.UUID(), nullable=True),
+        sa.Column("provider", sa.String(length=50), nullable=True),
```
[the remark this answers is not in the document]
czajkub: data source includes provider info, so its still kept, and for data from providers the data source will be non-null each time

PR #820, opened 2026-04-09
fix(frontend): set NODE_ENV=production in Docker build

bartmichalak:
## Summary

- Fixes `jsxDevRuntimeExports.jsxDEV is not a function` SSR crash on production
- Adds `ENV NODE_ENV=production` in both builder and runtime stages of the frontend Dockerfile

Caused by the TanStack Start bump in #809 (1.167.8 -> 1.167.16) which started respecting `NODE_ENV` when choosing JSX runtime for SSR bundles. Since `.env` has `NODE_ENV=development`, the production Docker build was importing `jsxDEV` from `react/jsx-dev-runtime` instead of `jsx` from `react/jsx-runtime`.

## Test plan

- [x] Rebuild frontend Docker image and verify no SSR errors in logs
- [x] Verify pages render correctly via browser

## Summary by CodeRabbit

* **Chores**
  * Updated build and runtime configuration for improved deployment consistency.

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -19,7 +19,8 @@ RUN pnpm install --frozen-lockfile
 # Copy source files
 COPY . .
 
-# Build the application
+# Build the application in production mode
+ENV NODE_ENV=production
 RUN pnpm run build
 
 # Determine build output directory and prepare it for copying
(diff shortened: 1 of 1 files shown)


bartmichalak:
Yeah, it works. 

Well done, Claude.

PR #826, opened 2026-04-11
feat: add celery tasks for calculating daily sleep scores

czajkub:
## Description

Add daily celery task for calculating sleep scores that runs a query collecting backfill days' worth of sleeps without a healht score, then dispattches internal sleep score calculation in bulk

### General

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Automated sleep health score backfilling now runs on a periodic schedule to compute and save missing nightly scores (defaults to a 30-day lookback).
  * Backfill interval is configurable (defaults to 600 seconds); tasks report saved vs skipped counts for visibility.


backend/app/integrations/celery/tasks/fill_missing_sleep_scores_task.py
```diff
@@ -0,0 +1,137 @@
...
+from app.utils.structured_logging import log_structured
+from celery import shared_task
+
+logger = getLogger(__name__)
+
+# Find all non-nap sleep sessions that have no corresponding OW sleep score.
+# The match is on user + calendar date (UTC) of start_datetime vs recorded_at.
+# Limited to sessions within the configured backfill window to avoid scanning
+# the entire history on every run.
+_MISSING_SCORES_QUERY = text("""
```
bartmichalak, line 24: One potential edge case just came to my mind.

If the user has multiple data sources for sleep (for example: Apple Watch and some ring) there will be two sleep sessions in the database and this query will catch whatever comes first - what's even worse, we'll have no easy way to check what data the sleep score was calculated from.

To handle that I'd consider adding an association column to the HealthScore model that would store a reference to the `SleepDetails` record and then taking it into account in this query.

So when the user has multiple data sources for sleep we'll calculate a separate score for each of them (they will differ).

bartmichalak (approved):
I think we can merge the current implementation, but I'd love to see the edge case mentioned in my comment addressed in a separate PR.

Solid job! 🚀

backend/app/integrations/celery/tasks/fill_missing_sleep_scores_task.py
FBegiello: The problem here is whether we should calculate separate score for each, or follow the device prioritisation and assume the more accurate one?
In my mind, as a data unification/normalisation platform, we should aim for uniformity here, and calculate a signle score, based on a decission - either arbitrally pick according to the priority, pick a longer one, or something in this vein.

czajkub: For now, if there are multiple sleep events, the longest one is counted for the sleep score, but adding a `SleepDetails` reference or using device/provider priority is a good direction for the future

PR #853, opened 2026-04-14
feat(backend): add created_at to BaseDbModel for all tables

bartmichalak:
## Description

Moves `created_at` from individual model declarations into `BaseDbModel` with `server_default=func.now()`, so every table gets it automatically. No more forgetting to add it to new models.

> **Warning:** The migration sets `created_at = NOW()` (migration time) for all existing rows in the 12 tables that didn't have the column before. This needs confirmation/discussion before running on production - we might want a different backfill strategy.

### General

- [x] New and existing tests pass locally (1580/1580)

### Backend Changes

## Testing Instructions

**Steps to test:**
1. Run migration: `alembic upgrade head`
2. Verify `created_at` column exists on all tables
3. Create new records (e.g. health scores) and verify `created_at` is auto-populated

**Expected behavior:**
Every table has a `created_at` column with `server_default=now()`. New records get the timestamp automatically without explicit assignment.

## Additional Notes

- 10 models already had `created_at` declared individually - removed in favor of inheritance from `BaseDbModel`, plus `server_default` added
- 12 tables get the new column via migration: `archival_settings`, `data_point_series`, `data_point_series_archive`, `data_source`, `event_record`, `event_record_detail`, `health_score`, `personal_record`, `provider_settings`, `series_type_definition`, `sleep_details`, `workout_details`
- Existing repo code that sets `created_at` explicitly still works (explicit values override server defaults)

## Summary by CodeRabbit

* **Refactor**
  * Centralized creation timestamp handling so all models now use a single, consistent server-managed created_at field.
* **Chores**
  * Added a migration to apply and normalize automatic creation timestamps across the database, ensuring new records receive server-side creation times.

### DIFF
--- a/backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
+++ b/backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
@@ -1,7 +1,7 @@
 """base model created_at
 
 Revision ID: 4bd01c907050
-Revises: cdac07b15b04
+Revises: e4f7a2c8b390
 
 """
 
@@ -12,7 +12,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision: str = "4bd01c907050"
-down_revision: Union[str, None] = "cdac07b15b04"
+down_revision: Union[str, None] = "e4f7a2c8b390"
 branch_labels: Union[str, Sequence[str], None] = None
 depends_on: Union[str, Sequence[str], None] = None
 
--- a/backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
+++ b/backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
@@ -30,7 +30,9 @@ TABLES_WITH_CREATED_AT = [
     "user_invitation_code",
 ]
 
-# Tables that need the column added
+# Tables that need the column added.
+# sleep_details and workout_details are polymorphic joined-table children of
+# event_record_detail; created_at lives on the parent table, not on the children.
 TABLES_WITHOUT_CREATED_AT = [
     "archival_settings",
     "data_point_series",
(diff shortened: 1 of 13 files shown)


backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def upgrade() -> None:
+    for table in TABLES_WITHOUT_CREATED_AT:
+        op.add_column(
+            table,
+            sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        )
+
+    for table in TABLES_WITH_CREATED_AT:
```
KaliszS, line 67: Isn't it doubled?
Also is it generates by alembic? Looks weird.

KaliszS (changes requested):

backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
bartmichalak: What kind of duplication do you mean?
There are two constants:

- TABLES_WITH_CREATED_AT
- TABLES_WITHOUT_CREATED_AT

And it's intentional. For tables that already have created_at, we now set NOW() at the db level, so we don't need to set it in Python code. 

Also, it wasn't generated by alembic, I don't think it's possible to create such a migration automatically? 🤔

backend/migrations/versions/2026_04_24_1500-4bd01c907050_base_model_created_at.py
```diff
@@ -0,0 +1,78 @@
...
+    "workout_details",
+]
+
+
+def upgrade() -> None:
+    for table in TABLES_WITHOUT_CREATED_AT:
+        op.add_column(
+            table,
+            sa.Column("created_at", sa.DateTime(timezone=True), server_default=sa.func.now(), nullable=False),
+        )
```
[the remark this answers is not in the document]
bartmichalak: @KaliszS I changed the approach a little bit. Instead of backfilling all existing records with now(), I set `EPOCH_SENTINEL = sa.text("'1970-01-01 00:00:00+00'")` to clearly distinguish which values were added after the records were created originally. WDYT? It's not perfect, but at the same time I don't want to leave this field nullable.

KaliszS (approved):

Issue #859, opened 2026-04-15
bug(workout_types): Health Connect exercise vocabulary falls through to 'other'


kjsteuer:
## Description

Every workout ingested from the Android SDK → HealthKit path gets persisted as `event_record.type = 'other'` when its underlying HC `ExerciseSessionRecord.exerciseType` isn't one of the ~14 Apple HealthKit strings hardcoded in `SDKWorkoutType`. Peloton indoor rides, treadmill runs, rowing machine sessions, strength training, HIIT, boot camp, yoga classes, etc. all collapse to `OTHER`.

The Android SDK's `HealthConnectManager.mapExerciseType` emits uppercase HC constants like `BIKING_STATIONARY`, `RUNNING_TREADMILL`, `ROWING_MACHINE`, `EXERCISE_CLASS`, `STRENGTH_TRAINING`, `HIIT`, `BOOT_CAMP`, `FOOTBALL_AMERICAN`, `FRISBEE_DISC`, etc. The backend's `SDKWorkoutType` enum in `backend/app/constants/workout_types/apple_sdk.py:6-118` only has the Apple HealthKit snake_case vocabulary (`cycling`, `running`, `walking`, …), so `SDK_TO_UNIFIED.get(sdk_activity_type, WorkoutType.OTHER)` at line 255 silently swallows everything HC-originated.

## Steps to Reproduce

1. Connect an Android device with Peloton (or any HC-writing fitness app) installed
2. Complete an indoor cycling workout in the writing app
3. Let the OW Android SDK sync to the backend
4. `SELECT type, COUNT(*) FROM event_record WHERE category='workout' GROUP BY type;`

## Expected Behavior

`BIKING_STATIONARY` → `WorkoutType.INDOOR_CYCLING`, `RUNNING_TREADMILL` → `TREADMILL`, `ROWING_MACHINE` → `ROWING`, etc. Each HC exercise type should map to the closest existing `WorkoutType`.

## Actual Behavior

11 of 21 workouts in my test set come out as `'other'`. The real distribution is ~half indoor cycling (Peloton), ~half walking, one hiking, but the dashboard shows mostly "other".

## Root Cause

`SDKWorkoutType` is Apple-HK-only; the HC vocabulary was never added. `get_unified_workout_type` also does case-sensitive lookup, so even if a string like `"cycling"` came through the pipeline it might not match a canonical snake_case entry.

## Proposed Fix

Two changes:

[...]

PR #865, opened 2026-04-15
feat(backend): persist workout title and notes from SDK payloads

### LINKED ISSUE #860: feat(event_record): persist workout title and notes from SDK payloads
## Problem Statement

HC `ExerciseSessionRecord` carries `title` (e.g. "45 min Power Zone Endurance Ride - Matt Wilpers") and `notes` (free-form description). Third-party writers like Peloton, Strava, and Zwift populate these with rich, human-readable context that users care about seeing on their dashboard.

The wire schema already accepts them (`SyncRequest.WorkoutJson` at `backend/app/schemas/providers/mobile_sdk/sync_request.py:54`) and the Android SDK's `UnifiedWorkout` schema already sends them. But the backend drops both fields during ingestion:

1. `event_record` table has no `title` or `notes` columns.
2. `EventRecordCreate` schema has no matching fields.
3. `ImportService._build_workout_bundles` (`backend/app/services/apple/healthkit/import_service.py:89-104`) never passes `wjson.title` / `wjson.notes` into `EventRecordCreate(...)`.

Net result: the user sees a 45-minute "indoor cycling" workout with no indication that it was a Peloton Power Zone Endurance Ride with Matt Wilpers.

## Proposed Solution

1. Alembic migration adding two nullable columns to `event_record`:
   - `title VARCHAR(255)`
   - `notes VARCHAR(2000)`
2. Update `EventRecord` ORM model.
3. Update `EventRecordBase` / `EventRecordCreate` schemas.
4. Forward `wjson.title` / `wjson.notes` in `ImportService._build_workout_bundles`.
5. Update `EventRecordRepository.bulk_create` values_list tuple so the bulk upsert path writes the new columns (the single-record path picks them up via `model_dump`).

I have a working implementation verified end-to-end against a seeded Peloton ride, and will open a PR against this issue.

## Use Case

Any OW user with a third-party HC writer (Peloton, Strava, Zwift, Komoot, Nike Training Club, Garmin Connect via HC, …) gets the full workout context on their dashboard and in any downstream MCP / agent / analysis tools. This is especially valuable for coached workouts where the title is the whole point.

## Alternatives Considered

Using a side-channel metadata JSON column — rejected because it would require schema-level handling everywhere `title`/`notes` are read, and because the wire format already has typed fields for these exactly.

## Environment

- Backend: `the-momentum/open-wearables` main
- Android SDK: `the-momentum/open_wearables_android_sdk` v0.9.0

## Additional Context

One of four defects in the Android SDK → HealthKit ingestion path that collectively cause rich Peloton ride data to arrive at the backend as a generic "other" workout with no metadata. Filed separately for independent triage.

kjsteuer:
## Description

Fixes #860.

`EventRecord` has no columns for the `title` and `notes` fields that the SDK already ships in its workout payloads, so `wjson.title` (e.g. "30 min Power Zone Ride") and `wjson.notes` (user-entered ride comments) are discarded on the way in. This PR persists both through the existing Apple / HealthKit ingestion path.

## What changed

1. **Alembic migration** `7ef3a1b2c4d5_event_record_title_notes` — adds nullable `title` and `notes` columns to `event_record`, chained from the current head `cdac07b15b04`.
2. **ORM model** `backend/app/models/event_record.py` — matching `Optional[str]` columns on `EventRecord`.
3. **Schemas** `backend/app/schemas/model_crud/activities/event_record.py` — `title` / `notes` added to `EventRecordBase` and `EventRecordCreate`.
4. **Import service** `backend/app/services/apple/healthkit/import_service.py` — forwards `wjson.title` / `wjson.notes` into `EventRecordCreate` on both the Apple HealthKit and SDK import paths.
5. **Repository** `backend/app/repositories/event_record_repository.py` — explicitly maps the two new columns in `bulk_create`'s `values_list` tuple so the bulk upsert path actually writes them. (The single-record `create` path already uses `model_dump` and picks them up for free.)

### General

- [x] Tests added — N/A for the ORM / migration; integration coverage of the import path is unchanged and still passes
- [x] Existing tests pass locally
- [x] Migration chains from the current head `cdac07b15b04`
- [ ] Docs update not needed

### Backend Changes

## Testing Instructions

**Steps to test:**
1. Run `alembic upgrade head` — verify the new columns appear on `event_record`.
2. Send a SDK sync payload containing `workouts[].title` and `workouts[].notes` (e.g. Peloton emits both per ride).
3. Confirm the resulting `event_record` rows now carry the title and notes values instead of dropping them.
4. Hit the dashboard activities list — titles now render instead of generic "Workout".

**Expected behavior:**

- `title` / `notes` fields flow from SDK payload → `EventRecordCreate` → Postgres
- Existing rows without title/notes are unaffected (columns are nullable)
- Both single-create and bulk-create code paths persist the fields

## Additional Notes

Useful in conjunction with #859 (HC exercise vocabulary) — a HC "BIKING" session with a meaningful title like "30 min Power Zone Ride" is much more useful than a generic cycling entry. Neither change depends on the other; they can merge in any order.

## Summary by CodeRabbit

* **New Features**
  * Added workout title (up to 255 chars) and notes (up to 2000 chars). These fields are captured during HealthKit imports and shown with workout records to provide display names and free-form descriptions.
* **Migration**
  * Database schema updated to store the new title and notes columns; imports and bulk create now populate them.


backend/app/models/event_record.py
```diff
@@ -30,6 +33,14 @@ class EventRecord(BaseDbModel):
     type: Mapped[str_32 | None]
     source_name: Mapped[str_64]
 
+    # Human-readable metadata forwarded from third-party providers via the
+    # Apple HealthKit and Health Connect SDKs. ``title`` is the workout
+    # display name (e.g. ``"45 min Power Zone Endurance Ride"``); ``notes``
+    # is a longer free-form description. Both are optional because many
+    # providers don't populate them.
+    title: Mapped[str_255 | None]
+    notes: Mapped[str_2000 | None]
```
KaliszS, line 42: Why 2000 as a limit?

KaliszS (changes requested):

backend/app/models/event_record.py
kjsteuer: 2000 is only generous bound rather than a hard requirement.

notes from third-party HC writers (Peloton, Strava, Zwift) is typically a few hundred chars at most — users rarely write long paragraphs. I picked 2000 to leave headroom for edge cases without going unbounded.

Alternatives considered:
- TEXT — cleanest if you'd rather avoid an arbitrary cliff, but app.mappings caps at varchar-bounded types today so introducing str_text would be a small departure from convention.
- str_1000 — tighter; almost certainly enough in practice.

Happy to shrink to 1000, switch to TEXT, or keep 2000 — whichever fits your convention best.

KaliszS:
Hi. We have decided to not support this feature at the moment.

PR #882, opened 2026-04-16
feat(mcp): add get_timeseries tool for granular sample queries

### LINKED ISSUE #881: feat(mcp): expose /timeseries endpoint as a
  get_timeseries tool
## Problem Statement

The MCP server currently exposes four tools — `get_users`, `get_activity_summary`, `get_sleep_summary`, `get_workout_events` — all of which are built on the daily-summary endpoints. None of them reach the `/api/v1/users/{user_id}/timeseries` endpoint.

That leaves an LLM client unable to answer questions about any metric that does not have a dedicated summary endpoint: weight, SpO2, HRV (SDNN / RMSSD), blood glucose, respiratory rate, blood pressure, body fat %, intraday heart rate, etc. All of this data is already ingested and queryable via the REST API; it's just not reachable from AI clients today.

Concrete example: asking Claude Desktop "What are my weight trends over the last 30 days?" against a live backend with 19 weight samples in `data_point_series` produces no useful answer, because the model has no tool to read the series.

## Proposed Solution

Add a single `get_timeseries` MCP tool that wraps the existing `/timeseries` endpoint, plus a matching `OpenWearablesClient.get_timeseries` method.

Shape mirrors the other tools:

```python
async def get_timeseries(
    user_id: str,
    start_time: str,   # ISO-8601
    end_time: str,     # ISO-8601
    types: list[str],  # SeriesType codes
    resolution: str = "raw",  # "raw" | "1min" | "5min" | "15min" | "1hour"
) -> dict
```

Returns `{user, period, records, summary, truncated}` — `records` is the flat sample list (`timestamp`, `type`, `value`, `unit`, `source`), `summary` is per-type aggregates (`count`, `avg`, `min`, `max`), and `truncated` flags rare cases where cursor pagination hits a safety ceiling (100 pages / 10k samples).

The docstring should explicitly steer LLMs toward `get_activity_summary` when a daily rollup already exists, and toward `get_timeseries` only for metrics without a dedicated summary or when the user needs intraday detail.

## Use Case

Any AI-client user whose questions touch metrics outside the four daily summaries — weight tracking, SpO2 trends, HRV analysis, intraday heart-rate zones, blood glucose curves, etc. A natural-language prompt like "How has my weight changed this month?" becomes answerable without any new backend work, since `/timeseries` already handles it.

## Alternatives Considered

- **Per-metric tools** (`get_weight`, `get_spo2`, `get_hrv`, ...): clean surface but N× the code and a new PR every time a new `SeriesType` ships. Rejected in favor of one pass-through tool.
- **Client-side aggregation inside the MCP server** (e.g. a `get_daily_average` tool that runs the summary math in the tool layer): conflates transport with analysis. Left to each client to do with the raw samples.
- **New `/timeseries/summary` backend endpoint**: larger scope, schema design work, and the backend already serves the raw data. Not needed to close the LLM-accessibility gap.

## Additional Context

I have a working implementation ready to PR (~235 lines, all additive under `mcp/`) — happy to open it against this issue. Smoke-tested locally against a real backend:

- 1,433 `heart_rate` samples for a single day, pagination walks cleanly, `truncated: false`
- 19 `weight` samples across 30 days with mixed `kg` units from Withings, aggregates correct
- Multi-type queries (`heart_rate` + `oxygen_saturation`) return the intersection with the expected per-type split

No new third-party dependencies. Ruff and format checks pass. I can add a `tests/` directory for the MCP package as a follow-up if you'd like — there's no test coverage there today, so it's a separate conversation about framework/layout.

kjsteuer:
Closes #881

## Summary

Adds a `get_timeseries` MCP tool that exposes `/api/v1/users/{user_id}/timeseries` to AI clients, alongside the matching `OpenWearablesClient.get_timeseries` method.

The existing tools cover daily summaries (activity / sleep / workouts) but there is no path into the raw sample stream, so metrics that do not have a dedicated summary — weight, SpO2, HRV, blood glucose, intraday heart rate, etc. — are invisible to an LLM client today.

## What it does

- **Client method** (`mcp/app/services/api_client.py`): async `get_timeseries(user_id, start_time, end_time, types, resolution="raw", limit=100, cursor=None)`. Mirrors the backend signature from `backend/app/api/routes/v1/timeseries.py` and returns the raw paginated payload.
- **MCP tool** (`mcp/app/tools/timeseries.py`): walks cursor pagination (ceiling at 100 pages / 10k samples, `truncated: true` flag when hit), returns `{user, period, records, summary, truncated}`. `summary` contains per-type aggregates (`count`, `avg`, `min`, `max`).
- **Router mount** (`mcp/app/main.py`): registered via `mcp.mount(timeseries_router)` and listed in the instructions block.
- **README** updated with the new tool in Features + the `tools/` layout.

## Design notes

- `resolution` is passed through as a `str` so the tool surface stays decoupled from the backend enum. The backend rejects anything outside `{"raw", "1min", "5min", "15min", "1hour"}` and surfaces a 422, which the existing `_request` path already translates into a `ValueError`.
- The "Notes for LLMs" section of the docstring explicitly steers the model toward `get_activity_summary` when a daily rollup is sufficient, and toward `get_timeseries` only for metrics without a dedicated summary or when the user needs intraday detail.
- Values are returned in the unit the provider supplied (`bpm`, `kg`, `%`, …); unit conversions stay client-side to match the existing convention.

## Testing

Smoke-tested end-to-end against a live backend with real wearable data:

| Scenario | Result |
|---|---|
| `heart_rate`, 1-day window, `resolution="1min"` | 1,433 samples, avg 97.57 bpm, pagination walked cleanly (`truncated: false`) |
| `weight`, 30-day window, `resolution="1hour"` | 19 samples across providers, avg 88.12 kg, sparse pagination handled correctly |
| Multi-type: `heart_rate` + `oxygen_saturation`, 1-day | Per-type split in `summary.by_type`; empty SpO2 confirmed as real data gap via direct curl |

Tooling checks:

- `uv run --frozen python -c "from app.tools.timeseries import timeseries_router; from app.main import mcp"` — imports cleanly, server initializes.
- `uv run --frozen --group code-quality ruff check app/` — all checks passed.
- `uv run --frozen --group code-quality ruff format --check app/` — 13 files already formatted.

No unit tests added: the `mcp/` package does not ship a `tests/` directory today and none of the existing tools (`users.py`, `activity.py`, `sleep.py`, `workouts.py`) have coverage. Happy to bolt on a test suite in a follow-up PR — let me know the preferred framework/layout.

## Checklist

- [x] Conventional commit title (scope `mcp`)
- [x] No new third-party dependencies
- [x] Backwards compatible (new tool, no changes to existing signatures)
- [x] `ruff check` + `ruff format --check` pass

## Summary by CodeRabbit

* **New Features**
  * Added time-series health metrics retrieval to query granular data (weight, SpO2, HRV, intraday heart rate) across flexible time periods.
  * Includes automatic pagination to fetch large datasets and a truncation indicator when results hit a hard cap.
  * Produces per-metric statistical summaries (count, average, minimum, maximum) and returns structured results with user and period context.
  * Returns clear error responses for user-lookup or fetch failures.

### DIFF
--- a/mcp/README.md
+++ b/mcp/README.md
@@ -8,6 +8,7 @@ MCP (Model Context Protocol) server for Open Wearables, enabling AI assistants l
 - **get_activity_summary**: Get daily activity data (steps, calories, heart rate, intensity minutes)
 - **get_sleep_summary**: Get sleep data for a user within a date range
 - **get_workout_events**: Get workout/exercise sessions for a user within a date range
+- **get_timeseries**: Get granular time-series samples (weight, SpO2, HRV, intraday heart rate, etc.)
 
 ## Prerequisites
 
@@ -188,7 +189,8 @@ mcp/
 │   │   ├── users.py      # get_users tool
 │   │   ├── activity.py   # get_activity_summary tool
 │   │   ├── sleep.py      # get_sleep_summary tool
-│   │   └── workouts.py   # get_workout_events tool
+│   │   ├── workouts.py   # get_workout_events tool
+│   │   └── timeseries.py # get_timeseries tool
 │   └── services/
 │       └── api_client.py # HTTP client for backend API
 ├── config/
--- a/mcp/app/main.py
+++ b/mcp/app/main.py
@@ -9,6 +9,7 @@ from app.config import settings
 from app.prompts import prompts_router
 from app.tools.activity import activity_router
 from app.tools.sleep import sleep_router
+from app.tools.timeseries import timeseries_router
 from app.tools.users import users_router
 from app.tools.workouts import workouts_router
 
@@ -35,6 +36,7 @@ mcp = FastMCP(
     - get_activity_summary: Get daily activity data (steps, calories, heart rate, intensity minutes)
     - get_sleep_summary: Get sleep data for a user over a specified time period
     - get_workout_events: Get workout/exercise data for a user over a specified time period
+    - get_timeseries: Get granular time-series samples (e.g. weight, SpO2, HRV, intraday heart rate)
 
     Available prompts:
     - present_health_data: Guidelines for formatting health data for human readability
@@ -115,6 +117,7 @@ mcp.mount(users_router)
 mcp.mount(activity_router)
 mcp.mount(sleep_router)
 mcp.mount(workouts_router)
+mcp.mount(timeseries_router)
 
 # Mount prompts
 mcp.mount(prompts_router)
--- a/mcp/app/services/api_client.py
+++ b/mcp/app/services/api_client.py
@@ -167,6 +167,42 @@ class OpenWearablesClient:
         }
         return await self._request("GET", f"/api/v1/users/{user_id}/summaries/activity", params=params)
 
+    async def get_timeseries(
+        self,
+        user_id: str,
+        start_time: str,
+        end_time: str,
+        types: list[str],
+        resolution: str = "raw",
+        limit: int = 100,
+        cursor: str | None = None,
+    ) -> dict[str, Any]:
+        """
+        Get granular time-series samples for a user within a time range.
+
+        Args:
+            user_id: UUID of the user
+            start_time: Start of the window (ISO-8601, e.g. "2026-04-05T00:00:00Z")
+            end_time: End of the window (ISO-8601, e.g. "2026-04-05T23:59:59Z")
+            types: SeriesType codes to include (e.g. ["heart_rate", "oxygen_saturation"])
+            resolution: One of "raw", "1min", "5min", "15min", "1hour"
+            limit: Page size (1-100)
+            cursor: Opaque pagination cursor returned by a previous call
+
+        Returns:
+            Paginated response with time-series samples
+        """
+        params: dict[str, Any] = {
+            "start_time": start_time,
+            "end_time": end_time,
+            "types": types,
+            "resolution": resolution,
+            "limit": limit,
+        }
+        if cursor:
+            params["cursor"] = cursor
+        return await self._request("GET", f"/api/v1/users/{user_id}/timeseries", params=params)
+
 
 # Singleton instance
 client = OpenWearablesClient()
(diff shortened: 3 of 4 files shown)


kjsteuer:
Thanks for the review! Just pushed `d5b5faf` addressing the feedback.

**Nit #2 (`timeseries.py:132` — redundant `or []`)**: agreed, dropped the fallback. `response.get("data", [])` already defaults cleanly when the key is absent.

**Nit #3 (`timeseries.py:27` — `dict` vs `dict[str, Any]`)**: intentionally matching the tool-layer convention — `activity.py`, `sleep.py`, `workouts.py`, and `users.py` all use bare `dict` as the return annotation. The `dict[str, Any]` form is the client-layer convention in `api_client.py`. Happy to change it if maintainers prefer unifying on the more specific form across both layers.

**Actionable comment on line 165 (numeric check)**: replied inline — the backend's `TimeSeriesSample` schema pins `value: float | int`, so Pydantic rejects non-numeric values at the API boundary before they reach this loop. I'd rather trust that contract than silently skip records if the schema ever changed.

kjsteuer:
One more round of CodeRabbit feedback — new comment on `timeseries.py:117` about the `except ValueError` block collapsing auth/404/config errors into "User not found".

The observation is accurate, but the block is a direct mirror of the existing pattern in `activity.py:105-108`:

```python
except ValueError as e:
    return {"error": f"User not found: {user_id}", "details": str(e)}
```

Same shape in `sleep.py` and `workouts.py`. I kept this one consistent with the neighbors intentionally. If maintainers agree with CodeRabbit that the pattern should be more discriminating, happy to open a follow-up PR applying a uniform fix across all four tool files rather than diverging just this one. Let me know the preference.

KaliszS (approved):
Thank you @kjsteuer, good catch!

I agree with coderabbit that we can improve error handling a little bit, so you can create a new PR for that :)

PR #892, opened 2026-04-16
fix(backend): align health scores and workout fields with real provider formats

FBegiello:
## Description

Aligns the seed data service with real provider API formats so seed data is indistinguishable from live data when querying or building features against it.

**Health scores** — adds provider-specific daily health scores for the three providers that emit them:
- **Oura**: `SLEEP`, `READINESS`, `ACTIVITY` with correct `contributors` dict keys (`deep_sleep`, `efficiency`, `rem_sleep`, `activity_balance`, `hrv_balance`, `meet_daily_targets`, etc.)
- **Garmin**: `SLEEP` (with `EXCELLENT`/`GOOD`/`FAIR`/`POOR` qualifier + `deepSleep`/`remSleep`/`restlessness`/`sleepDuration`/`sleepInterruption` components), `BODY_BATTERY`, and `STRESS` — qualifiers stored in normalized form matching the real parser (`"Low Stress"`, not `"LOW_STRESS"`)
- **WHOOP**: `SLEEP` and `RECOVERY` with metric keys matching the real parser (`hrv_rmssd_milli`, `spo2_percentage`, `skin_temp_celsius`, `resting_heart_rate`, `sleep_consistency_percentage`, etc.)

**Workout fields** — provider-specific fields now populated: `energy_burned` (Garmin/Apple/Polar/Suunto), `total_elevation_gain` (outdoor workout types only), `average_speed` (Garmin, pace-based workouts only).

**Bug fixes**:
- `heart_rate_min`/`heart_rate_avg` were being passed to `EventRecordDetailCreate` for sleep records. These columns exist only in `WorkoutDetails`, causing a hard `TypeError` on every sleep creation.
- Garmin stress qualifiers were stored in raw API format (`LOW_STRESS`) rather than the normalized form the real parser produces (`.replace("_", " ").title()` → `Low Stress`), making seed and live data inconsistent.

### General

### Backend Changes

### Frontend Changes

N/A

## Testing Instructions

**Steps to test:**
1. Hit `POST /api/v1/settings/seed` with a default `SeedDataRequest` (or run `python scripts/init/seed_activity_data.py` directly)
2. Query `GET /api/v1/health-scores?user_id=<seeded_user>` and verify Oura, Garmin, and WHOOP scores appear with populated `components`
3. Confirm Garmin stress scores have qualifier `"Low Stress"` / `"Medium Stress"` / `"High Stress"` (not `LOW_STRESS`)
4. Confirm sleep records are created without errors and `sleep_details` rows are present

**Expected behavior:**

Seed completes without errors and produces health scores whose `provider`, `category`, `qualifier`, and `components` fields match what the real provider parsers would write for the same data.

## Additional Notes

`_GARMIN_STRESS_QUALIFIERS` (module-level constant) was a dead reference — the inline computation was using its own hardcoded strings. Both are now consistent. The constant is kept for documentation but the inline strings are the source of truth.

## Summary by CodeRabbit

* **Refactor**
  * Reorganized internal seed data generation service into a modular architecture with specialized components for improved maintainability.
  * Restructured configuration and helper utilities into dedicated modules for better code organization.


backend/app/services/seed_data_service.py
czajkub: Oura doesnt pass any device info via API, so if we want to reproduce real data we'd have to remove these somewhere, but its only a minor thing so not that important

backend/app/services/seed_data_service.py
```diff
@@ -585,19 +825,36 @@ def generate(self, db: Session, request: SeedDataRequest) -> dict:
...
                     event_record_service.create(db, record)
                     event_detail_repo.create(db, detail, detail_type="sleep")
                     summary["sleeps"] += 1
 
+            # Health scores — one batch per provider covering the full seeded date range
+            if profile.generate_workouts or profile.generate_sleep:
+                date_range_months = max(
+                    profile.workout_config.date_range_months if profile.generate_workouts else 0,
+                    profile.sleep_config.date_range_months if profile.generate_sleep else 0,
+                ) or 6
```
czajkub: btw, will this fallback ever be hit? or will it be `0` if both options here are missing

czajkub (approved):
looks good to me 😄 , i think there were only some formatting changes / pr title check failing

backend/app/services/seed_data_service.py
czajkub: nevermind, it got outdated before i submitted the review :P

backend/app/services/seed_data_service.py
FBegiello: Fair point - modified it so that everything follows the real data

KaliszS (approved):

PR #900, opened 2026-04-16
fix: align sleep score date to wake-up date via direct record linkage

FBegiello:
## Description

Sleep scores were stored with `recorded_at = midnight(local start_date)`, but the dashboard shows sessions by their **wake-up date** (`end_datetime`). For an overnight session (e.g. Mon 23:00 → Tue 07:00) this caused the score to appear a day before the session it belongs to.

Additionally, users in extreme time zones could have two sessions with the same local `end_date`, causing a date-keyed approach to silently drop one score.

Fix: add a `sleep_record_id` FK to `health_score` so each internal sleep score is linked directly to its `event_record`. `recorded_at` is set to midnight of the local **wake date** (`end_datetime + zone_offset`). Sleep summaries are also updated to group by `end_datetime` for full alignment.

Changes:
- `HealthScore` model: add nullable `sleep_record_id` FK + partial unique index (`UNIQUE(sleep_record_id) WHERE sleep_record_id IS NOT NULL`)
- Migration `a3f8c1d2e4b5`: add column/FK/index, clear stale internal scores so they regenerate on the correct date
- `SleepScoreService`: new `get_sleep_scores_for_records` method keyed by `(record_id, wake_date)`
- `fill_missing_sleep_scores` task: match on `sleep_record_id` (idempotent), derive `wake_date` from `end_datetime`
- Sleep summaries: `local_sleep_date` grouped by `end_datetime` instead of `start_datetime`
- Calculation logic (score algorithm, historical bedtimes, stage parsing) is **unchanged**

### General

### Backend Changes

### Frontend Changes

- N/A

## Testing Instructions

**Steps to test:**
1. Apply migration `a3f8c1d2e4b5` — existing internal sleep scores are cleared, new column + partial unique index are created
2. Trigger `fill_missing_sleep_scores` (Celery task) — confirm new scores have `sleep_record_id` populated and `recorded_at = midnight(end_date local)`
3. On the dashboard, open an overnight session (starts before midnight, ends the next morning) — confirm the session and its score both appear on the **wake-up date**
4. For a user with two sessions ending on the same calendar day, confirm both get distinct scores (no collision / silent drop)

**Expected behavior:**
- Sleep score date matches the session wake-up date on the dashboard
- Each score has a non-null `sleep_record_id` pointing to its source record
- No duplicate or missing scores; idempotent on re-run

## Additional Notes

The `ty` check reports one pre-existing `unused-ignore-comment` warning in `event_record_service.py` (not touched in this PR). All 36 sleep score tests pass.

## Release Notes

* **New Features**
  * Sleep scores are now directly linked to individual sleep sessions for improved accuracy and consistency tracking.

* **Bug Fixes**
  * Sleep date calculations now use wake-up time instead of session start time for more accurate sleep period identification.

* **Documentation**
  * Added guides explaining how to run data migration scripts, including instructions for backfilling missing sleep scores and managing one-off database corrections.


backend/app/repositories/health_score_repository.py
```diff
@@ -51,11 +51,7 @@ def bulk_create(self, db_session: DbSession, creators: list[HealthScoreCreate])
 
         values = [c.model_dump() for c in creators]
 
-        stmt = (
-            insert(HealthScore)
-            .values(values)
-            .on_conflict_do_nothing(index_elements=["user_id", "provider", "category", "recorded_at"])
-        )
+        stmt = insert(HealthScore).values(values).on_conflict_do_nothing()
```
czajkub, line 54: Hmm, after these changes, I wonder if this is gonna cause any issues. Now, `recorded_at` will be shifted to the right by 1 day (usually), but the scores for this date also might exist (unique index `("user_id", "provider", "category", "recorded_at")`). Also, even if the date stays the same, we will always do nothing on conflict, while the sleep record id will remain null for that score, so i'm thinking we could update the values instead of skipping unique index violations

backend/app/integrations/celery/tasks/fill_missing_sleep_scores_task.py
```diff
@@ -18,28 +18,33 @@
...
-        (er.start_datetime + COALESCE(er.zone_offset, '+00:00')::interval)::date AS sleep_date
+        er.id AS record_id,
+        (er.end_datetime + COALESCE(er.zone_offset, '+00:00')::interval)::date AS wake_date,
+        (er.end_datetime + COALESCE(er.zone_offset, '+00:00')::interval) AS local_end_datetime
     FROM event_record er
     JOIN data_source ds   ON ds.id = er.data_source_id
     JOIN sleep_details sd ON sd.record_id = er.id
     LEFT JOIN health_score hs
-           ON hs.user_id     = ds.user_id
+           ON hs.sleep_record_id = er.id
```
czajkub, line 39: Does this mean we'll be now calculating sleep scores for each session (for one day multiple providers might give a sleep session for one user) instead of one score per day?

backend/migrations/versions/2026_04_16_1200-a3f8c1d2e4b5_clear_internal_sleep_scores.py
czajkub, line 1: I'd regenerate this migration using alembic just in case to prevent any weirdness that might arise later

backend/app/repositories/health_score_repository.py
czajkub: So considering this constraint will be hit, we would need to update:
- `value`, `qualifier` (might not be the same session -> shift by one day to the future)
- `components` (same case)
- `sleep_record_id` - would be null for scores already calculated by us
- `data_source_id` i think is not filled now - but if we're going by sleep session, every record will have a data source id connected to it - we can pin it there now as well

But, since recorded_at is now not just the date but entire timestamp - it's very unlikely that a user will wake up at exactly `00:00`

backend/app/integrations/celery/tasks/fill_missing_sleep_scores_task.py
```diff
@@ -99,25 +108,26 @@ def fill_missing_sleep_scores() -> dict:
                     provider=ProviderName.INTERNAL,
                     category=HealthScoreCategory.SLEEP,
                     value=result.overall_score,
-                    recorded_at=datetime(d.year, d.month, d.day, tzinfo=timezone.utc),
+                    sleep_record_id=record_id,
```
czajkub, line 113: here, we can also add `data_source_id` - data source is already joined with the query, so shouldn't have any performance overhead

czajkub (changes requested):

backend/app/integrations/celery/tasks/fill_missing_sleep_scores_task.py
FBegiello: I had in mind staying with one score per day - the ID is pointing to direct sleep data used as per priority. We are still calculating once, but we are tying the score with the driect data used as a main source for this score to get a matching start/end times.
In addition, this is not conflicting with later changing priority/recalculating for other providers, as long as we retrieve the correct matching score

backend/app/repositories/health_score_repository.py
FBegiello: Good point - we can polish this up, so that we have a migration script resolving any possible conflicts and taking into account older versions as we discussed.

czajkub (approved):

Issue #916, opened 2026-04-17
Refactor: backend ingestion under apple/ actually serves all SDK providers


kjsteuer:
## Background

The OW Android SDK translates Health Connect constants to HealthKit-compatible strings **before** the backend `/data/sync` endpoint sees them. This gives the backend a single unified wire vocabulary to ingest from any SDK-using provider (apple, google, samsung), which is a good property — and was confirmed in a CodeRabbit/maintainer exchange on #864: https://github.com/the-momentum/open-wearables/pull/864#discussion_r3083891636

However, the backend file and module layout doesn't reflect this. Paths and module names under `backend/app/services/apple/` and `backend/app/constants/.../apple_sdk.py` read as Apple-only, when in reality they are the shared post-SDK-translation pipeline for every provider using the SDK route.

This creates two concrete problems:

## Problem 1 — Apple-specific unit assumptions silently bleed into Android/Samsung

The SDK does *name* translation but not *unit* translation. Unit conversions that are correct for HealthKit (e.g., `HKUnit.percent()` returns a 0-1 ratio) are applied as-is to Health Connect payloads where the same field is already percent-scale — inflating Android/Samsung data by 100x.

Concrete instance: `body_fat_percentage` was being stored around `3040` instead of `30.4` for every `provider="google"` user. The narrow fix is in PR **#917** (this issue should be filed before that PR — I'll link it once opened). That fix is symptomatic, not structural: any other place in `apple/healthkit/` that bakes in HealthKit unit semantics is a latent bug for non-Apple providers.

## Problem 2 — The naming misleads readers

Someone adding an Android-only feature or fixing an Android-specific bug would reasonably assume `backend/app/services/apple/healthkit/import_service.py` is scoped to Apple. Nothing in the path or module name signals that it also handles Samsung and Google. Grepping for the right code is harder; the mental model for new contributors is wrong by default.

## Suggested directions

Listed least-to-most invasive:

1. **Rename for accuracy.** Move `apple/healthkit/` ingestion paths to something provider-neutral (e.g., `sdk/ingest/`), and `apple_sdk.py` to `sdk_wire_vocabulary.py` (or similar). Same behavior, names match reality.
[...]

PR #863, opened 2026-04-15
fix(backend): fall back to source.app_id / bundle_identifier for original_source_name

kjsteuer:
## Description

Fixes #858.

`resolve_device_info` in `backend/app/services/apple/healthkit/device_resolution.py` returns `original_source_name = None` when the incoming SDK payload does not provide a friendly `source.name` — even when `source.app_id` or `source.bundle_identifier` are present. Android SDK payloads consistently only supply `app_id` (for HC writers like Peloton this is `com.onepeloton.callisto`), so `source_name` lands in Postgres as NULL and the device / source analytics go dark.

This change falls back to `source.app_id` → `source.bundle_identifier` → `None` when `source.name` is absent, so third-party HC writers are identifiable end-to-end.

### General

- [x] My code follows the project's code style
- [x] I have performed a self-review of my code
- [x] Tests added — N/A (purely a field-fallback; no logic change)
- [x] `uv run pre-commit run --all-files` passes locally (from the `backend/` directory)
- [ ] Docs update not needed

### Backend Changes
- [x] `uv run pre-commit run --all-files` passes

## Testing Instructions

**Steps to test:**
1. Send a SDK sync payload whose `source` object contains only `app_id` (e.g. `"com.onepeloton.callisto"`) — no `name`.
2. Verify the resulting row in `data_point_series` / `event_record` / `device_*` has `original_source_name = "com.onepeloton.callisto"` instead of `NULL`.
3. Confirm payloads that already provide `source.name` still use that value (regression check).

**Expected behavior:**

- `source.name` present → used as-is (unchanged)
- `source.name` absent + `source.app_id` present → use `app_id`
- `source.name` absent + `source.bundle_identifier` present → use `bundle_identifier`
- Nothing present → `None` (unchanged)

## Additional Notes

One of several companion fixes to get Peloton (and more generally any HC-only writer) data through the Apple/HealthKit ingestion path cleanly. The SDK side does not need a matching change — it was already forwarding `app_id` correctly; only the backend was dropping it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced device source identification in Apple Health data imports by implementing additional fallback mechanisms, ensuring better device recognition when standard identifiers are unavailable.

PR #534, opened 2026-02-27
feat(backend): adapt SDK import services to new payload

czajkub:
## Description

Adjusted import services (namely Apple SDK on iOS and Samsung Health / Google Health Connect on Android) to accept the new, unified payload coming from Open Wearables mobile app.
- removed Samsung import task
- changed the Apple import task to accept new SDK payload
- add new metric types / workout statistics 

### General

- [x] My code follows the project's code style
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix/feature works (if applicable)
- [x] New and existing tests pass locally

### Backend Changes

You have to be in `backend` directory to make it work:
- [x] `uv run pre-commit run --all-files` passes

## Additional Notes

Example payload is updated in  `backend/app/schemas/apple/healthkit/sync_request.py` (it should not be called `apple/healthkit` anymore, but refactor is underway

## Summary by CodeRabbit

* **New Features**
  * Unified SDK-based sync across multiple providers (apple, samsung, google, auto-health-export).
  * Sync requests now include provider, sdkVersion, and syncTimestamp; richer device/workout metadata and new device/record enums.
  * Added Google Health Connect support and provider-aware import flow.
  * Expanded metrics: body fat mass, skeletal muscle mass, hydration; more workout statistics.

* **Chores / Infrastructure**
  * Background processing consolidated to an SDK sync queue.

* **Tests**
  * Test fixtures and payloads updated to the new SDK envelope and provider model.

PR #917, opened 2026-04-17
fix(backend): scope x100 ratio->percent conversion to Apple provider only

kjsteuer:
## Summary

`ImportService` in `backend/app/services/apple/healthkit/import_service.py` is routed for Apple, Samsung, and Google SDK payloads (see `process_sdk_upload_task._get_import_service`), but `body_fat_percentage` was unconditionally multiplied by `100`. That ratio->percent conversion is correct only for Apple HealthKit, where `HKUnit.percent()` returns a `0..1` ratio. Android Health Connect's `BodyFatRecord.percentage.value` is already reported in percent, so Google/Samsung body-fat samples were stored ~100x too large.

## Why

On Android users, stored `body_fat_percentage` samples come back around `3040` (from an actual value like `30.4`). This made downstream summaries and charts meaningless for any provider using the SDK path on Android. Samsung is presumably also affected — Samsung's Health Connect export uses the same semantics — though I don't personally have Samsung data to confirm directly.

Height (`SeriesType.height`) is left unchanged: both HealthKit and Health Connect report height in meters, so the m -> cm conversion is correct for all providers.

## Changes

- `backend/app/services/apple/healthkit/import_service.py`: gate the `x100` for `body_fat_percentage` on `provider == "apple"`. Height conversion remains provider-agnostic.
- `backend/tests/integrations/test_apple_sdk_import.py`: new `TestSDKImportUnitConversion` class covering the provider-dispatched branches.

This is intentionally a minimal, targeted fix. A proper provider-dispatch refactor for ingestion (splitting the HealthKit service from a Health Connect service, or introducing an explicit provider adapter layer) is a larger change that this PR unblocks rather than attempts.

## Question for maintainers

This fix lands in `apple/healthkit/import_service.py` because that's where the bug lives today — but given that `ImportService` is actually the shared post-SDK-translation path for Apple, Samsung, and Google payloads, it's not obvious this is the right long-term home. Would you prefer:

- **(a)** keep the provider branch here, consistent with the rest of the file; or
- **(b)** start splitting Android/HC-specific handling into a separate module now, with this PR as the first move?

Happy to rework if you have a layout in mind. I've also opened **#916** proposing a broader refactor of this area; this PR is the narrow regression fix, but the two are tied.

## Test plan

- [x] `pytest tests/integrations/test_apple_sdk_import.py` — all 15 tests pass (10 existing + 5 new).
- [x] `pytest -k "body_fat or height or sdk_import or apple"` — 52 tests pass.
- [x] `ruff check` and `ruff format --check` on both modified files — clean.

New test cases:
- `provider="apple"`, `body_fat_percentage` input `0.304` -> stored `30.4`
- `provider="google"`, `body_fat_percentage` input `30.4` -> stored `30.4`
- `provider="samsung"`, `body_fat_percentage` input `18.5` -> stored `18.5`
- `provider="apple"`, `height` input `1.7526` -> stored `175.26`
- `provider="google"`, `height` input `1.7526` -> stored `175.26` (regression guard)

## Notes

No data backfill is included in this PR — existing incorrect rows will need a separate one-off correction, which is infra-specific to each deployment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected unit scaling for body fat percentage in health data imports. Apple provider measurements are now scaled by 100, while data from Google and Samsung providers remain unscaled.

* **Tests**
  * Added comprehensive test coverage for health metric unit conversions across different health data providers.

### DIFF
--- a/backend/app/services/apple/healthkit/import_service.py
+++ b/backend/app/services/apple/healthkit/import_service.py
@@ -127,8 +127,14 @@ class ImportService:
 
             if not series_type:
                 continue
-            # Convert from meters to centimeters or ratio to percentage
-            if series_type in (SeriesType.height, SeriesType.body_fat_percentage):
+            # Convert meters -> centimeters for height (both HealthKit and Health Connect report meters)
+            # and ratio (0..1) -> percent for Apple body_fat_percentage (HealthKit HKUnit.percent()).
+            # Android Health Connect's BodyFatRecord.percentage is already in percent, so only scale
+            # body_fat_percentage for provider == "apple" — otherwise Google/Samsung values are stored
+            # ~100x too large.
+            if series_type == SeriesType.height or (
+                series_type == SeriesType.body_fat_percentage and provider == "apple"
+            ):
                 value = value * 100
 
             # Extract device info
--- a/backend/tests/integrations/test_apple_sdk_import.py
+++ b/backend/tests/integrations/test_apple_sdk_import.py
@@ -8,11 +8,14 @@ import logging
 from decimal import Decimal
 from typing import Any
 from unittest.mock import MagicMock, patch
+from uuid import uuid4
 
 import pytest
 from sqlalchemy.orm import Session
 
 from app.models import EventRecord, WorkoutDetails
+from app.schemas.enums import SeriesType
+from app.schemas.providers.mobile_sdk import SyncRequest as SDKSyncRequest
 from app.services.apple.healthkit.import_service import ImportService
 from tests.factories import UserFactory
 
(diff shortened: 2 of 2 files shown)


kjsteuer:
> _Nitpick: Existing over-scaled body-fat rows for google/samsung won't be corrected._

Thanks — agreed the historical rows are real, but I'd like to push back on adding the backfill to this PR.

The SDK path for `google` / `samsung` is relatively new: provider routing through `ImportService` came in #534 (2026-02-27), and the Health Connect vocabulary fixes that make most Android payloads actually ingestible only landed last week (#863, #864, #866 — all merged 2026-04-12). Before those, a lot of Android data was silently dropped or misclassified, so the realistic affected rowset on any deployment is narrow and heavily deployment-specific.

A universal `UPDATE WHERE value > 1` migration has the downside that it fires indiscriminately on every deployment — including ones with no affected data, or with other sources (Oura, Withings, etc.) whose values might happen to also be > 1 for unrelated reasons. I'd prefer to leave the correction as an operator-side recipe and ship a SQL snippet in the PR notes or a follow-up docs PR if useful.

That said — if maintainers would rather see an Alembic migration here, happy to add one guarded on provider. Or I can file a follow-up issue so it doesn't get lost.

czajkub (approved):
Good catch, we didn't notice this so far. For now this solution is fine, but maybe in the future we will think about another degree of separation (although the goal was to unify payloads between apple and google health connect)

czajkub:
Also, as for the data already there - we shouldn't use migrations for these types of things, we have some data migration scripts in `backend/scripts/data_migrations` for this purpose

kjsteuer:
> Also, as for the data already there - we shouldn't use migrations for these types of things, we have some data migration scripts in `backend/scripts/data_migrations` for this purpose

Hi [czajkub](https://github.com/czajkub), would you like me to like me to explore a migration here , or for this issue we should skip and were just mentioning for scenarios where we actually need a migration?

KaliszS:
> > Also, as for the data already there - we shouldn't use migrations for these types of things, we have some data migration scripts in `backend/scripts/data_migrations` for this purpose
> 
> Hi [czajkub](https://github.com/czajkub), would you like me to like me to explore a migration here , or for this issue we should skip and were just mentioning for scenarios where we actually need a migration?

Hi. We wanted to close this topic before 0.5.1 release, so @czajkub already imlemented and tested this script: https://github.com/the-momentum/open-wearables/pull/1016

PR #928, opened 2026-04-18
feat(polar): daily activity + Sleep Plus Stages + continuous HR sync

mathebrito:
## Summary

Closes the three unsupported Polar AccessLink data types — daily activity, Sleep Plus Stages, and continuous heart rate — in `PolarData247Template`. The Polar Loop (and any Polar wearable) now syncs end-to-end through the existing `POST /api/v1/providers/polar/users/{user_id}/sync?data_type=247` route.

Also fixes two latent defects in the shared persistence paths surfaced by running the full 247 pipeline for the first time (see commit `e59b782` and `6e50029`).

### What's new

| Data type | Endpoint | Persistence |
|---|---|---|
| Daily activity (Phase 1, pre-existing) | `GET /v3/users/activities/{date}` | `EventRecord(category="daily_activity")` + workout-shaped detail |
| **Sleep Plus Stages** | `GET /v3/users/sleep` | `EventRecord(category="sleep")` + `SleepDetails` with `sleep_stages` JSONB built from Polar's hypnogram (0-5 → unified `SleepStageType`, 2/3 both collapse to `LIGHT` since the unified enum has no N1/N2 split) |
| **Continuous heart rate** | `GET /v3/users/continuous-heart-rate/{date}` | `DataPointSeries(series_type=heart_rate)` via `timeseries_service.bulk_create_samples` + explicit commit |

Each data-type block in `load_and_save_all` runs in its own try/except and calls `db.rollback()` on failure so a crash in one block doesn't poison the FastAPI request-scoped session for the other two. Nightly Recharge is still stubbed (documented as future work).

### Bugs fixed in passing

1. **`save_daily_activity_statistics` was not idempotent on re-sync.** The caller generated a fresh `uuid4` per item, called `event_record_service.create(...)`, threw away the returned record (which would correctly reconcile with the existing row after the unique-index rollback), and tried to insert a detail keyed off the discarded uuid → `ForeignKeyViolation`. Now uses `persisted = event_record_service.create(...).id` for the detail. This was invisible in the Phase 1 smoke test because that ran against an empty DB.

2. **Continuous-HR samples were staged but never committed.** `timeseries_service.bulk_create_samples` only issues `INSERT ... ON CONFLICT DO UPDATE` via `db.execute`; the `/sync` route does not commit on return. `save_continuous_hr` now commits explicitly — tested with a `MagicMock` session so regressing this fails the unit test instead of silently dropping rows. (The same pattern appears in `suunto/data_247.py::save_activity_samples`; not fixed here to keep the PR focused, but likely needs the same treatment.)

### Tests

16 new unit tests (`tests/providers/polar/`):
- 8 sleep — normalization, hypnogram → stage-interval construction with cross-midnight handling, empty-hypnogram fallback, missing-start-time skip, list-unwrap, 204 tolerance
- 6 HR — sample building with UTC-naive date+time combine, malformed-sample skip, day-by-day fetch iteration, commit-is-called regression, empty-list no-op
- 2 session-isolation regression — assert that a DB error in daily_activity does not prevent sleep/HR from running (and vice versa)

49/49 polar tests pass; no regressions in existing `test_polar_oauth.py`, `test_polar_strategy.py`, `test_polar_workouts.py`.

### Docs

`docs/providers/polar-api-integration.mdx` support table: workouts + daily activity + sleep + continuous HR flip to "Yes"; Nightly Recharge stays flagged as upcoming.

## Test plan

- [ ] `pytest tests/providers/polar/ -q` — expect 49 passed
- [ ] OAuth an active Polar device (Loop or other), then `POST /api/v1/providers/polar/users/{uuid}/sync?data_type=247&async=false` — response should show non-zero `daily_activity_synced`, `sleep_sessions_synced`, `continuous_hr_synced` for a user with recent data
- [ ] Verify persistence: `SELECT category, count(*) FROM event_record er JOIN data_source ds ON er.data_source_id = ds.id WHERE ds.provider='polar' GROUP BY category;` → both `daily_activity` and `sleep` present
- [ ] Verify stages populated: `SELECT jsonb_array_length(sleep_stages) FROM sleep_details sd JOIN event_record er ON er.id = sd.record_id JOIN data_source ds ON er.data_source_id = ds.id WHERE ds.provider='polar' ORDER BY er.start_datetime DESC LIMIT 5;` → non-zero counts
- [ ] Verify HR samples: `SELECT count(*) FROM data_point_series dps JOIN data_source ds ON dps.data_source_id = ds.id JOIN series_type_definition std ON dps.series_type_definition_id = std.id WHERE ds.provider='polar' AND std.code='heart_rate';` → non-zero
- [ ] Re-sync twice and confirm no duplicate `event_record` rows and no FK violations in logs (regression for the Phase 1 idempotency fix)

## Notes for reviewers

- The hypnogram → stage-interval builder anchors the first transition to `sleep_start_time`'s calendar date and bumps subsequent transitions forward one day whenever they would regress (for nights that cross midnight). Only tolerant to within-24h sleep sessions, which matches the Polar API contract (one `sleep` object = one night).
- Continuous-HR `sample_time` comes with no timezone; we combine `date + HH:MM:SS` as UTC-naive to match the existing policy in `_parse_polar_datetime`. Downstream consumers convert to the user's tz.
- `hasattr(self.event_record_repo, "get_by_external_id")` guard that was in Phase 1 was dead code — the method never existed — so removing it has no behavioral impact beyond the bug it was masking.

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end Polar sync: daily activity, sleep with stage-level hypnograms, continuous heart rate samples, and nightly recharge metrics.
  * Added new series types for Polar-specific scores and charge/status metrics.

* **Documentation**
  * Updated Polar integration docs to document the newly supported data types and persisted fields.

* **Tests**
  * Added comprehensive tests covering Polar sync, normalization, saving, and DB-session isolation.

* **Chores**
  * Updated local volume bindings used by compose for development.

### DIFF
--- a/backend/app/constants/series_types/polar/__init__.py
+++ b/backend/app/constants/series_types/polar/__init__.py
@@ -1,0 +1,3 @@
+from app.constants.series_types.polar.sleep_stage import POLAR_HYPNOGRAM_MAP
+
+__all__ = ["POLAR_HYPNOGRAM_MAP"]
--- a/backend/app/constants/series_types/polar/sleep_stage.py
+++ b/backend/app/constants/series_types/polar/sleep_stage.py
@@ -1,0 +1,26 @@
+"""Polar Sleep Plus Stages hypnogram code → unified SleepStageType.
+
+Polar's AccessLink v3 `sleep.hypnogram` is a mapping of ``"HH:MM"`` transition
+times to stage codes. Codes are documented as::
+
+    0 = WAKE
+    1 = REM
+    2 = LIGHTER NON-REM   (AASM N1)
+    3 = LIGHT NON-REM     (AASM N2)
+    4 = DEEP NON-REM      (AASM N3)
+    5 = UNKNOWN           (e.g. poor skin contact)
+
+OW's `SleepStageType` has a single ``LIGHT`` bucket, so 2 and 3 both map to it.
+Codes outside this range fall back to ``UNKNOWN`` at the call site.
+"""
+
+from app.constants.sleep import SleepStageType
+
+POLAR_HYPNOGRAM_MAP: dict[int, SleepStageType] = {
+    0: SleepStageType.AWAKE,
+    1: SleepStageType.REM,
+    2: SleepStageType.LIGHT,
+    3: SleepStageType.LIGHT,
+    4: SleepStageType.DEEP,
+    5: SleepStageType.UNKNOWN,
+}
--- a/backend/app/schemas/enums/series_types.py
+++ b/backend/app/schemas/enums/series_types.py
@@ -150,6 +150,13 @@ class SeriesType(str, Enum):
     garmin_body_battery = "garmin_body_battery"  # Garmin body battery (0-100)
 
     # =========================================================================
+    # POLAR-SPECIFIC METRICS (IDs 240-259)
+    # =========================================================================
+    polar_nightly_recharge_status = "polar_nightly_recharge_status"  # 1-5 ordinal
+    polar_ans_charge = "polar_ans_charge"                             # -10 to +10 signed
+    polar_ans_charge_status = "polar_ans_charge_status"               # 1-5 ordinal
+
+    # =========================================================================
     # OTHER (IDs 500-)
     # =========================================================================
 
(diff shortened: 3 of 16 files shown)


bartmichalak (changes requested):
@mathebrito please refer to the CodeRabbit comments, then someone from the team will take a look!

KaliszS:
> @mathebrito please refer to the CodeRabbit comments, then someone from the team will take a look!

+1 Also CI failed, so those errors also has to be corrected first.

bartmichalak:
@mathebrito please let us know if you're still planning work on this.

mathebrito:
Hi guys, pretty busy this week but will work on it if you could wait till next week.

czajkub:
hi, we started refactoring Polar ourselves, soon each non-supported (so far) data type will be included alongside webhook support

PR #936, opened 2026-04-20
fix: three startup crashes introduced in 0.4.3

slavarosin:
## Summary

Three bugs introduced in 0.4.3 that prevent the server from starting in production (`uv sync --no-dev`):

- **Circular import #1** — `garmin_webhook_task.py` imports `GarminStrategy` at module level, which creates a cycle: `factory` → `garmin/strategy` → `celery/tasks/__init__` → `garmin_webhook_task` → `garmin/strategy`. Fixed by deferring the import inside the function body.

- **Circular import #2** — `garmin/strategy.py` imports `start_full_backfill` from `garmin_backfill_task` at module level, creating: `factory` → `garmin/strategy` → `celery/tasks/__init__` → `process_aws_upload_task` → `user_service` → `factory`. Fixed by deferring the import inside `start_historical_sync`.

- **Missing prod dependency** — `seed_data_service` and related modules import `faker`, but `faker` is declared as a dev-only dependency. Production Docker builds use `uv sync --no-dev`, causing `ModuleNotFoundError: No module named 'faker'` on startup. Fixed by moving `faker` to the main `dependencies` list.

## Test plan

- [ ] `python -c "from app.services.providers.factory import ProviderFactory"` exits with code 0
- [ ] `python -c "from app.main import app"` exits with code 0
- [ ] Docker build with `uv sync --no-dev` starts successfully

## Summary by CodeRabbit

* **Chores**
  * Improved module initialization and import management
  * Updated project dependencies to include the faker library as a runtime component

### DIFF
--- a/backend/app/integrations/celery/tasks/garmin_webhook_task.py
+++ b/backend/app/integrations/celery/tasks/garmin_webhook_task.py
@@ -11,7 +11,6 @@ from logging import getLogger
 from typing import Any, cast
 
 from app.database import SessionLocal
-from app.services.providers.garmin.strategy import GarminStrategy
 from app.services.providers.garmin.webhook_handler import GarminWebhookHandler
 from app.utils.structured_logging import log_structured
 from celery import Task, shared_task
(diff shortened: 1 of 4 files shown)


KaliszS:
Closing because Claude made up those circular import "errors". Faker issue also doesn't exist probably, but in fact it should have been moved to prod deps, so I will do it in separate PR.

PR #949, opened 2026-04-23
fix: add hrv rmssd to body summary

### LINKED ISSUE #934: Body summary endpoint returns null HRV for
  users with RMSSD-only sources (Oura/HC/Apple)
## Problem

The `/api/v1/summaries/body` endpoint always reports `avg_hrv_sdnn_ms: null` for users whose HRV source only ingests `heart_rate_variability_rmssd`. This is common for users on Oura / Apple Health / Health Connect pipelines — those sources publish RMSSD, not SDNN.

In my local DB (OW 0.4.3):

```
code                         | rows   | last recorded
-----------------------------+--------+---------------
heart_rate_variability_rmssd | 40,323 | 2026-04-15
heart_rate_variability_sdnn  |      0 | —
```

Portal Body tab shows `HRV: -` despite the 7-day window containing tens of thousands of RMSSD samples.

## Root cause

`backend/app/services/summaries_service.py`:

```python
# L82
BODY_AVERAGED_SERIES = [
    SeriesType.resting_heart_rate,
    SeriesType.heart_rate_variability_sdnn,   # RMSSD not queried
]

# L717
hrv_data = vitals_aggregates.get(SeriesType.heart_rate_variability_sdnn)
```

`resilience_service.py:210` already does the right thing on the other HRV-consuming endpoint — queries `rmssd_type_id` first and falls back to `sdnn_type_id`. The Body summary endpoint just wasn't written with that same fallback.

## Related

Not a duplicate of #281 but adjacent: #281 is an **ingest-side** issue (Suunto's RMSSD being stored under the SDNN series type). This issue is on the **query side** — even when RMSSD is correctly ingested, the Body summary endpoint doesn't read it.

## Proposed fix

Mirror the resilience-service pattern: include both series types in `BODY_AVERAGED_SERIES`, and select RMSSD → fallback to SDNN when building the `BodyAveraged` response.

Schema-wise, two options — happy to PR either, would like guidance on direction:

1. **Keep `avg_hrv_sdnn_ms` field, populate from either source.** Minimum churn, no portal changes needed. Downside: the field name becomes misleading since the value can be RMSSD.
2. **Add `avg_hrv_rmssd_ms` alongside `avg_hrv_sdnn_ms`.** Correct and self-documenting, but a breaking schema change for any frontend consumer (portal + MCP body tool, once #575 lands).

My lean is (2) — it's clearer in the API, and since the portal is in the same repo you can coordinate both sides of the change in one PR. But (1) is faster if you'd prefer to ship a fix now and revisit the schema later.

## Reproduction

1. Ingest RMSSD-only HRV data (Oura via Health Connect, Apple Watch via HealthKit, Whoop, etc.).
2. `GET /api/v1/summaries/body`.
3. Observe `avg_hrv_sdnn_ms: null` despite:

```sql
SELECT COUNT(*) FROM data_point_series dps
JOIN series_type_definition std ON dps.series_type_definition_id = std.id
WHERE std.code = 'heart_rate_variability_rmssd';
```

returning non-zero with recent `recorded_at`.

czajkub:
## Description

Add RMSSD HRV to body summaries (now it only queries SDNN). Also, add it to frontend schema and display RMSSD as a fallback when SDNN data is null to keep it consistent with current state of frontedn.

Resolves #934 

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes

## Summary by CodeRabbit

* **New Features**
  * Heart rate variability tracking extended to include an additional averaged HRV metric (RMSSD) alongside the existing SDNN metric.
  * HRV vitals card now selects whichever averaged HRV metric is available and updates its label to indicate which measurement is shown.

* **Documentation**
  * API/guide examples updated to include the new averaged HRV metric in averaged body summaries.

### DIFF
--- a/backend/app/schemas/responses/activity/summaries.py
+++ b/backend/app/schemas/responses/activity/summaries.py
@@ -105,6 +105,7 @@ class BodyAveraged(BaseModel):
     period_days: int = Field(..., description="Number of days averaged (1 or 7)", example=7)
     resting_heart_rate_bpm: int | None = Field(None, description="Average resting heart rate", example=62)
     avg_hrv_sdnn_ms: float | None = Field(None, description="Average HRV (SDNN)", example=45.2)
+    avg_hrv_rmssd_ms: float | None = Field(None, description="Average HRV (RMSSD)", example=42.0)
     period_start: datetime = Field(..., description="Start of averaging period")
     period_end: datetime = Field(..., description="End of averaging period")
 
--- a/backend/app/services/summaries_service.py
+++ b/backend/app/services/summaries_service.py
@@ -80,6 +80,7 @@ BODY_SLOW_CHANGING_SERIES = [
 BODY_AVERAGED_SERIES = [
     SeriesType.resting_heart_rate,
     SeriesType.heart_rate_variability_sdnn,
+    SeriesType.heart_rate_variability_rmssd,
 ]
 
 # Default settings for body summary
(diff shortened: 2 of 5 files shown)


KaliszS (changes requested):
Don't we need any docs update?

czajkub:
Yes, added it in documentation regarding body summary endpoint

KaliszS (approved):

farce1 (approved):

PR #950, opened 2026-04-23
refactor: migrate all recovery scores from `data_point_series` to `health_score`

### LINKED ISSUE #799: chore: remove health-score-related series types from timeseries
## Context

Follow-up to #793 (health score model).

## Scope

Remove series types from `SeriesType` enum that are now covered by the `HealthScore` model:

- `recovery_score` (ID 6)
- `garmin_stress_level` (ID 220)
- `garmin_body_battery` (ID 223)
- any other score-related types that were migrated

## Prerequisite

**Blocked until all providers (Oura, Whoop, Garmin) have been migrated to the new HealthScore model.** Only then is it safe to drop these series types and their associated timeseries data.

czajkub:
## Description

- remove Oura and Whoop saving recovery scores as timeseries alongside health score
- save Suunto recovery to health score
- remove notifications from webhooks about creation
- add migration to remove stale timeseries recovery scores (not used anywhere)

Resolves #799 

### General

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Suunto now saves recovery as a health-score record (instead of timeseries samples).

* **Removals**
  * Recovery-score timeseries type, aggregation mapping, and granular webhook event were removed.
  * Oura and Whoop no longer emit recovery-score timeseries samples; example webhook payload removed.

* **Chores**
  * Added a data-migration and startup step to delete legacy recovery-score timeseries and definitions.

* **Tests**
  * Minor test adjustment to ensure a defined start time when verifying sleep-duration behavior.


KaliszS (changes requested):
I'm not sure this use case fits to migrations. They are supposed to be changes in data structure, not in data intself. If we need to modify production data, it should have been made by one-time script.

Anyway, what's the purspoe of this change? Is there any issue about that?

czajkub:
Forgot to add that it resolves #799. Migration might be too much to be fair, I'll change it to a script, but the idea was to not save any recovery scores to the timeseries table and refactor everything to use the `HealthScore` model. The recovery score timeseries entry was not used anywhere, at least on our frontend

backend/scripts/data_migrations/drop_recovery_score_series_type.py
KaliszS, line 1: 1) We need to ensure this won't end up with exception if someone will have a dtaabse without recovery score types already.
2) Then we want to add this to startup script so it will update that table for all users with existing recovery score series types. We should add there a comment that it will be removed in the future (for example in a month or, so so we will be sure that everyone already dropped those types)

KaliszS (changes requested):

KaliszS (approved):

PR #985, opened 2026-04-28
fix(backend): keep after_commit webhook listener safe across rollback

artursmolin:
## Problem
\`sync_vendor_data\` crashed during a historical sync (and any live sync that happened to hit a duplicate) with:

\`\`\`
sqlalchemy.exc.InvalidRequestError: This session is in 'committed' state;
no further SQL can be emitted within this transaction.
\`\`\`

Two interacting issues caused it.

### 1. \`expire_on_commit\` mismatch on the sync session

\`backend/app/database.py\` configures both a sync and an async sessionmaker. The async one passes \`expire_on_commit=False\`; the sync one didn't. Most of the codebase implicitly assumes attributes survive \`commit()\` — \`sync_vendor_data_task.py:197\` reads \`connection.last_synced_at\` after the workouts commit, for example — and the sync default of \`True\` was the outlier.

### 2. \`after_commit\` listener closed over live ORM instances

\`EventRecordService.create_detail()\` and \`bulk_create_details()\` register an \`after_commit\` listener that captures \`EventRecord\` and \`DataSource\` ORM instances and reads \`record.category\` etc. inside the listener body.

\`event_record_repository.create()\` catches \`IntegrityError\` on duplicate inserts and calls \`db_session.rollback()\`. Historical syncs hit many duplicates (Whoop \`/v2/activity/workout\` returns the full window each call), so they roll back often.

\`Session.rollback()\` **expires every attached instance unconditionally**, regardless of \`expire_on_commit\`. The next commit then fires the listener, which lazy-loads \`record.category\` against a session that is already in 'committed' state — and crashes.

Live syncs escaped this because their cursor-based window contained no duplicates.

## Fix
**\`backend/app/database.py\`** — pass \`expire_on_commit=False\` to the sync sessionmaker, matching the async one. One-line behavioural alignment.

**\`backend/app/services/event_record_service.py\`** — snapshot the handful of fields the webhook needs into frozen \`@dataclass\` value objects *before* registering the listener. The listener now closes over plain-Python data, so it is immune to both attribute expiry on commit and the wholesale expiry that follows a rollback.

Pydantic is reserved for \`schemas/\` in this codebase; internal value objects in \`services/\` already use \`@dataclass\` (see \`services/providers/base_strategy.py:33\`), so the snapshot types follow that convention.

## Why both changes are kept
The snapshot change alone is sufficient to fix the listener crash. The sessionmaker change is kept because:
- it removes the second crash site at \`sync_vendor_data_task.py:197\` (reading \`connection.last_synced_at\` post-commit), and
- it brings the sync session in line with the async one, with no other call site relying on the auto-expire-on-commit behaviour (verified — repos that need a fresh row after commit already call \`db_session.refresh()\` explicitly; no model uses \`server_default\` / \`onupdate\` that would require reload).

## Test plan
- [ ] Trigger a historical sync (\`is_historical=True\`, 30-day window) for a Whoop user with existing workouts in the DB. Expect: task completes successfully, no \`InvalidRequestError\` in the worker log.
- [ ] Trigger a regular live sync. Expect: live path still works, webhook fires once per new record.
- [ ] Confirm late-arriving Whoop sleep (typical 2–12 h delay) lands in \`event_record\` after the next sync that overlaps the actual sleep window.
- [ ] Spot-check that webhooks are dispatched with the same payload as before (snapshot fields are byte-for-byte identical to the previous ORM reads).

## Risk
Low. Webhook payload is unchanged. The sessionmaker change matches the async session's existing behaviour. The snapshot dataclasses are private (\`_RecordSnapshot\` / \`_DataSourceSnapshot\`) and only used by the webhook dispatcher.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced webhook event processing reliability to prevent data access issues during database transactions.
  * Improved session management to ensure event record data remains accessible and consistent after transaction completion.

### DIFF
--- a/backend/app/database.py
+++ b/backend/app/database.py
@@ -39,7 +39,10 @@ async_engine = create_async_engine(settings.db_uri)
 
 
 def _prepare_sessionmaker(engine: Engine) -> sessionmaker:
-    return sessionmaker(autocommit=False, autoflush=False, bind=engine)
+    # expire_on_commit=False matches the async sessionmaker below and lets
+    # `after_commit` listeners (e.g. webhook dispatch in EventRecordService)
+    # read ORM attributes without triggering a lazy-load on a closed session.
+    return sessionmaker(autocommit=False, autoflush=False, expire_on_commit=False, bind=engine)
 
 
 def _prepare_async_sessionmaker(engine: AsyncEngine) -> async_sessionmaker:
--- a/backend/app/services/event_record_service.py
+++ b/backend/app/services/event_record_service.py
@@ -1,6 +1,8 @@
+from dataclasses import dataclass
 from datetime import datetime
 from decimal import Decimal
 from logging import Logger, getLogger
+from typing import Any
 from uuid import UUID, uuid4
 
 from sqlalchemy import event as sa_event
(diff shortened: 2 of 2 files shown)


KaliszS:
Hi @artursmolin. Sorry for the huge delay, but this issue has been already solved in #1208. This PR didn't solve the root cause of the issue, so it should be closed anyway. Thanks for taking the time to look into it though - I really appreciate it!

PR #1004, opened 2026-05-05
fix(bazard): emit connection.created on OAuth update path too

Ri087:
## Bazard fork patch

Upstream OW : \`on_connection_created\` only fires when a brand-new row is created in \`user_connections\`. The update branch (when an existing connection is reauthorised after revoke) emits nothing.

In Bazard's flow, every disconnect+reconnect goes through the update branch (revoke is a soft-delete in OW). Result : downstream consumer (api.bazard.run) never receives the webhook → \`wearable_connections.status\` stuck at \`pending\` → card UI says \"En attente\" forever.

## Fix

Patch the update branch in \`base_oauth.py\` to also call \`on_connection_created\`. The downstream handler is idempotent (UPDATE to \`active\` is a no-op when already active), so re-emitting on every successful (re)authorisation is safe.

## Test plan

- [ ] Build OW image, redeploy on Coolify.
- [ ] Disconnect Strava on Bazard, reconnect.
- [ ] Logs api.bazard.run : \`POST /api/v1/webhooks/openwearables 200\` + \`ow webhook: connection activated\`.
- [ ] Card UI passes from \"En attente\" to \"Connecté\" within 10s.

Linear : BAZ-177 (suite).

## Release Notes

* **Bug Fixes**
  * Webhook events now properly trigger when re-authorizing connections, ensuring seamless integration updates.

* **Chores**
  * Production deployment configuration updated to support environment-variable overrides for database credentials, improving security and flexibility.
  * Operational documentation enhanced with fork management guidelines.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,5 +1,119 @@
-# CLAUDE.md
+# CLAUDE.md — wearables.bazard.run
 
-Please follow the guidelines and project structure defined in ./AGENTS.md
+> **Fork Bazard.run** de [`the-momentum/open-wearables`](https://github.com/the-momentum/open-wearables) (MIT).
+> Sert de **passerelle wearables** pour l'écosystème Bazard.run : agrège Garmin, Strava, Oura, Whoop, Fitbit, Polar, Suunto, Ultrahuman, Apple Health, Google Health Connect derrière une API REST unifiée.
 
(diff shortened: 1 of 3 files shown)


Ri087:
Closing — implementing this differently in our consumer (api.bazard.run polls OW connection state on OAuth return) instead of patching OW. Sorry for the noise.

PR #1042, opened 2026-05-17
fix(integrations): handle Suunto workout webhook edge cases

HugoHeneault:
## Summary

`WORKOUT_CREATED` Suunto webhooks crashed in Celery with `AttributeError: 'str' object has no attribute 'gear'` because `/v3/workouts/{workoutKey}` returns a single workout dict (`{"payload": {...}}`), not a list. The previous normalization iterated dict keys (strings) and passed them to `_process_single_workout`.

After fixing the iteration, a second crash surfaced: `SuuntoWorkoutJSON.stopTime` was marked required, but Suunto omits the field for workouts it has not yet finalized server-side, so Pydantic raised a `ValidationError`.

Periodic sync (`/v3/workouts` list endpoint) was unaffected because it always returns a list and full workout records with `stopTime` populated.

## Change

- `SuuntoWebhookHandler._process_workout`: normalize `payload` to a list before iterating, regardless of whether the REST API returned a single dict or a list.
- `SuuntoWorkoutJSON.stopTime` is now `int | None = None`.
- `SuuntoWorkouts._normalize_workout` falls back to `startTime + totalTime * 1000` when `stopTime` is absent. This matches Suunto's FIT-derived convention (`totalTime` is the active timer time; pauses are excluded). The periodic sync overwrites with the canonical value later if Suunto emits a different one.
- New fixture `backend/tests/providers/suunto/fixtures/workout_single_payload.json` captures the exact live response shape that reproduced the bug (committed via `-f` because the repo's `.gitignore` blanket-ignores `*.json`).

## Test plan

- [x] Added `test_suunto_webhook_handler.py` covering single-dict and list `payload` shapes, the missing-`workoutKey` early return, and Pydantic acceptance of the live response.
- [x] Full `pytest tests/providers/suunto/` suite green (36 passed).
- [x] `ruff check` / `ruff format` clean on touched files.
- [x] Validated against staging: pushed a real Suunto workout, observed:
  ```
  Processing Suunto webhook event ... event_type=WORKOUT_CREATED
  Fetching Suunto workout via REST ... workout_key=...
  HTTP/1.1 200 OK (GET /v3/workouts/<key>)
  process_webhook_push[...] succeeded ... 'status': 'saved', 'saved_count': 1
  ```
  No `AttributeError`, no `ValidationError`.

## Known limitation (out of scope here)

Suunto's `totalTime` is the active timer time (FIT `total_timer_time`), so pauses are excluded. The fallback `stopTime` therefore equals `startTime + active_time`, which matches `duration_seconds` but underestimates real elapsed time when the user paused. Standard fitness-app convention. Accurate elapsed handling would require querying `?extensions=PauseMarkerExtension` and summing pause markers; that is left for a follow-up PR #1043  because the extension's response shape is not publicly documented.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve Suunto workout import: handle missing end-time by deriving it from start time + duration; correctly process both single-workout and batch-sync webhook payload shapes; gracefully ignore duplicate saves.
* **Tests**
  * Add coverage validating payload normalization, missing end-time parsing, batch vs single payload handling, error paths, and duplicate-save behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/the-momentum/open-wearables/pull/1042?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/app/schemas/providers/suunto/workout_import.py
+++ b/backend/app/schemas/providers/suunto/workout_import.py
@@ -35,7 +35,9 @@ class WorkoutJSON(BaseModel):
 
     # Unix timestamp (ms)
     startTime: int
-    stopTime: int
+    # Sometimes missing in fresh webhook payloads (Suunto computes it asynchronously);
+    # callers fall back to startTime + totalTime when None.
+    stopTime: int | None = None
     # Seconds
     totalTime: float
     timeOffsetInMinutes: int | None = None
--- a/backend/app/services/providers/suunto/webhook_handler.py
+++ b/backend/app/services/providers/suunto/webhook_handler.py
@@ -21,6 +21,7 @@ from uuid import uuid4
 
 from celery import current_app as celery_app
 from fastapi import HTTPException, Request
+from sqlalchemy.exc import IntegrityError
 
 from app.config import settings
 from app.database import DbSession
@@ -223,13 +224,27 @@ class SuuntoWebhookHandler(BaseWebhookHandler):
 
         try:
             raw_detail = self.suunto_workouts.get_workout_detail(db, user_id, str(workout_key))
-            # REST response wraps workout in 'payload' list
-            workouts_list = raw_detail.get("payload", [raw_detail]) if isinstance(raw_detail, dict) else [raw_detail]
+            # `/v3/workouts/{workoutKey}` returns a single dict under 'payload'; `/v3/workouts` (sync) returns a list.
+            payload_detail = raw_detail.get("payload", raw_detail) if isinstance(raw_detail, dict) else raw_detail
+            workouts_list = payload_detail if isinstance(payload_detail, list) else [payload_detail]
             saved = 0
             for raw in workouts_list:
                 self.suunto_workouts._process_single_workout(db, user_id, raw)
                 saved += 1
             return {"status": "saved", "workout_key": str(workout_key), "saved_count": saved}
+        except IntegrityError:
+            db.rollback()
+            log_structured(
+                logger,
+                "info",
+                "Suunto workout already exists, skipping",
+                provider="suunto",
+                trace_id=trace_id,
+                action="webhook_duplicate_workout",
+                workout_key=str(workout_key),
+                user_id=str(user_id),
+            )
+            return {"status": "ignored", "reason": "duplicate_workout", "workout_key": str(workout_key)}
         except Exception as exc:
             log_structured(
                 logger,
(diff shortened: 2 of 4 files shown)


backend/tests/providers/suunto/fixtures/workout_single_payload.json
czajkub, line 1: Can you please turn this json file to an inline dict inside the test?  It's the way tests are done currently, and the less variance in them the less issues in the future

czajkub (changes requested):
Btw, what will happen if a webhook will send the workout again after it finishes, will it update the one already existing (with wrong stop time one), will it duplicate the events or will it hit an integrity violation somewhere? Did you test that?

backend/tests/providers/suunto/fixtures/workout_single_payload.json
HugoHeneault: Hey @czajkub 

Thanks for reviewing!

I didn't test before but added a test about a duplicate webhook.
I also changed fixture to match existing ones :)

czajkub (changes requested):
Okay, I looked deeper into this. Im pretty sure this approach is gonna cause duplicate workouts to be inserted to the database, as the index on `EventRecord` is `"data_source_id", "start_datetime", "end_datetime", unique=True`; and since the end datetime is different on each partial webhook sent, both rows are gonna be inserted.

The solution i see is that there should be an upsert logic in `event_record_repository` that will check whether records with the same external id exist in database, and update the fields, whether it would be adding a new constraint in the model or python level logic, however that is something that we would need to discuss among the team. For now I won't be approving, but I'll do it once I make sure that duplication wont happen here.

HugoHeneault:
> Okay, I looked deeper into this. Im pretty sure this approach is gonna cause duplicate workouts to be inserted to the database, as the index on `EventRecord` is `"data_source_id", "start_datetime", "end_datetime", unique=True`; and since the end datetime is different on each partial webhook sent, both rows are gonna be inserted.

Hey @czajkub thanks for looking deeper!

Why would `"start_datetime", "end_datetime"` change between webhooks?
If they change for sure it's not good with current DB constraints.

czajkub:
@HugoHeneault im talking about this part:
```python
stop_time_ms = raw_workout.stopTime or raw_workout.startTime + int(raw_workout.totalTime * 1000)
start_date, end_date = self._extract_dates(raw_workout.startTime, stop_time_ms)
```
and then this end_date is added to `EventRecordCreate`, which is inserted into database - and then the `end_date` from this partial webhook would differ from the end_date when the workout is finalized. So, what I would do is either only accept the webhook when it is finalized (has the stop_time filled) or create a partial index, however the more i think about it I'm curious why the webhooks are sent in a non-finished state

czajkub (approved):

PR #1043, opened 2026-05-17
feat(integrations): pause-aware stopTime + gear from SummaryExtension

HugoHeneault:
> Stacked on top of #1042. Review #1042 first; this PR's diff against `main` will shrink to a single commit once #1042 merges.

## Summary

The webhook handler now fetches `/v3/workouts/{workoutKey}?extensions=PauseMarkerExtension,SummaryExtension` and uses the returned data to:

1. Compute `end_datetime` as `startTime + active_time + sum(pauses)` instead of approximating `start + active_time`. Suunto's `totalTime` is the active timer time (FIT `total_timer_time`); pauses live in `PauseMarkerExtension` and were previously silently dropped, so the persisted `end_datetime` was off by the total paused duration whenever the user paused mid-workout.
2. Populate `source_name`/`device_model` from `SummaryExtension.gear` when the top-level `gear` block is absent. Newer Suunto watches (Suunto Race 2 in particular) ship gear info only inside the extension; without this fix, those workouts were saved with `source_name="Suunto"` and no `device_model`.

## Why

Discovered while validating #1042 on staging: a test workout with two manual pauses (69.92 s + 20.77 s) returned `totalTime = 210.56 s` and no `stopTime`, so the fallback in #1042 stored `end - start = 210.56 s` while the real elapsed time was 301.25 s. The same workout came from a Race 2, which surfaced the gear-in-extension shape.

## Change

- `SuuntoWorkoutJSON` gains an `extensions: list[dict] | None` field, a `pause_markers` helper, and a `gear_from_summary_extension` helper.
- `DeviceJSON` accepts both `swVersion`/`hwVersion` (legacy) and `softwareVersion`/`hardwareVersion` (SummaryExtension) via `AliasChoices` so the same model parses both shapes.
- `SuuntoWorkouts.get_workout_detail` accepts an `extensions: list[str] | None` parameter and forwards it as the `?extensions=` query string.
- `SuuntoWebhookHandler._process_workout` requests `PauseMarkerExtension,SummaryExtension` on every call.
- `_normalize_workout` and `_process_single_workout` consult the new helpers; the periodic sync path is untouched.

## Test plan

- [x] New fixture `backend/tests/providers/suunto/fixtures/workout_with_pauses.json`. Timestamps are remapped onto a synthetic `startTime = 1_700_000_000_000`, but the active time (`210.56 s`), pause durations (`69.92 s`, `20.77 s`), and gear shape are copied directly from the live capture so the arithmetic in tests mirrors reality.
- [x] `tests/providers/suunto/test_suunto_webhook_handler.py` covers pause-marker parsing, gear extraction from `SummaryExtension`, the no-extensions baseline, and an end-to-end `_normalize_workout` assertion that `end - start = 301.25 s` with `duration_seconds = 210`.
- [x] Full Suunto suite green (`pytest tests/providers/suunto/` — 41 passed).
- [x] `ruff check` / `ruff format --check` clean on touched files.
- [x] Staging validation: pushed a real Suunto workout with pauses, observed `GET /v3/workouts/<key>?extensions=PauseMarkerExtension,SummaryExtension → 200`, task completed with `status: saved, saved_count: 1`, no exceptions.

## Out of scope

- No periodic sync changes. Sync still hits `/v3/workouts` without extensions; if Suunto returns a canonical `stopTime` there, it overrides the webhook estimate the next time around.
- No backfill for previously-saved workouts whose `end_datetime` is off because of pauses.

## Summary by CodeRabbit

* **New Features**
  * Added support for tracking workout pauses and pause durations.
  * Enhanced device compatibility to accept alternative firmware and hardware version formats.

* **Bug Fixes**
  * Improved handling of duplicate workouts with proper error reporting.
  * Added automatic calculation of workout end time when not explicitly provided.

* **Tests**
  * Added comprehensive test coverage for workout processing and pause-aware timing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/the-momentum/open-wearables/pull/1043?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


backend/app/schemas/providers/suunto/workout_import.py
```diff
@@ -73,6 +94,39 @@ class WorkoutJSON(BaseModel):
     workoutName: str | None = Field(default=None, alias="name")
     notes: str | None = None
 
+    # Polymorphic list of typed extensions returned when fetched with ?extensions=...
+    # (e.g. PauseMarkerExtension, SummaryExtension). Kept loose because each extension
+    # has a different shape; helpers in `services.providers.suunto.workouts` parse them.
+    extensions: list[dict[str, Any]] | None = None
+
+    @property
+    def pause_markers(self) -> list[PauseMarker]:
```
czajkub, line 161: Are there any other extensions that may be extracted in the future? If so, this would need to be adjusted to accept other types, or a different method for each extension, which would make little sense. Maybe a method like `extract_extensions` that would return a list of extensions, or maybe even a dict with keys as type of extensions and values as the extension type would work

backend/app/schemas/providers/suunto/workout_import.py
```diff
@@ -73,6 +94,39 @@ class WorkoutJSON(BaseModel):
...
+                continue
+            # `pauseMarkers` may be null, missing, or unexpectedly shaped.
+            markers = ext.get("pauseMarkers")
+            if not isinstance(markers, list):
+                return []
+            return [PauseMarker(**m) for m in markers if isinstance(m, dict)]
+        return []
+
+    @property
+    def gear_from_summary_extension(self) -> "DeviceJSON | None":
```
czajkub: Here again is another method iterating through every extension, maybe could parse all the extensions at once

backend/app/schemas/providers/suunto/workout_import.py
```diff
@@ -73,6 +94,39 @@ class WorkoutJSON(BaseModel):
     workoutName: str | None = Field(default=None, alias="name")
     notes: str | None = None
 
+    # Polymorphic list of typed extensions returned when fetched with ?extensions=...
+    # (e.g. PauseMarkerExtension, SummaryExtension). Kept loose because each extension
+    # has a different shape; helpers in `services.providers.suunto.workouts` parse them.
+    extensions: list[dict[str, Any]] | None = None
```
czajkub: This here could be a list of abstract `BaseModel`, like `Extension`, and then in schemas there could be defined classes that inherit from `Extension`, to verify its integrity

backend/app/services/providers/suunto/webhook_handler.py
```diff
@@ -222,9 +222,24 @@ def _process_workout(self, db: DbSession, user_id: Any, payload: dict[str, Any],
...
-            raw_detail = self.suunto_workouts.get_workout_detail(db, user_id, str(workout_key))
-            # REST response wraps workout in 'payload' list
-            workouts_list = raw_detail.get("payload", [raw_detail]) if isinstance(raw_detail, dict) else [raw_detail]
+            # Request PauseMarkerExtension so we can compute real elapsed time including pauses,
+            # and SummaryExtension because newer watches (e.g. Race 2) ship gear info there.
+            raw_detail = self.suunto_workouts.get_workout_detail(
+                db,
+                user_id,
+                str(workout_key),
+                extensions=["PauseMarkerExtension", "SummaryExtension"],
```
czajkub: If you want, you can try to extract these extensions somewhere else, to signal that when someone wants to add support to another extension, that it would also need to be added in this passed list

backend/tests/providers/suunto/fixtures/workout_single_payload.json
czajkub, line 1: For tests, same case as in #1042 - could you move them inline?

czajkub (changes requested):

backend/app/schemas/providers/suunto/workout_import.py
HugoHeneault: Switched extensions to a discriminated Pydantic union (`PauseMarkerExtension` / `SummaryExtension` + `UnknownExtension` catch-all), pattern from `backend/app/schemas/providers/mobile_sdk/sdk_log_events.py`. 

Discriminator is callable so unknown type values route to `UnknownExtension` instead of failing validation.

backend/app/schemas/providers/suunto/workout_import.py
HugoHeneault: Suunto's API exposes ~36 extension types (full list in their `?extensions=` query param docs). We only care about two today (`PauseMarkerExtension`, `SummaryExtension`), so I kept the union narrow and added a typed catch-all `UnknownExtension` so unknown types never break validation.

Adding a new extension is now scoped:

1. Define the typed Pydantic class (e.g. `WeatherExtension(BaseModel)` with `type: Literal["WeatherExtension"]`).
2. Add it to the `WorkoutExtension` discriminated union with `Tag(WeatherExtension.__name__)`.
3. Add the class name to `REQUESTED_WORKOUT_EXTENSIONS`.

backend/app/schemas/providers/suunto/workout_import.py
HugoHeneault: Added the _find_extension helper

backend/tests/providers/suunto/fixtures/workout_single_payload.json
HugoHeneault: Yup, did that :)

HugoHeneault:
@czajkub just checked live, seem to work on my side: 

```
 external_id |      start_datetime       |        end_datetime        | elapsed_s | active_s |  source_name  | device_model
-------------+---------------------------+----------------------------+-----------+----------+---------------+---------------
 0           | 2026-05-19 08:57:26.62+00 | 2026-05-19 08:58:35.995+00 |        69 |       37 | Suunto Race 2 | Suunto Race 2

```

czajkub (commented):
Looks good now, ill give approve when issue from #1042 is resolved

czajkub:
@coderabbitai did anything break during the merge conflict resolution?

czajkub (approved):

PR #1045, opened 2026-05-17
fix(integrations): emit outgoing webhook for suunto live workout push

HugoHeneault:
## Problem

Suunto workouts received over the live webhook push path (`WORKOUT_CREATED`) never trigger an outgoing `workout.created` event. Only signup-time backfill workouts reach downstream Svix subscribers; every live watch upload is silently dropped between OW and the consumer.

## Root cause

`SuuntoWorkouts._process_single_workout` (called from `SuuntoWebhookHandler._process_workout`) delegates to `BaseWorkouts._save_workout`:

```python
def _save_workout(self, db, record, detail):
    self.workout_repo.create(db, record)
    # Detail saving is handled by services in load_data implementations
```

That method inserts the `event_record` row but silently ignores the `detail` argument. Consequences:

- No `event_record_detail` row is written.
- The SQLAlchemy `after_commit` listener registered by `EventRecordService.create_detail` never fires.
- The outgoing `workout.created` event is never dispatched to Svix.

Strava (`process_push_activity`), Garmin (`process_push_activities`), Whoop (`load_single_workout`) and Oura (`save_by_id`) already handle the live push path correctly by calling `event_record_service.create` + `create_detail`. Suunto was the only remaining provider routing live push through the broken base helper.

## Fix

Add `SuuntoWorkouts.process_push_activity` mirroring the other live-push paths, and switch `SuuntoWebhookHandler` to call it instead of `_process_single_workout`. The new method:

- Ensures the data source for the gear.
- Builds the record + detail bundle via `_build_bundles`.
- Calls `event_record_service.create` + `event_record_service.create_detail`, so the `after_commit` listener schedules `emit_webhook_event` and the workout flows downstream.

## Tests

- New regression `test_process_push_activity_creates_record_and_detail` asserts both `create` and `create_detail` run on the push path.
- Full Suunto test module passes (13 tests).

## End-to-end verification

Verified on a staging deployment: with the fix, a new live Suunto workout triggers `emit_webhook_event` → Svix → consumer ingest. Without the fix, only signup-time backfill workouts ever reach the consumer.

## Summary by CodeRabbit

* **Improvements**
  * Improved webhook handling for pushed workouts so they reliably create full workout records and associated details; the saved count now only increments when a workout is actually persisted, reducing false positives and improving sync consistency.

* **Tests**
  * Added test coverage to validate webhook-driven workout record and detail creation and related data provisioning.

[Review Change Stack](https://app.coderabbit.ai/change-stack/the-momentum/open-wearables/pull/1045?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/app/services/providers/suunto/webhook_handler.py
+++ b/backend/app/services/providers/suunto/webhook_handler.py
@@ -227,8 +227,8 @@ class SuuntoWebhookHandler(BaseWebhookHandler):
             workouts_list = raw_detail.get("payload", [raw_detail]) if isinstance(raw_detail, dict) else [raw_detail]
             saved = 0
             for raw in workouts_list:
-                self.suunto_workouts.process_push_activity(db, user_id, raw)
-                saved += 1
+                if self.suunto_workouts.process_push_activity(db, user_id, raw) is not None:
+                    saved += 1
             return {"status": "saved", "workout_key": str(workout_key), "saved_count": saved}
         except Exception as exc:
             log_structured(
--- a/backend/app/services/providers/suunto/webhook_handler.py
+++ b/backend/app/services/providers/suunto/webhook_handler.py
@@ -227,7 +227,7 @@ class SuuntoWebhookHandler(BaseWebhookHandler):
             workouts_list = raw_detail.get("payload", [raw_detail]) if isinstance(raw_detail, dict) else [raw_detail]
             saved = 0
             for raw in workouts_list:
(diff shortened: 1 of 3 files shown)


czajkub (approved):
Looks good to me, it just seems that PR title check is failed (it's not that important, but if you do decide to change it you would need to commit something again to re-run the checks (or there might be an option to re-run the check, not sure if its available to non-members)

HugoHeneault:
@czajkub I updated the title of commits/PR

PR #1047, opened 2026-05-18
fix: Handle duplicate EventRecordDetail inserts gracefully with savepoint

### LINKED ISSUE #730: Handle duplicate EventRecordDetail inserts gracefully with savepoint
## Problem

`EventRecordDetailRepository.create_and_flush()` does a plain ORM add+flush with no conflict handling. When a sync re-processes data that already exists (e.g. due to overlapping date windows), it hits a `UniqueViolation` on `event_record_detail_pkey`, which poisons the SQLAlchemy session and causes a `PendingRollbackError` on the next `commit()`.

199 occurrences since January in our internal instance.

## Root cause

In `create_or_merge_sleep()` (`event_record_service.py:306-312`), when no adjacent session is found:

1. `EventRecordRepository.create_and_flush()` is called - this **does** handle duplicates via savepoint, returning the existing record on `IntegrityError`
2. `EventRecordDetailRepository.create_and_flush()` is called next for that `record_id` - but this method has **no** savepoint/conflict handling, so it blows up if the detail already exists

## Fix

Add savepoint handling to `EventRecordDetailRepository.create_and_flush()`, matching the pattern already used in `EventRecordRepository.create_and_flush()` (lines 94-111). On `IntegrityError`, rollback the nested transaction and return the existing detail.

## Related

Related to #704 - the fixed 30-day sync window causes massive data overlap on every hourly sync, which is the main trigger for these duplicate inserts. Even after #704 is resolved, some overlap is intentional (to catch delayed data), so this defensive handling is still needed.

czajkub:
## Description

`EventRecordDetailRepository.create_and_flush()` does a plain ORM add+flush with no conflict handling. When a sync re-processes data that already exists (e.g. due to overlapping date windows), it hits a `UniqueViolation` on `event_record_detail_pkey`, which poisons the SQLAlchemy session and causes a `PendingRollbackError` on the next `commit()`.

Resolves #730 

### General

### Backend Changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of duplicate event record details. The system now gracefully manages conflicts when creating event records, preventing errors and recovering existing records instead of failing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/the-momentum/open-wearables/pull/1047?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/app/repositories/event_record_detail_repository.py
+++ b/backend/app/repositories/event_record_detail_repository.py
@@ -3,6 +3,7 @@ from uuid import UUID
 
 from sqlalchemy import Table
 from sqlalchemy.dialects.postgresql import insert
+from sqlalchemy.exc import IntegrityError
 
 from app.database import DbSession
 from app.models import (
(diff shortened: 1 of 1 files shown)


backend/app/repositories/event_record_detail_repository.py
```diff
@@ -61,11 +62,23 @@ def create_and_flush(
...
+        """Like create() but flushes instead of committing; caller is responsible for the commit.
+
+        Uses a savepoint for IntegrityError handling so a conflict rolls back only
+        the INSERT and leaves the outer transaction intact.
+        """
         detail = self._build_detail(creator, detail_type)
-        db_session.add(detail)
-        db_session.flush()
-        return detail
+        nested = db_session.begin_nested()
```
KaliszS, line 71: @coderabbitai Won't it slow down entire ORM flow? We have to make sure those sub-sessions will be efficient enough.

KaliszS (approved):

PR #1066, opened 2026-05-23
fix(mcp): suppress FastMCP banner — defense-in-depth for stdio transport

justingreenberg:
## TL;DR

Small defense-in-depth change: pass `show_banner=False` to `mcp.run()` so STDOUT stays clean when this MCP server is invoked directly via `mcp.run()` (rather than via the `fastmcp run` CLI runner). Not a load-blocker fix on its own — see "What's NOT this PR" below.

## What this PR does

`mcp.run()` defaults to `show_banner=True`, which prints a decorative ASCII-art banner to **STDOUT** during startup. For HTTP/SSE transports this is harmless. For the **stdio transport** (default for Claude Code and most MCP clients), STDOUT is the JSON-RPC channel — any non-JSON bytes there corrupt the stream.

Passing `show_banner=False` ensures STDOUT is clean during direct-invocation startup. FastMCP's Python `logging` continues to go to STDERR as normal (the correct destination for diagnostic logs in stdio mode).

## What's NOT this PR (additional discovery)

After opening this PR, I discovered that **even with the banner suppressed, the server still failed to register with Claude Code via the current `start` entry-point.** Root cause turned out to be different: the `start` script invokes `mcp.run()` directly, which spawns the **FastMCP `docket` worker synchronously** during stdio init. The worker startup adds latency that exceeds Claude Code's MCP handshake timeout → Claude Code silently drops the server.

The actual load-blocker fix (not part of this PR) is at the **invocation layer** — clients should invoke this server via:

```bash
uv run --directory ./mcp fastmcp run app/main.py
```

instead of the `start` console-script entry-point. The `fastmcp run` CLI runner defers docket worker startup until after the MCP handshake completes.

## Why this PR is still worth merging

Even though the banner isn't the load-blocker, this change is:

1. **Correct behavior** — printing decorative banners to STDOUT in a server that uses STDOUT for JSON-RPC is a bug regardless of which client/runner is used
2. **Defense-in-depth** — any future user invoking `from app.main import mcp; mcp.run()` directly (e.g., for testing, embedding, alternative runners) hits the banner-on-stdout issue this fixes
3. **Zero behavior change for the CLI-runner path** — `fastmcp run` already manages banner output differently; this change doesn't affect that path
4. **Trivial diff** — one keyword argument

## Suggested follow-up (separate PR or issue)

Consider also updating this repo's `start` console-script (pyproject.toml `[project.scripts] start = "app.main:main"`) to mirror the apple-health-style wrapper pattern — subprocess-exec via `fastmcp run` — so the existing `start` entry-point is also Claude-Code-compatible without requiring users to change their MCP client config. That's a more invasive change so leaving for separate discussion.

## Verification (banner suppression specifically)

```bash
# Before
uv run --frozen --directory ./mcp start 1>stdout.txt 2>stderr.txt &
sleep 4 && kill %1
wc -c stdout.txt
# → 2479 stdout.txt (banner pollutes JSON-RPC channel)

# After
uv run --frozen --directory ./mcp start 1>stdout.txt 2>stderr.txt &
sleep 4 && kill %1
wc -c stdout.txt
# → 0 stdout.txt ✅
```

## References

- FastMCP `show_cli_banner` setting + `FASTMCP_SHOW_CLI_BANNER` env var
- MCP stdio transport spec: https://spec.modelcontextprotocol.io/specification/transports/#stdio

### DIFF
--- a/mcp/app/main.py
+++ b/mcp/app/main.py
@@ -126,8 +126,15 @@ logger.info(f"Open Wearables MCP server initialized. API URL: {settings.open_wea
 
 
 def main() -> None:
-    """Entry point for the MCP server."""
-    mcp.run()
+    """Entry point for the MCP server.
+
+    show_banner=False is REQUIRED for Claude Code MCP integration.
+    FastMCP's decorative banner output uses print() → stdout, which
+    pollutes the JSON-RPC stream Claude Code expects on stdio.
+    With banner enabled, Claude Code silently fails to load the
+    MCP server (no error surfaced — just no tools available).
+    """
+    mcp.run(show_banner=False)
 
 
 if __name__ == "__main__":


KaliszS:
Hi Justin Greenberg. Sorry for the silence. We have to keep this PR on hold for a little while longer because we're not ready to return to the MCP topic yet, but we will do so in the near future. At that point, this PR and the entire MCP server will be reviewed. Thanks for your patience!

PR #1089, opened 2026-05-29
Fire sleep.created on all committed saves (match docstring contract)

davecoached:
## Problem

`SLEEP_CREATED`'s description in `event_types.py` reads *"A new (or merged) sleep session was saved."* and the module docstring is explicit that there's no update/patch flow. But `event_record_service.create_or_merge_sleep` (line 211) only fires `on_sleep_created` when `inserted=True`, which skips three merge-path returns inside `_create_or_merge_sleep_inner`:

- Line 285 — re-ingestion (same `external_id`): detail replaced with fresh values, then `return adjacent, False, detail`.
- Line 366 — `same_window` merge: detail updated in-place, `return adjacent, False, detail`.
- Line 378 — constraint collision treated as same-window: same outcome, `return adjacent, False, detail`.

All three modify the database. None fire the webhook.

## Real-world impact

In a production deployment, a freshly-paired WHOOP user has 91 sleep-category `HealthScore` rows but only 1 `EventRecord`. That single record fired `sleep.created` once; every subsequent WHOOP backfill night merged into it via the same-window / external_id paths and was silently swallowed. Subscribers see one stale snapshot.

## Fix

1. **`event_record_service.create_or_merge_sleep`** — drop the `if inserted:` gate. Fire `on_sleep_created` on every committed save, passing `is_merge=not inserted` so consumers can distinguish fresh vs. merged saves.
2. **`outgoing_webhooks/events.on_sleep_created`** — accept the new `is_merge: bool = False` kwarg, include it in the payload, and hash the mutable fields into the `idempotency_key`. Preserves Svix dedup for genuine retries of identical payloads while still producing distinct delivery attempts for successive merges of the same record.

No new event type, no `*_UPDATED` flow — stays inside the project's documented "created-only" design.

## Test plan

I have not added tests in this PR — happy to follow up once you can point me at the test scaffolding (e.g. `tests/services/event_record_service/...`). Suggested assertions:

- Fresh insert → `on_sleep_created(is_merge=False)`; idempotency_key contains content hash.
- Re-ingestion (same `external_id`) with modified detail → `on_sleep_created(is_merge=True)` with different idempotency_key.
- No-op re-ingestion (byte-identical detail) → same idempotency_key (Svix-side dedup still works for retries).

## Summary by CodeRabbit

* **Bug Fixes**
  * Sleep session webhooks now dispatch for both newly created and merged sessions, ensuring comprehensive tracking of all sleep data modifications without gaps.
  * Webhook event payloads now explicitly identify whether each update resulted from a create or merge operation, providing clear visibility into how sleep records are being modified and synchronized.

[Review Change Stack](https://app.coderabbit.ai/change-stack/the-momentum/open-wearables/pull/1089?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/app/services/event_record_service.py
+++ b/backend/app/services/event_record_service.py
@@ -209,41 +209,45 @@ class EventRecordService(
         first, and the old record is deleted only after a successful insert — so a failure
         never loses the original data.
         """
-        result, inserted, final_detail = self._create_or_merge_sleep_inner(
+                result, inserted, final_detail = self._create_or_merge_sleep_inner(
             db_session, user_id, record, detail, threshold_minutes
         )
-        if inserted:
-            eff = final_detail.sleep_efficiency_score
-            has_stages = any(
-                [
-                    final_detail.sleep_awake_minutes,
-                    final_detail.sleep_light_minutes,
-                    final_detail.sleep_deep_minutes,
-                    final_detail.sleep_rem_minutes,
-                ]
-            )
-            on_sleep_created(
-                record_id=result.id,
-                user_id=user_id,
-                provider=record.provider or record.source_name,
-                device=record.device_model,
-                start_time=result.start_datetime.isoformat(),
-                end_time=result.end_datetime.isoformat(),
-                zone_offset=record.zone_offset,
-                duration_seconds=result.duration_seconds,
-                efficiency_percent=float(eff) if eff is not None else None,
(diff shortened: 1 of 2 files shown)


KaliszS:
In order to get review, all pipeline steps has to be green first. Thanks!

PR #1093, opened 2026-06-01
feat(backend): Strava workout samples

### LINKED ISSUE #1050: Pre-fill per-sample data on workout webhook arrival?
Hi there,

When receiving a workout webhook it would be nice to have per-sample data already filled (HR curves, speed, cadence, etc.), so consumers can render rich recap UIs immediately instead of waiting for a delayed 24/7 or activity stream.

**Proposal:** pre-fill per-sample data at workout webhook reception, behind a single global feature flag.

An half way to #953 so more data are provided but not all of them.

## Problem

Across providers, OW receives a workout/activity webhook within seconds of the workout ending, fetches summary stats, and saves `event_record` + `event_record_detail`. **Per-sample series (HR/sec, steps/sec, speed, cadence, power, elevation) are not fetched on workout arrival.** They land later (sometimes hours later) via a separate stream, a separate webhook, or never at all.

Consumers reading `/timeseries` or subscribing to `heart_rate.created` / `series.heart_rate.created` at workout arrival time get an empty window. Recap UIs rendering an HR curve at claim time show null values. And even later it's not showing enough data for accurate lookup (like intervals which sometimes happen in <60s)

### Suunto

- `WORKOUT_CREATED` arrives within seconds. `_process_workout` ([`webhook_handler.py`](https://github.com/the-momentum/open-wearables/blob/main/backend/app/services/providers/suunto/webhook_handler.py)) saves summary only.
- Per-sample HR, steps, SpO₂, energy, HRV only ingested on `SUUNTO_247_ACTIVITY_CREATED` (minutes to hours later).
- API path: `GET /247samples/activity?from=<ms>&to=<ms>` (≤1Hz). Helper `Suunto247Data.get_activity_samples` already exists.
- Normaliser `Suunto247Data.normalize_activity_samples` ([`data_247.py`](https://github.com/the-momentum/open-wearables/blob/main/backend/app/services/providers/suunto/data_247.py)) already fans into 5 `SeriesType` keys.
- Raw file: none. JSON `/247samples/activity` is the ceiling.

### Garmin

- Already tracked by [`#953`](https://github.com/the-momentum/open-wearables/issues/953). `activityDetails.samples[]` (HR, GPS, speed, cadence, power ~1Hz) lives inside the webhook payload but is dropped today. FIT available via PING `callbackURL`, currently skipped.

### Strava

- `strava/webhook_handler.py` calls `GET /api/v3/activities/{id}` (summary only). `/streams` endpoint never called.
- `GET /api/v3/activities/{id}/streams?keys=heartrate,velocity_smooth,cadence,watts,altitude,distance,latlng,time&key_by_type=true` returns every per-sample series in one call.
- Raw file: none via API.

### Polar

- Open PR [`#1041`](https://github.com/the-momentum/open-wearables/pull/1041) implements Polar webhooks plus per-sample HR, steps, route via JSON plus single-exercise webhook-driven fetch pattern. Issue scope after merge: lap parsing, summary field mapping, route polyline persistence, training-load fields, activity zones.
- **Schemas only.** PR does not call `bulk_create_samples` on `samples[]` and does not extract `route[]` into a polyline or per-point series. Both land in `event_record_detail` raw JSON.                                                              
- After merge: same per-sample gap as Suunto/Garmin. Wire the existing schemas to `timeseries_service.bulk_create_samples` for HR, persist `route[]` separately.   

### Wahoo

- Audit needed: verify whether `backend/app/services/providers/wahoo/` is wired today or only stubbed.
- Wahoo Cloud API exposes per-workout FIT downloads. Cleanest case in OW's roster to validate FIT ingestion infra `#953` Phase 2 needs anyway.

### Oura

- `oura/webhook_handler.py:299-300` consumes daily aggregates.
- `GET /v2/usercollection/heartrate?start_datetime=<iso>&end_datetime=<iso>` exposes per-minute HR. Never called on workout completion.
- Raw file: none.

### Fitbit

- Audit needed.
- Per-second HR via `GET /1/user/-/activities/heart/date/{date}/1d/1sec/time/{start}/{end}.json`.
- TCX per logged activity via `GET /1/user/-/activities/{log-id}.tcx` (HR, GPS, per-record fields).
- Mixed path: JSON for HR, TCX for richer fields.

### Persistence

Every per-sample series flows through `timeseries_service.bulk_create_samples`. Upsert-style on `(user_id, source, recorded_at, series_type)`. Later replays over the same window are no-ops.

### Outgoing events

Existing `_emit_timeseries_webhooks` listener inside `bulk_create_samples` fires `on_timeseries_batch_saved` per `series_type`. Outgoing Svix events (`heart_rate.created`, `series.heart_rate.created`, `steps.created`, ...) identical to existing 24/7 or activity paths.

### Failure isolation

Wrap fetch and save per provider in `try/except Exception` plus `logger.exception`. Sample failures must never break workout ingestion. The provider's existing fallback path (24/7 stream, activity stream) remains.

### Feature flag

Single global knob, default `False` so existing deployments keep current behaviour (as it will generate more rows in DB, should be enabled only if needed):

```python
settings.ingest_workout_samples: bool = False
```

Applies across every provider's workout handler. Consumers that want inline samples flip it once via env.

### Retention / archival

Workout-fetched samples land in `data_point_series` via the same upsert path as 24/7 samples, so the existing pipeline handles ageing without new infra:

- `archival_service.run_daily_archival` ([`archival_service.py:56`](https://github.com/the-momentum/open-wearables/blob/main/backend/app/services/archival_service.py)) runs daily.
- `archive_after_days`: rolls live rows older than the cutoff into `data_point_series_archive` (one daily bucket per `(data_source, series_type, aggregation_type)`, aggregated per series type's `AggregationMethod`: avg for HR, sum for steps/energy, etc.).
- `delete_after_days`: drops archive rows past the cutoff (or live rows directly if archival not active).

Trade-offs:

- Raw HR/sec curve is replayable only while inside `archive_after_days`. After that, aggregate-only. Fine for recap UIs at claim time (claim happens within minutes); consumers wanting long-term raw replay tune `archive_after_days` accordingly.
- No per-source / per-series-type retention today. Workout samples and 24/7 samples share the cutoff. If a deployment wants workout samples to outlive 24/7 noise, that's a follow-up (per-`series_type` or per-`data_source` policy on `ArchivalSetting`).

Two options to keep workout samples surviving longer than the global cutoff, ranked by recommendation:

**Option 1 : document downstream persistence as the contract.**
OW emits `heart_rate.created` / `series.heart_rate.created` Svix events on ingestion. Consumers catch the workout window and persist what they care about in their own DB. OW retention stays OW-internal. Zero schema change, matches sidecar role. Risk: missed webhook plus archive elapsed equals raw gone forever for that consumer (mitigated by reliable Svix delivery plus consumer-side replay endpoint).

**Option 2 : per-`data_source` retention overlay.**
`data_point_series` already FKs `data_source_id`. Today every provider has one `data_source` covering both 24/7 and workout context. Split workout context into its own `data_source` (e.g. `suunto-workout` vs `suunto-247`). Extend `ArchivalSetting` with a per-`data_source` overlay (JSONB or join table). Workout-source rows keep raw longer, 24/7 noise rolls up sooner. Reuses existing FK, no new column on `data_point_series`. Ship only if a deployment requires OW itself to retain raw workout samples long-term.

### Shared sample-window helper

After Suunto and Strava ship, extract the (start, end) fetch, normalise, bulk_create, swallow-failure template into the `BaseWebhookHandler` family.

## Test plan (per provider)

`backend/tests/providers/<provider>/test_<provider>_webhook_handler.py`:

- happy path: fetch helper called with workout window. Normaliser plus persistence invoked.
- global flag off: no provider API call.
- fetch raises: swallowed, returns 0 saved.
- missing start or end timestamps: no provider call.
- empty samples returned: no normalise or save.

Ri087:
Resolves #1050

## Report

Ingests Strava per-sample workout streams (HR, speed, cadence, power) on
workout webhook arrival, behind the existing `INGEST_WORKOUT_SAMPLES` flag
(default off), at full fidelity. Mirrors the Garmin slice (#1082): same flag,
same `bulk_create_samples` path (which also emits the `*.created` Svix events),
same failure isolation, self-contained in the provider.

Storage impact (measured, single user, 12 Strava workouts):
- `data_point_series`: +135,696 rows — heart_rate / speed / cadence (~42.3k each),
  power (8.7k, cycling only); ~11,300 rows/workout at ~1 Hz
- ~43 MB for this user's `data_point_series`

## Notes

- `device_model` is set from `activity.device_name` (mirrors fitbit's
  `source_name` and polar's `raw_workout.device`). Without it the outgoing Svix
  payload carried `"device": null` and all samples collapsed into a single
  `data_source` (identity key is `user_id + device_model + source`).
- Swallowed fetch/persist failures go through `log_and_capture_error`
  (per `backend/AGENTS.md`) so they reach Sentry.
- Only seeded `SeriesType`s are ingested; `distance` / `altitude` / `latlng`
  are intentionally excluded until their types are seeded (mirrors Garmin's
  `_ACTIVITY_SAMPLE_FIELD_MAP`).
- The shared sample-window template extraction is deferred to a follow-up, as
  the issue states ("after Suunto and Strava ship").

## Summary by CodeRabbit

* **New Features**
  * Optional per-sample Strava workout ingestion: captures aligned time-series metrics (heart rate, speed, cadence, power) with per-sample timestamps and metadata (source, device, zone). Configurable guard ensures ingestion can be toggled; failures are logged and do not block activity records.

* **Tests**
  * Added tests covering sample generation, metric mapping, time alignment, exclusion of distance, config gating, edge cases, and error-handling for backfill and webhook flows.


Ri087:
@coderabbitai full review

The earlier out-of-scope warning was raised against the initial push, which accidentally included fork-specific files. That has been fixed via force-push — the PR head (5102767) now contains only the two Strava files (`backend/app/services/providers/strava/workouts.py` + its tests). Please re-review the current diff.

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +242,105 @@ def _normalize_workout(
...
+            timeseries_service.bulk_create_samples(db, samples)
+            return len(samples)
+        except Exception as exc:
+            log_and_capture_error(
+                exc,
+                self.logger,
+                "Strava workout sample ingestion failed; continuing",
+                extra={"activity_id": activity.id, "sample_count": len(samples)},
+            )
+            return 0
```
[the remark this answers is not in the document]
Ri087: Good catch — fixed in `bd1ca60`. Wrapped the bulk insert in a savepoint (`db.begin_nested()`) so a failed sample insert rolls back only the samples and leaves the workout's outer transaction usable (mirrors the savepoint pattern in the event_record repositories). The flag-off path still makes no DB call.

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +241,110 @@ def _normalize_workout(
...
+        if not isinstance(streams, dict):
+            return []
+
+        time_data = (streams.get("time") or {}).get("data") or []
+        if not time_data:
+            return []
+
+        mapped: list[tuple[list[Any], SeriesType]] = []
+        for key, series_type in _STRAVA_SAMPLE_KEY_MAP:
+            data = (streams.get(key) or {}).get("data")
```
czajkub: Are you sure this part is correct? From what i see, the response from Strava should be:
```json
[ {
  "type" : "distance",
  "data" : [ 2.9, 5.8, 8.5, 11.7, 15, 19, 23.2, 28, 32.8, 38.1, 43.8, 49.5 ],
  "series_type" : "distance",
  "original_size" : 12,
  "resolution" : "high"
} ] 
```
so basically, instead of the key being associated with the json object itself, the key would be kept inside `"type"` of the object itself. Or is strava documentation wrong? did you test this?

czajkub: so instead it would look something like this:
```python
for s in streams:
    series_type = _STRAVA_SAMPLE_KEY_MAP.get(s["type"])
    data = s.get("data")
```

czajkub: And even better - a schema should be added in `app/schemas/providers/strava` for the API response - then, we won't have to operate on raw dicts and just operate on parsed pydantic classses.

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +241,110 @@ def _normalize_workout(
...
+        strava_activity_id: int | str,
+        start_dt: datetime,
+        zone_offset: str | None,
+        device_model: str | None,
+    ) -> list[TimeSeriesSampleCreate]:
+        """Fetch full-fidelity Strava streams and normalize to TimeSeriesSampleCreate rows."""
+        streams = self._make_api_request(
+            db,
+            user_id,
+            f"/api/v3/activities/{strava_activity_id}/streams",
```
czajkub, line 254: doesn't strava workouts expose `api_base_url` property, or something along those lines? we could use it here if so

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +241,110 @@ def _normalize_workout(
...
+        for key, series_type in _STRAVA_SAMPLE_KEY_MAP:
+            data = (streams.get(key) or {}).get("data")
+            if data:
+                mapped.append((data, series_type))
+
+        if not mapped:
+            return []
+
+        result: list[TimeSeriesSampleCreate] = []
+        for i, offset in enumerate(time_data):
```
czajkub, line 278: isnt offset based on the resolution value here? what i mean by that is:
```
resolution (string):	
The level of detail (sampling) in which this stream was returned
May take one of the following values: low, medium, high
```

also, what is time_data stream supposed to represent here? how does the time stream correlate to offset values inside other streams? is time[i] supposed to be the offset from start at which data[i] sample from other type (like heart rate) was taken? is it a value in seconds?

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +241,110 @@ def _normalize_workout(
...
+            if offset is None:
+                continue
+            recorded_at = start_dt + timedelta(seconds=int(offset))
+            for data, series_type in mapped:
+                if i >= len(data):
```
czajkub, line 283: the stream has a field `original_size` - it can be used to check stream size instead of `len(data)`

backend/app/services/providers/strava/workouts.py
```diff
@@ -229,6 +241,110 @@ def _normalize_workout(
...
+
+        if not mapped:
+            return []
+
+        result: list[TimeSeriesSampleCreate] = []
+        for i, offset in enumerate(time_data):
+            if offset is None:
+                continue
+            recorded_at = start_dt + timedelta(seconds=int(offset))
+            for data, series_type in mapped:
```
czajkub, line 282: also, here we are iterating over each type separately. Wouldn't it be simpler and less computationally expensive to use some sort of list comprehension here?

backend/app/services/providers/strava/workouts.py
```diff
@@ -6,18 +6,30 @@
...
 )
 from app.schemas.providers.strava import ActivityJSON as StravaActivityJSON
 from app.services.event_record_service import event_record_service
 from app.services.providers.templates.base_workouts import BaseWorkoutsTemplate
+from app.services.timeseries_service import timeseries_service
 from app.utils.dates import offset_to_iso
+from app.utils.sentry_helpers import log_and_capture_error
 from app.utils.structured_logging import log_structured
 
+_STRAVA_SAMPLE_KEY_MAP: list[tuple[str, SeriesType]] = [
```
czajkub: these maps here can be moved to `app/constants/strava` or someplace similar

backend/app/services/providers/strava/workouts.py
```diff
@@ -292,11 +408,13 @@ def load_data(
                 )
 
         count = 0
-        for record, detail in self._build_bundles(parsed_activities, user_id):
```
czajkub, line 295: why stop using `_build_bundles` here? now the method is unused, so it should either be deleted, or `_build_bundles` should call `_ingest_workout_streams` within itself so it doesnt have to be done here

backend/tests/providers/strava/test_strava_workout_samples.py
czajkub, line 1: is every test here necessary? it seems a bit excessive

czajkub (changes requested):

backend/app/services/providers/strava/workouts.py
Ri087: This path passes `key_by_type=true`, so Strava returns the `StreamSet` form — a single object keyed by stream type (`{"time": {...}, "heartrate": {...}, ...}`), not the array-with-`type` form. That's why the code reads `streams["heartrate"]` instead of iterating a list. Per the docs, `key_by_type` ("Must be true") returns a `StreamSet` whose properties are named per type. I tested this against a real activity — heart rate and the other streams came back with all their data and mapped to the right offsets.

On the schema suggestion — done: added `StravaStreamSet` / `StravaStream` in `app/schemas/providers/strava/streams.py`, and `_build_workout_samples` now parses via `model_validate()` instead of raw dicts.

backend/app/services/providers/strava/workouts.py
Ri087: Done — moved to `app/constants/series_types/strava.py` (mirroring `series_types/polar.py`), exposing `STREAM_KEY_SERIES_TYPE` and `STREAM_KEYS_PARAM`.

backend/app/services/providers/strava/workouts.py
Ri087: Removed it. Both `load_data` and `process_push_activity` inline `_normalize_workout` + `_ingest_workout_streams` — the latter needs the parsed `activity`, which `_build_bundles` discarded — so it was dead. Deleted rather than re-routing through it.

backend/app/services/providers/strava/workouts.py
Ri087: `time_data` is the `time` stream: each element is the elapsed seconds from activity start for the sample at that index. All streams in a `StreamSet` are index-aligned and equal length, so `data[i]` of any metric matches `time[i]` — hence `recorded_at = start + time[i]s`. This holds regardless of `resolution`: a lower resolution returns fewer points, but the `time` stream still gives the correct per-index offset, so we never assume a fixed (e.g. 1 Hz) rate. We don't pass `resolution`, so Strava returns its default.

backend/app/services/providers/strava/workouts.py
Ri087: I kept `len(data)` on purpose: `original_size` is the pre-downsampling size, so if a stream is ever returned at a reduced `resolution` (`< high`), `original_size > len(data)` and indexing by it would overrun. `len(data)` is correct in both cases. We don't request a reduced resolution today so they're equal in practice — but `len(data)` is the safe bound. Happy to switch if you have a case where `original_size` is preferable.

backend/app/services/providers/strava/workouts.py
Ri087: `_make_api_request` already prepends `self.api_base_url`; the `/api/v3/...` prefix is the path arg here, matching the three existing call sites in this file (`/api/v3/athlete/activities`, `/api/v3/activities/{id}`) which carry the `# hard-coded value - update with base template changes` note. I kept the streams call consistent with that. If you'd rather route the version through a property, I'll do it — but as a separate sweep over all four call sites so they stay consistent. Which do you prefer?

backend/app/services/providers/strava/workouts.py
Ri087: The cost is inherent (time points × series), so a comprehension wouldn't change complexity — but I can rewrite the inner build as one if you find it cleaner. Want me to?

backend/tests/providers/strava/test_strava_workout_samples.py
Ri087: Which ones do you see as redundant? Happy to trim. The set covers: sample mapping (counts / offsets / values), the empty / non-dict / short-stream guards, the flag-off no-call paths, and failure isolation (fetch / persist raising → swallowed). I can collapse the guard cases with `parametrize` if that's the concern.

backend/app/constants/series_types/strava.py
```diff
@@ -0,0 +1,16 @@
...
+STREAM_KEY_SERIES_TYPE: dict[str, SeriesType] = {
+    "heartrate": SeriesType.heart_rate,
+    "velocity_smooth": SeriesType.speed,
+    "cadence": SeriesType.cadence,
+    "watts": SeriesType.power,
+}
+
+# Value for the Strava streams `keys` query param: the time axis first, then
+# every metric stream we ingest.
+STREAM_KEYS_PARAM: str = ",".join(["time", *STREAM_KEY_SERIES_TYPE])
```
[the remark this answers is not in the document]
Ri087: Updated `docs/providers/coverage.mdx`: Strava **Timeseries**, `heart_rate`, `cadence` and `power` now marked ◐ (per the legend — "available only within workout/activity context"). **24/7 Data** stays ❌. Added a Strava accordion note that these per-sample streams are flag-gated behind `ingest_workout_samples` and scoped to workouts, not continuous 24/7.

backend/app/services/providers/strava/workouts.py
czajkub: fair point, didn't notice that part

backend/app/services/providers/strava/workouts.py
czajkub: nvm, i was thinking of a separate issue, its fine this way

backend/app/services/providers/strava/workouts.py
czajkub: okay, got it. makes sense

backend/app/services/providers/strava/workouts.py
czajkub: in that case it's good, i just thought these two things were equivalent, my bad

czajkub:
looks good to me now, one thing though - can you look at #1082 implementing samples for Garmin and add the check from settings for sample ingestion? im mainly talking about that check:
```python
if settings.ingest_workout_samples:
```

When this will be added, ill merge this, or if you prefer I can add this myself. Thank you for the contribution!

KaliszS:
Hi @Ri087. Have you seen @czajkub comment? Do you still plan to proceed this PR?

backend/app/services/providers/strava/workouts.py
```diff
@@ -1,21 +1,30 @@
 from datetime import datetime, timedelta, timezone
 from decimal import Decimal
-from typing import Any, Iterable
+from typing import Any
 from uuid import UUID, uuid4
 
 from app.config import settings
+from app.constants.series_types.strava import STREAM_KEY_SERIES_TYPE, STREAM_KEYS_PARAM
```
KaliszS, line 7: It should be based on coverage.py

KaliszS (changes requested):

backend/app/services/providers/strava/workouts.py
KaliszS: @coderabbitai we can't commit our changes to this PR and contributor mosy likely won't respond, so I'm going to merge this one, but can you create an issue to cover my comment in another PR?

KaliszS (approved):

PR #1109, opened 2026-06-06
feat(integrations): add SensorBio provider

sontakey:
## Summary
- Adds SensorBio OAuth provider wiring and factory registration.
- Adds SensorBio workout, sleep, recovery, and biometric data mapping.
- Adds SensorBio workout type mapping, provider icon, config/env placeholders, and focused provider tests.
- Rebases branch onto current upstream `main` and aligns imports with the current schema layout.

## Verification
- `ENVIRONMENT=test SECRET_KEY=test-secret uv run pytest -q` ✅ **2,184 passed**
- `ENVIRONMENT=test SECRET_KEY=test-secret uv run pytest tests/providers/sensorbio tests/providers/test_provider_factory.py tests/providers/test_provider_coverage.py -q` ✅ **149 passed**
- `uv run ruff check` ✅
- `uv run ruff format --check` ✅ 611 files already formatted
- `uv run ty check` ✅
- Coverage docs regenerate idempotently and `uv lock --check` passes.

## Redacted live import-granularity evidence
The screenshot below was generated from a live Sensor Bio API verification run. User identifiers, API credentials, and biometric values are excluded; only aggregate counts, field names, and observed timestamp cadence are shown.

Sensor Bio import granularity evidence

## Notes
- The branch now includes current upstream `main`; merge conflicts in `.env.example` and `uv.lock` are resolved.
- The redacted verification user had no `/v1/scores` records in the current 30-day window. Recovery, sleep, and activity score imports remain covered by schema and contract tests.
- Independent security/API review found no leaked secrets, auth-header logging, or non-Sensor-Bio API domains.

## Summary by CodeRabbit

* **New Features**
  * Added Sensor Bio provider: OAuth sign‑in, configurable credentials, and syncing for workouts, sleep (with stages), recovery, continuous biometrics, and daily activity stats; workout-type normalization and HTTP/2-enabled API requests.

* **Tests**
  * Added unit/integration tests and a manual E2E harness covering OAuth, provider wiring, data normalization, pagination, and HTTP/2 behavior.

* **Documentation**
  * New Sensor Bio integration guide and updated provider coverage matrix.

* **Chores**
  * Example env updated with Sensor Bio variables and HTTP client HTTP/2 enabled.


sontakey:
## CodeRabbit Review Cleanup — commit e739177

All 9 actionable CodeRabbit comments addressed. Findings vs. current code:

### Fixed (5 real issues)

**`data_247.py` — sleep_sessions_synced overcount (batch 1)**
`save_sleep_data` now returns `bool` (True = persisted, False = skipped/failed). `load_and_save_sleep` only increments `sleep_sessions_synced` when the return value is True. Previously it counted every successfully-normalized record regardless of whether the DB write succeeded.

**`data_247.py` — activity_samples_synced / body_measurement_samples_synced always zero (batch 1)**
`load_and_save_all` declared both counters but called neither pathway. Implemented `save_activity_samples` (bulk timeseries insert for HR, HRV RMSSD, SpO2, respiratory rate) and wired it in `load_and_save_all`. Removed `body_measurement_samples_synced` — no Sensor Bio endpoint for body measurements exists.

**`test_sensorbio_api_spec.py` — function-scoped imports (batch 2)**
`from app.schemas.enums import WorkoutType` and `from app.services.providers.api_client import make_authenticated_request` moved to module-level imports; function-local imports removed.

**`manual_test_sensorbio_app.py` — missing type annotations on nested mocks (batch 3)**
All 5 nested mock callbacks (`_mock_api`, `_capture_cursor`, `_capture`, `_mock_sleep`, `_mock_scores`, `_mock_steps`) annotated with `(db: Any, user_id: Any, endpoint: str, **kwargs: Any) -> dict[str, Any]`.

**Docs — SensorBio missing from coverage matrix and provider pages (batches 1 & 2)**
- `docs/providers/coverage.mdx`: added Sensor Bio column to Quick Overview table and Heart & Cardiovascular detail matrix (HR, resting HR, HRV RMSSD columns all marked ✅).
- `docs/providers/sensorbio-api-integration.mdx`: new provider page with supported data types, OAuth flow, sync behavior per endpoint, and a Provider Quirks accordion documenting the continuous 24/7 timeseries behavior and epoch-seconds vs. epoch-milliseconds inconsistency.

### Already resolved before this run (1 item)

**`strategy.py` — `__init__` missing `-> None`**: already fixed in commit `aade34f` on the branch. No change needed.

---

**Test run:** `uv run pytest tests/providers/sensorbio tests/providers/test_provider_factory.py -q` → **52 passed**

sontakey:
Ready for review.

Cleanup pass is done:
- Addressed the remaining CodeRabbit feedback.
- Removed the manual/browser tester scaffolding from the PR.
- Full backend test suite passes locally: 1809 passed, 2 skipped.
- SensorBio-focused tests pass: 72/72.
- Ruff is clean.
- Current PR head: 7346a78.

The PR is mergeable with passing checks; it is only blocked on human review approval now.

sontakey:
@KaliszS , @bartmichalak , @czajkub, ready for review.

KaliszS:
Hi @sontakey. Do you own any SensorBio wearable and have you confirmed this integration to work on your data?

sontakey:
Yes, I'm a co-founder. It works on my data.

bartmichalak:
@sontakey that's great, so I'm pretty sure you were able to verify that all the expected data is in place 😀 Nice to meet you, and we're glad you thought it'd be cool to see SensorBio among the supported integrations in OW 

Thanks for this contribution! We definitely want to merge it, but right now we're super busy with a bunch of other urgent topics. This type of submission where a new integration is added takes a lot more effort on our end than a typical PR (we need to understand how the ecosystem works, what data types are available, what the constraints are, and so on), so we need a bit more time. Hope you understand!

sontakey:
@bartmichalak nice to meet you! Happy to set up a call so we can chat. check out our site if you haven't already: [https://sensorbio.com](https://sensorbio.com)

sontakey:
@bartmichalak updated PR. Ready to merge when you are :)

backend/app/constants/workout_types/sensorbio.py
```diff
@@ -0,0 +1,59 @@
+import logging
+
+from app.schemas.enums import WorkoutType
+
+logger = logging.getLogger(__name__)
+
+SENSORBIO_NAME_TO_WORKOUT_TYPE: dict[str, WorkoutType] = {
+    "run": WorkoutType.RUNNING,
+    "running": WorkoutType.RUNNING,
```
czajkub, line 8: Are `run` and `running` separate entities detected in SensorBio? What would be the difference between `run` and `running`, or between `hike` and `hiking`?

backend/app/constants/workout_types/sensorbio.py
```diff
@@ -0,0 +1,59 @@
...
+    for candidate in (likely_name, activity_type):
+        if not candidate:
+            continue
+        normalized = candidate.lower().strip()
+        if normalized in SENSORBIO_NAME_TO_WORKOUT_TYPE:
+            return SENSORBIO_NAME_TO_WORKOUT_TYPE[normalized]
+    # Log unknown names so novel SensorBio activity types are discoverable.
+    raw_name = likely_name or activity_type
+    if raw_name:
+        logger.info(
```
czajkub: We should rather refrain from using raw logger calls, more standardized way accross the project is to use `log_structured` - but I'm not sure whether logging this is necessary, if you believe so you can leave it

backend/app/schemas/providers/sensorbio/models.py
```diff
@@ -0,0 +1,232 @@
...
+    def coerce_score(cls, v: Any) -> Any:
+        if v == {} or v is None:
+            return None
+        return v
+
+
+class RecoveryInScores(BaseModel):
+    """Recovery sub-object inside a /v1/scores response.
+
+    Real shape confirmed live (t_4841e021): data.recovery.value is the 0-100
```
czajkub: Do we need those comments about confirming the shapes live? I am happy that this was tested and verified, however I don't think this needs explanation in code.

Also - if you're certain that a value is between `0-100`, you can use Pydantic `Field`s to set min/max value of them (with `ge`/`le` attributes iirc)

backend/app/schemas/providers/sensorbio/models.py
```diff
@@ -0,0 +1,232 @@
...
+    resting_bpm: float | None = None
+    resting_hrv: float | None = None
+    hrv: float | None = None
+    spo2: float | None = None
+
+
+class ActivityInScores(BaseModel):
+    """Activity sub-object inside a /v1/scores response.
+
+    Real shape confirmed live (t_5912f22f): data.activity.value is the 0-100
```
czajkub: Another comment same way as above - IMO not necessary, also the `0-100` concern

backend/app/services/providers/sensorbio/data_247.py
```diff
@@ -0,0 +1,869 @@
...
+    # -------------------------------------------------------------------------
+
+    def get_sleep_data(
+        self, db: DbSession, user_id: UUID, start_time: datetime, end_time: datetime
+    ) -> list[dict[str, Any]]:
+        all_sleep_data: list[dict[str, Any]] = []
+        current_date = start_time.astimezone(timezone.utc).date()
+        end_date = end_time.astimezone(timezone.utc).date()
+
+        while current_date <= end_date:
```
czajkub, line 131: Is there no way to query the sleep data with a range of days instead of 1 query per day? It seems like a lot of additional overhead for longer lookups

backend/app/services/providers/sensorbio/data_247.py
```diff
@@ -0,0 +1,869 @@
...
+        return {
+            "id": uuid4(),
+            "user_id": user_id,
+            "provider": self.provider_name,
+            "timestamp": start_dt or end_dt,
+            "start_time": start_dt,
+            "end_time": end_dt,
+            "duration_seconds": duration_seconds,
+            "efficiency_percent": score.value if score else None,
+            "is_nap": duration_seconds < 3 * 3600 if duration_seconds else False,
```
czajkub: Is this a separation made in SensorBio? I do recognise that some wearables like Oura detect a nap by `< 3hrs` sleep, but is it done the same way in SensorBio?

backend/app/services/providers/sensorbio/data_247.py
```diff
@@ -0,0 +1,869 @@
...
+    # -------------------------------------------------------------------------
+
+    def get_recovery_data(
+        self, db: DbSession, user_id: UUID, start_time: datetime, end_time: datetime
+    ) -> list[dict[str, Any]]:
+        all_scores: list[dict[str, Any]] = []
+        current_date = start_time.astimezone(timezone.utc).date()
+        end_date = end_time.astimezone(timezone.utc).date()
+
+        while current_date <= end_date:
```
czajkub, line 299: here the same thing with the per day loop

backend/app/services/providers/sensorbio/data_247.py
```diff
@@ -0,0 +1,869 @@
...
+            try:
+                sample = TimeSeriesSampleCreate(
+                    id=uuid4(),
+                    user_id=user_id,
+                    source=self.provider_name,
+                    recorded_at=timestamp,
+                    value=Decimal(str(value)),
+                    series_type=series_type,
+                )
+                timeseries_service.crud.create(db, sample)
```
czajkub: Manually calling `.crud.create` here feels like an antipattern, we should use simple service methods when necessary (e.g. `create`).

Another thing - if this is a daily average there should be a field in `TimeSeriesSampleCreate` that specifies that fact

Also in theory, we have 3 possible samples here - we could even bulk_create with those three, its still less calls to db than now

backend/app/services/providers/sensorbio/data_247.py
```diff
@@ -0,0 +1,869 @@
...
+            except Exception as e:
+                log_structured(
+                    self.logger,
+                    "warning",
+                    f"Failed to save sleep health score from /v1/scores: {e}",
+                    provider="sensorbio",
+                    task="save_recovery_data",
+                )
+
+        return count
```
czajkub: Here, we are counting scores and biometrics the same way - maybe we could differentiate in some way how many scores were saved and how many metrics? and then after sync, the info could be propagated in logs, would be easier to debug in the future

backend/app/utils/structured_logging.py
```diff
@@ -99,4 +100,7 @@ def log_structured(
     # IMPORTANT: Celery workers and other services must redirect stderr to stdout
     # in their startup scripts (using `exec 2>&1`) to prevent platforms from
     # converting all logs to level.error. See scripts/start/*.sh for examples.
-    print(json_str, file=sys.stdout, flush=True)
+    with contextlib.suppress(BrokenPipeError, OSError):
```
czajkub: What is the context of this change, has this occured to you? Could you explain why this happened?

Also, isnt `BrokenPipeError` a subcclass of `OSError`?

backend/app/config.py
```diff
@@ -144,6 +144,12 @@ class Settings(BaseSettings):
     whoop_redirect_uri: str | None = None  # Deprecated: use API_BASE_URL
     whoop_default_scope: str = "offline read:cycles read:sleep read:recovery read:workout"
 
+    # SENSORBIO OAUTH SETTINGS
+    sensorbio_client_id: str | None = None
+    sensorbio_client_secret: SecretStr | None = None
+    sensorbio_redirect_uri: str | None = None  # Deprecated: use API_BASE_URL
```
czajkub: if this is deprecated, it doesn't need to be added here. just leave it up to `API_BASE_URL`.

backend/app/config.py
```diff
@@ -144,6 +144,12 @@ class Settings(BaseSettings):
...
+    sensorbio_default_scope: str = ""
```
czajkub, line 161: why is it empty? shouldnt we leave here scopes for what Open Wearables would extract instead of leaving it like this?

backend/config/.env.example
```diff
@@ -157,5 +157,10 @@ FITBIT_DEFAULT_SCOPE=activity heartrate sleep profile
 
 #--- Ultrahuman ---#
 ULTRAHUMAN_CLIENT_ID=your-ultrahuman-client-id
-ULTRAHUMAN_CLIENT_SECRET=your-ultrahuman-client-secret
+ULTRAHUMAN_CLIENT_SECRET=your-u...cret
```
czajkub: why the change here?

backend/config/.env.example
```diff
@@ -157,5 +157,10 @@ FITBIT_DEFAULT_SCOPE=activity heartrate sleep profile
...
 ULTRAHUMAN_DEFAULT_SCOPE=ring_data cgm_data profile
+
+#--- Sensor Bio ---#
+SENSORBIO_CLIENT_ID=your-sensorbio-client-id
+SENSORBIO_CLIENT_SECRET=your-sb...cret
+SENSORBIO_DEFAULT_SCOPE=
```
czajkub, line 173: also the scope issue here from above

backend/scripts/sensorbio_integration_tester.py
czajkub, line 1: I didn't read this file, what is this script exactly testing? Is this an end-to-end integration test? Is this supposed to be a part of the repo, or was this a mock for testing the changes?
Also, if this was used for testing, was the integration tested with real, live data from SensorBio or only using this mock?

backend/pyproject.toml
```diff
@@ -80,6 +81,8 @@ ignore = [
 "app/api/routes/v*/*.py" = ["ANN201"] # missing-type-annotation
 "app/schemas/providers/garmin/*.py" = ["N815"]  # mixedCase from external API
 "app/schemas/providers/suunto/*.py" = ["N815"]  # mixedCase from external API
+# Integration tester: HTML template strings have intentionally long lines; not production code
```
czajkub: if its not production code should it be included in the repo?

backend/pyproject.toml
```diff
@@ -45,6 +45,7 @@ dev = [
     "factory-boy>=3.3.0",
     "psycopg-binary>=3.3.4",
     "testcontainers[postgres,redis]>=4.14.2",
+    "h2>=4.3.0",
```
czajkub: what is the `h2` package added here?

czajkub (changes requested):
Hi, thank you very much for the contribution!

I left some comments in the review, also some few issues not included there:

- This change might be after the PR was already created, but now we use `coverage.py` modules in provider services as a source of truth for the data types that are covered by providers. You could check how it's done for current providers after rebase/merge. After this, a pre-commit hook should regenerate the coverage matrix accordingly.
- If possible, could you provide some screenshots with the granularity of data / all data types imported from SensorBio?

backend/app/constants/workout_types/sensorbio.py
sontakey: Not separate entities — just string aliases.

Sensor Bio’s `likely_name` / parent workout name can come back as either form (`"Run"` vs `"Running"`, `"Hike"` vs `"Hiking"`). Both map to the same unified `WorkoutType`. Kept both keys so we don’t miss a form; there’s no distinct Sensor Bio activity type behind each spelling.

backend/app/constants/workout_types/sensorbio.py
sontakey: Agreed — logging wasn’t necessary. Removed the OTHER-workout logger entirely.

backend/app/schemas/providers/sensorbio/models.py
sontakey: Removed the live-shape confirmation comments.

Also added Pydantic `Field(ge=0, le=100)` on the 0–100 score value fields (`RecoveryInScores.value`, `ActivityInScores.value`, `SleepInScores.value`, `ScoreInSleep.value`).

backend/app/schemas/providers/sensorbio/models.py
sontakey: Same fix as above — comments removed and `Field(ge=0, le=100)` applied.

backend/app/services/providers/sensorbio/data_247.py
sontakey: Unfortunately no — confirmed against the current Sensor Bio OpenAPI:

`GET /v1/sleep` only accepts a single required `date` (YYYY-MM-DD). Session-level sleep is always date-scoped; a day may return multiple sessions, but there’s no multi-day range query for that payload.

Left the per-day loop and documented the constraint in code + the integration doc. (There are granular *summary* endpoints for sleep details, but they don’t replace session-level `/v1/sleep`.)

backend/app/services/providers/sensorbio/data_247.py
sontakey: Good catch — **no**, Sensor Bio does not classify naps.

The API can return multiple sleep sessions for a day (e.g. overnight + nap) but has **no** `is_nap` / session-type field. The `< 3h` heuristic was ours, not theirs.

Changed to always set `is_nap=False` and documented that. Multi-session days still persist as separate sleep records.

backend/app/services/providers/sensorbio/data_247.py
sontakey: Same constraint as sleep: `GET /v1/scores` requires a single `date`. No range query for the top-level activity/sleep/recovery scores payload. Documented in code.

backend/app/services/providers/sensorbio/data_247.py
sontakey: Fixed:
1. Switched recovery metric writes to `timeseries_service.bulk_create_samples(...)` (service method, one call for all present metrics).
2. Health scores use `health_score_service.bulk_create(...)`.
3. Daily step/energy/distance samples now set `is_daily_total=True` (those are true daily totals from `/v1/step/details`). Recovery biometric snapshots are not summable dailies, so they leave `is_daily_total` unset.

backend/app/services/providers/sensorbio/data_247.py
sontakey: Done. `save_recovery_data` / `load_and_save_recovery` now return and log:

- `metrics_synced` — timeseries samples written
- `scores_synced` — HealthScore rows written

`load_and_save_all` exposes `recovery_metrics_synced` + `recovery_scores_synced` separately (instead of a single blended counter).

backend/app/utils/structured_logging.py
sontakey: Unrelated leftover from local tunnel/tester work — stdout sometimes closed under a detached Cloudflare tunnel and the swallow was masking that. **`BrokenPipeError` is indeed a subclass of `OSError`**, so listing both was redundant.

Reverted the change; `structured_logging` is back to plain `print(...)`.

backend/app/config.py
sontakey: Removed. Redirect URI is now derived only from `API_BASE_URL` (`{api_base_url}/api/v1/oauth/callback/sensorbio`).

backend/app/config.py
sontakey: Sensor Bio OAuth currently has **no granular scopes** defined in the developer portal (docs UI shows “No Scopes Defined”). Empty `default_scope` means the authorize URL omits `scope=` — same pattern as Garmin/Suunto.

Left empty with an explicit comment in `config.py` + `.env.example`. Happy to fill in real scopes if/when the portal exposes them.

backend/config/.env.example
sontakey: That was an accidental truncation of the Ultrahuman placeholder secret (`your-u...cret`). Restored to `your-ultrahuman-client-secret`. Sensor Bio placeholders cleaned up the same way.

backend/config/.env.example
sontakey: Same as above — empty because Sensor Bio has no granular scopes yet. Comment added in `.env.example`.

backend/scripts/sensorbio_integration_tester.py
sontakey: It was a local manual QA harness (browser UI + mock/live modes) used while validating the integration against real Sensor Bio data. **Not production code and not meant to ship.**

Removed from the PR. Automated coverage is the `tests/providers/sensorbio/` suite (127 selected tests green on this push), and the integration was also verified live against my own Sensor Bio data earlier in the PR discussion.

backend/pyproject.toml
sontakey: Agreed — removed the `scripts/sensorbio_integration_tester.py` ruff per-file ignore and deleted the script from the PR.

backend/pyproject.toml
sontakey: `h2` is the HTTP/2 protocol stack pulled in by `httpx[http2]` (required because Sensor Bio rejects HTTP/1.x with 464). It doesn’t need an explicit top-level dev dependency — removed the redundant `h2>=4.3.0` entry from `pyproject.toml` (still comes transitively via `httpx[http2]`).

sontakey:
## Addressed review feedback — `09102c1`

Thanks @czajkub — integrated the inline comments + the two top-level notes. Summary:

### Code changes
| Area | Change |
|------|--------|
| **coverage.py** | Added `backend/app/services/providers/sensorbio/coverage.py` and wired `SensorBioStrategy.coverage` (TIMESERIES / WORKOUT_FIELDS / SLEEP_FIELDS / HEALTH_SCORES). Mapping tables imported by `data_247.py` so coverage can’t drift. |
| **Dev scaffolding** | Removed `sensorbio_integration_tester.py` + its ruff ignore (not production). |
| **structured_logging** | Reverted the unrelated BrokenPipe swallow. |
| **config / env** | Dropped deprecated `sensorbio_redirect_uri` (uses `API_BASE_URL` only). Empty scopes kept with comments (Sensor Bio OAuth has no granular scopes). Fixed accidental Ultrahuman secret placeholder truncation. |
| **pyproject** | Removed redundant top-level `h2` dev dep (`httpx[http2]` already pulls it). |
| **schemas** | Stripped live-validation comments; `Field(ge=0, le=100)` on 0–100 score values. |
| **workout types** | Kept `run`/`running` etc. as string aliases (not distinct entities); removed OTHER logger. |
| **sleep/scores fetch** | Per-day loops retained — official API requires single `date` (no range). Documented in code + docs. |
| **is_nap** | No longer inventing a `<3h` heuristic; Sensor Bio has no nap flag → `is_nap=False`. |
| **save paths** | `timeseries_service.bulk_create_samples` + `health_score_service.bulk_create`; daily steps/energy/distance set `is_daily_total=True`. |
| **sync counters** | Split `recovery_metrics_synced` vs `recovery_scores_synced` (logged + returned). |

### Tests
```
uv run pytest tests/providers/sensorbio tests/providers/test_provider_factory.py tests/providers/test_provider_coverage.py -k 'sensorbio or coverage'
127 passed
```
Ruff clean on touched paths.

### Top-level notes
1. **coverage.py** — done in this commit (source of truth for meta coverage / matrix).
2. **Screenshots of data granularity** — happy to attach next. Live validation was done against my own Sensor Bio account (workouts, multi-session sleep days, recovery/activity/sleep scores, continuous biometrics, daily steps). If you want a specific view (e.g. OW UI after sync vs raw API payloads), tell me which and I’ll post them.

Ready for another pass when you have time.

sontakey:
Updated this PR against current `main` and completed the outstanding review work:

- merged upstream `main` and resolved the `.env.example` / `uv.lock` conflicts
- regenerated the provider coverage docs
- fixed the new `ty` generic-bound issue in the Sensor Bio schema parser
- resolved all 17 review threads after confirming each requested change or API constraint was addressed
- added redacted live import-granularity evidence to the PR description

Verification on the updated head:

- full backend suite: **2,184 passed**
- Sensor Bio/provider suite: **149 passed**
- Ruff: passed
- format check: passed
- `ty check`: passed
- coverage-doc generation: idempotent
- `uv lock --check`: passed

GitHub now reports the branch as mergeable. The remaining blocker is the existing `CHANGES_REQUESTED` review state.

@czajkub, could you please re-review the updated head when you have a chance?

KaliszS (approved):
Thanks for all your hard work! We don't have a device to test it on our own, but the code looks fine. Approved and wolecome to the Open Wearables family!

sontakey:
Thank you! We shipped out a device to Anna.

Issue #424, opened 2026-02-03
Implement centralized unit conversion service


bartmichalak:
We store metrics in canonical units (e.g., height in cm, distance in meters) but have no proper conversion layer. Current state
- unit field from incoming requests is ignored                                                                                                                                                     
- Hardcoded conversion only for height and body_fat_percentage in import_service.py:104-106  
- No validation that incoming unit matches expected unit
- If a provider sends data in unexpected units → silently stored incorrectly                                                                                                                                                                                   
                                                                                                                                                                                                                                                                 
## Scope                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                 
Affects all providers, not just Apple HealthKit.                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                 
**Proposed solution**                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                            - Create app/services/unit_conversion.py:                                                                                                                                                                                                                        
- Map (series_type, source_unit) → converter_fn                                                                                                                                                                                                                - Validate incoming units against expected values per metric type                                                                                                                                                                                              - Single source of truth for canonical units (already have SERIES_TYPE_DEFINITIONS)

MichalZak17:
Hey @bartmichalak @czajkub I put together a concept implementation for this under #1120. Essentially I built `app/services/unit_conversion.py` as the single source of truth and migrated the Apple HealthKit `_build_statistic_bundles` height/body_fat conversions onto it. The rest of the details you can check under the PR draft. Let me know what you think here.

MichalZak17:
Coming back to this one.

Back in March the read was that most data comes through in metric anyway, so it'd probably be fine to keep handling units inline. I went and checked how that's actually played out, and I based on what I see the inline approach costs more than it looks.

The thing that convinced me is the fact that soneone already fixed this exact class of bug four separate times. `_normalize_unit` in the healthkit import service currently special-cases height, walking_step_length, body_fat, and a few walking percentages, and each of those lines up with a bug someone filed and we patched one at a time

- #351 height stored in meters, labeled cm
- #1106 walking_step_length stored in meters, labeled cm
- #353 body fat stored as a 0-1 fraction, labeled percent
- #1105 walking_double_support stored as a fraction, labeled percent

And there are three more open right now that are the same shape, where the import path stores the value and ignores the unit the record came with:

[...]

PR #1120, opened 2026-06-09
feat(backend): centralized unit conversion service (concept for #424)

MichalZak17:
**Concept implementation for #424 — for discussion/verification, not a finished PR.**

## Scope
Builds \`app/services/unit_conversion.py\` as a single source of truth for converting incoming provider values to canonical units, and migrates the Apple HealthKit \`_build_statistic_bundles\` height/body_fat conversions onto it as a proven reference. Garmin/Oura/Whoop/Suunto/Fitbit and the ms→s duration conversions are intentionally left as follow-ups.

## Design
- **Registry keyed by `(series_type, source_unit)`** with a **per-provider default source unit** fallback. Incoming SDK records usually omit the unit string, so this keeps the Apple-vs-HealthConnect body_fat asymmetry as *data* rather than an `if provider == "apple"` in the importer.
- **Canonical units** come from the existing `SERIES_TYPE_DEFINITIONS` / `get_series_type_unit()` — no second source of truth.
- **Unknown/unexpected unit → log + skip** the sample (no silent wrong-magnitude writes), matching the existing import convention.
- **Behaviour preserved:** `TestSDKImportUnitConversion` stays green **unchanged**; new pure unit tests cover the service directly.

## Unit assumptions verified against platform docs
- Apple body fat is a 0–1 ratio (`HKUnit percentUnit // % (0.0 - 1.0)`) → ×100 to percent.
- Health Connect `Percentage` is already 0–100 → no scaling.
- Height is meters on both Apple and Health Connect → ×100 to cm.

## Open question
Unknown-unit policy: **log + skip** (current) vs **raise**. Worth deciding before extending to the other providers.

## Follow-ups (out of scope here)
Garmin weight (g→kg), Oura/Whoop height, Suunto balance/SpO2/joules→kcal, Fitbit/Whoop/Apple duration (ms→s), plus unit-string alias/case handling and an optional strict mode.

### DIFF
--- a/backend/app/services/apple/healthkit/import_service.py
+++ b/backend/app/services/apple/healthkit/import_service.py
@@ -32,6 +32,7 @@ from app.schemas.providers.mobile_sdk import (
 from app.schemas.responses.upload import UploadDataResponse
 from app.services.event_record_service import event_record_service
 from app.services.timeseries_service import timeseries_service
+from app.services.unit_conversion import unit_conversion_service
 from app.utils.structured_logging import log_structured
 
 from .device_resolution import extract_device_info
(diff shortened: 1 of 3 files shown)


KaliszS:
Hi. As I replied to the email, we can't accept PR at this form. It break single source of truth rule. Such conversions need to be done using streategies.

Issue #1113, opened 2026-06-08
Bug: whoop app is showing workout but openwearables dashboard is not under strain data


Atul240202:
### Description

A workout is successfully recorded and visible in the WHOOP app, but it does not appear under the **Strain** data section in the OpenWearables dashboard.

### Steps to Reproduce

1. Record or complete a workout using WHOOP.
2. Confirm that the workout is visible in the WHOOP app.
3. Open the OpenWearables dashboard.
4. Navigate to the **Strain** data section.
5. Observe that the workout data is missing.

### Expected Behaviour

Workouts available in the WHOOP app should be synced and displayed under the **Strain** data section in the OpenWearables dashboard.

### Actual Behaviour

The workout appears in the WHOOP app but is not displayed under the **Strain** data section in the OpenWearables dashboard.

### Additional Information

* Data source: WHOOP
* Affected section: Strain
* Sync status: The WHOOP workout exists, but the corresponding OpenWearables strain data is missing.

bartmichalak:
@Atul240202 The Open Wearables dashboard doesn't have a Strain section, and I can confirm that workouts available in the Whoop app sync properly with OW. Could you clarify a bit more what you meant here?

bartmichalak:
Ok, I think you were referring to strain data from the Whoop API, which is covered by Whoop's not-yet-implemented endpoint, described in this issue: https://github.com/the-momentum/open-wearables/issues/1008

I'm closing this one as a duplicate. However, if the intention was different, feel free to reopen

PR #1124, opened 2026-06-09
feat(backend): add Whoop cycle endpoint for daily strain scores

### LINKED ISSUE #1008: Cycle endpoint support for Whoop
Whoop sends daily aggregates, for example daily strain score (even if there are no workouts), and certain things like sleep sessions are associated with cycles. 

We will need to add the `/v2/cycle`  endpoint support to `app/services/providers/whoop/data_247.py` module. It is very similar to other endpoint returning data, so most of that is just connecting the endpoint and copying some code (or preferably separating a part of the code and making it reusable).

knowald:
## Description

Adds `/v2/cycle` support to the Whoop provider, as requested in #1008. Whoop cycles carry the daily strain score (present even on days without workouts), which was previously never ingested - only per-workout strain was stored.

- `get_cycle_data` fetches cycles with pagination from `/v2/cycle`
- `normalize_cycle` maps a scored cycle to a `HealthScoreCreate` with `category=STRAIN`, `qualifier="daily"`, the cycle's `timezone_offset` as `zone_offset`, and kilojoule / average_heart_rate / max_heart_rate as score components. The qualifier separates day-level cycle strain from the per-workout strain scores `workouts.py` already emits under the same `(provider, category)` - without it, consumers could not tell the two apart.
- `load_and_save_cycles` is wired into `load_and_save_all` (new `cycle_scores_synced` result key)
- Per the issue's suggestion to make the copied code reusable, the identical pagination loops in `get_sleep_data` and `get_recovery_data` were extracted into `_get_paginated_records`, which the new cycle fetch also uses
- `docs/providers/coverage.mdx`: updated the stale "Strain scores available but not yet implemented" line in the Whoop accordion

In-progress cycles (no `end` yet) are skipped: their strain is provisional, and since health score inserts dedupe via `on_conflict_do_nothing`, saving a partial value would lock it in permanently. The completed cycle is picked up on the next sync.

No scope changes needed - `read:cycles` is already in `whoop_default_scope`.

Resolves #1008

### General

### Backend Changes

- [x] `uv run pre-commit run --all-files` passes (ruff + ty pass; prettier hook needs frontend `node_modules`, no frontend files touched)

## Testing Instructions

**Steps to test:**
1. `cd backend && uv run pytest tests/providers/whoop/`
2. Connect a Whoop account and trigger a 247 sync.
3. Check health scores for the user: daily `strain` entries (qualifier `daily`) should appear for each completed cycle, alongside `recovery`.

**Expected behavior:**

One strain health score per completed, scored cycle, with kilojoule and heart rate values as components and the cycle's timezone offset preserved. Unscored or in-progress cycles are skipped. Re-syncing the same window does not duplicate scores (covered by a test).

## Additional Notes

- This should also help with #1113 (strain visible in Whoop app but missing from the dashboard).
- Merge-order heads-up: open PR #1092 (transaction boundaries to callers) rewrites every `db.commit()`/`db.rollback()` in this file to the savepoint/flush pattern. Whichever of the two merges second needs a small rebase; if #1092 lands first I will adapt the cycle block to `begin_nested()`/`flush()`.
- Known limitation, intentionally out of scope: a cycle that is still `PENDING_SCORE` when a sync window passes over it can be missed if it scores only after the next window starts. Padding the cycle fetch window behind `last_synced_at` (dedupe makes re-fetching idempotent) would close that gap; happy to do it here or in a follow-up if desired.
- Related limitation: if Whoop re-scores an already-completed cycle (e.g. after late device data upload), the first saved strain stays in place because health score inserts use `on_conflict_do_nothing`. Fixing that means switching the dedupe to `on_conflict_do_update`, which changes shared `HealthScoreRepository.bulk_create` behavior for all providers - deliberately not done in this PR. The consumer-facing semantics of `daily` strain scores are documented in the `qualifier` field description and `coverage.mdx`.

## Summary by CodeRabbit

* **New Features**
  * Added Whoop cycle syncing and daily strain score support with paginated fetching and partial-result recovery on downstream errors.

* **Tests**
  * Added tests for cycle normalization, pagination behavior, error recovery, persistence, and idempotent re-syncing to prevent duplicates.

* **Documentation**
  * Updated provider docs and rate-limit guidance to reflect cycle/strain syncing and adjusted request estimates.

### DIFF
--- a/backend/app/schemas/model_crud/activities/health_score.py
+++ b/backend/app/schemas/model_crud/activities/health_score.py
@@ -25,7 +25,13 @@ class HealthScoreBase(BaseModel):
         None,
         description="Overall numeric score. Range varies by provider and category — see HEALTH_SCORE_RANGES for scale.",
     )
-    qualifier: str | None = Field(None, description="Textual rating from the provider, e.g. GOOD or EXCELLENT")
+    qualifier: str | None = Field(
+        None,
+        description=(
+            "Textual rating from the provider, e.g. GOOD or EXCELLENT. "
+            "For Whoop strain, 'daily' marks day-level cycle strain vs per-workout strain (null)."
+        ),
+    )
     recorded_at: datetime
     zone_offset: ZoneOffset = None
     components: dict[str, ScoreComponent] | None = None
--- a/backend/app/services/providers/whoop/data_247.py
+++ b/backend/app/services/providers/whoop/data_247.py
@@ -37,7 +37,7 @@ class Whoop247Data(Base247DataTemplate):
         provider_name: str,
         api_base_url: str,
         oauth: BaseOAuthTemplate,
-    ):
+    ) -> None:
         super().__init__(provider_name, api_base_url, oauth)
         self.event_record_repo = EventRecordRepository(EventRecord)
         self.data_source_repo = DataSourceRepository(DataSource)
@@ -964,8 +964,12 @@ class Whoop247Data(Base247DataTemplate):
             user_id=user_id,
             provider=ProviderName.WHOOP,
             category=HealthScoreCategory.STRAIN,
+            # Whoop also emits per-workout strain scores (workouts.py); the qualifier
+            # separates the day-level cycle strain from those.
+            qualifier="daily",
             value=strain,
             recorded_at=recorded_at,
+            zone_offset=raw_cycle.get("timezone_offset"),
             components=components or None,
         )
 
--- a/backend/app/services/providers/whoop/data_247.py
+++ b/backend/app/services/providers/whoop/data_247.py
@@ -30,7 +30,7 @@ from app.utils.structured_logging import log_structured
 
 
 class Whoop247Data(Base247DataTemplate):
-    """Whoop implementation for 247 data (sleep, recovery, activity)."""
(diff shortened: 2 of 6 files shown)


knowald:
Addressing the three CodeRabbit findings from the review body:

- `__init__` missing `-> None`: already fixed in this PR's diff.
- Naive `datetime.now()` defaults in `load_and_save_all` (lines 467-470): that block predates this PR and is untouched by it. On closer inspection the behavior claim does not hold either - the only consumer is `.astimezone(timezone.utc)` formatting, which interprets naive as system-local, exactly what `datetime.now()` produces, so the window is correct on any host timezone. Still worth making aware-from-construction as hygiene; shipped separately as #1157 with details.
- `cycle_scores_synced` counting deduped rows: intentional and documented in the method docstring ("rows already present ... are still counted here"). Every other counter in this file has the same semantics (`load_and_save_activity` and body measurement return `len(samples_to_create)`), so changing only the cycle field would make the sync report internally inconsistent. If inserted-row counts are wanted, that is a file-wide (or provider-wide) change better tracked as its own issue.

knowald:
Closing this PR for now. If a concrete need for this arises we can revisit it through a new issue.

Issue #1078, opened 2026-05-27
bug(garmin): heart_rate_min stores average instead of minimum


KaliszS:
`WorkoutDetails.heart_rate_min` is populated with the average heart rate value instead of the minimum.

**Location:** `backend/app/services/providers/garmin/workouts.py`, `_build_metrics()`:

```python
"heart_rate_min": int(heart_rate_avg) if heart_rate_avg is not None else None,
```

The field should use `minHeartRateInBeatsPerMinute` from the Garmin activity payload, which is available alongside `averageHeartRateInBeatsPerMinute` and `maxHeartRateInBeatsPerMinute`.

PR #1133, opened 2026-06-10
fix(backend): null workout heart_rate_min rows copied from average HR

knowald:
## Description

Follow-up to #1121 (issue #1078). That PR stopped writing the average heart rate into `workout_details.heart_rate_min` for Garmin workouts, but every row created before the fix still carries the copied average. Polar had the identical bug (`heart_rate_min = int(hr_avg)`) until the refactor in #1041, so its pre-refactor rows are equally affected.

Adds `backend/scripts/data_migrations/null_bogus_workout_heart_rate_min.py` following the existing convention in that directory (`--dry-run` preview, sample listing, single guarded UPDATE).

Selection is deliberately narrow:

- `provider = 'garmin' AND heart_rate_min = heart_rate_avg` - Garmin averages are integers in the API (`averageHeartRateInBeatsPerMinute: int`), so every bogus row matches exactly. Exact equality (rather than `floor`) also protects SDK-imported rows attributed to provider `garmin` via source inference that carry a genuine HealthKit minimum.
- `provider = 'polar' AND heart_rate_min = FLOOR(heart_rate_avg)` - Polar averages may be fractional, matching the old `int()` truncation.
- `category = 'workout'` only; Garmin sleep minimum HR is genuine and lives elsewhere.
- `heart_rate_avg` and `heart_rate_max` are genuine and untouched. Post-fix rows have `heart_rate_min` NULL and are not matched.

No other provider ever wrote a copied average: Suunto, Apple XML, and the mobile SDK write genuine minimums; Fitbit, Strava, Whoop, Oura, Ultrahuman never set the field.

### General

- [x] I have added tests that prove my fix/feature works (if applicable) - verified against a scratch database, see below

### Backend Changes

- [x] `uv run pre-commit run --all-files` passes (ruff passes; prettier hook needs frontend `node_modules`, no frontend files touched)

## Testing Instructions

**Steps to test:**
1. `docker compose exec app uv run python scripts/data_migrations/null_bogus_workout_heart_rate_min.py --dry-run`
2. Review the listed rows, then run without `--dry-run`.

**Expected behavior:**

Only Garmin workout rows with `heart_rate_min = heart_rate_avg` and Polar workout rows with `heart_rate_min = floor(heart_rate_avg)` get `heart_rate_min` set to NULL. Re-running reports nothing to do.

Verified against a scratch database migrated to head with seven fixture rows: Garmin bogus (nulled), Garmin distinct-min (kept), Garmin post-fix NULL (untouched), Polar bogus fractional avg (nulled), Polar distinct-min (kept), Suunto constant-HR min == avg (kept, provider filter), Garmin sleep-category row (kept). Re-run after the update found no affected rows.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected inaccurate minimum heart rate values for Garmin and Polar workout records; affected entries have been identified and cleared.
* **Chores**
  * Added an automated data-cleanup step at startup to apply the fixes. The cleanup reports affected counts and sample rows and can be run in a non-destructive mode for verification.

### DIFF
--- a/backend/scripts/data_migrations/null_bogus_workout_heart_rate_min.py
+++ b/backend/scripts/data_migrations/null_bogus_workout_heart_rate_min.py
@@ -1,0 +1,104 @@
+#!/usr/bin/env python3
+"""Null workout heart_rate_min values that actually hold the average heart rate.
+
+Garmin activity payloads carry no minimum heart rate, but until PR #1121 the
+Garmin workout normalizer wrote heart_rate_min = int(heart_rate_avg). Polar had
+the same bug until PR #1041. Both providers therefore have workout_details rows
+where heart_rate_min is a copy of the average, not a minimum.
+
+Selection is deliberately narrow:
+- Garmin averages are integers in the API, so every bogus row satisfies
+  heart_rate_min = heart_rate_avg exactly. Exact equality also protects
+  SDK-imported rows that were attributed to provider 'garmin' via source
+  inference but carry a genuine HealthKit minimum.
+- Polar averages may be fractional, so bogus rows satisfy
+  heart_rate_min = floor(heart_rate_avg) (the int() truncation).
+
+heart_rate_avg and heart_rate_max are genuine and stay untouched. Rows written
+after the fixes have heart_rate_min NULL and are not matched.
+
+Usage (inside Docker):
+    docker compose exec app uv run python scripts/data_migrations/null_bogus_workout_heart_rate_min.py --dry-run
+    docker compose exec app uv run python scripts/data_migrations/null_bogus_workout_heart_rate_min.py
+"""
+
+import argparse
+
+from sqlalchemy import text
+
+from app.database import SessionLocal
+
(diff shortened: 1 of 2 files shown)


KaliszS (changes requested):
That script should has been run by startup app.sh script. Has it been tested and proved to be indempotent?

knowald:
Good call - wired into app.sh now, same pattern as the body_fat_percentage normalization (non-fatal guard, TODO removal date ~2026-08-01).

On idempotency: yes. The UPDATE sets heart_rate_min to NULL and the predicate requires heart_rate_min IS NOT NULL, so a second run matches nothing. Verified against a scratch DB migrated to head with fixture rows for every case: first run nulled exactly the two bogus rows (Garmin min == avg, Polar min == floor(avg)) and left legit/post-fix/other-provider/sleep rows untouched; an immediate re-run reported no affected rows.

KaliszS (approved):

PR #1176, opened 2026-06-13
fix(backend): persist daily resting heart rate from sleep data

andrewle21:
## Problem

Oura's sleep endpoint returns a per-night `lowest_heart_rate` (the resting heart rate). `Oura247Data` already captures it into the normalized sleep dict (`"lowest_heart_rate": sleep.lowest_heart_rate`), but `save_sleep_data` never writes it to a series. The intraday `heart_rate` and `heart_rate_variability_rmssd` interval series and the daily `respiratory_rate` scalar are persisted, but resting heart rate is dropped.

As a result, the `resting_heart_rate` series stays empty for Oura-connected users, and `GET /api/v1/users/{user_id}/summaries/body` reports `resting_heart_rate_bpm: null` even when the user has full nightly sleep data.

## Fix

Persist `lowest_heart_rate` as a daily `resting_heart_rate` sample, mirroring the existing `average_breath` → `respiratory_rate` block in the same method (same `start_dt` / `zone_offset` / `db.commit()` / error-handling pattern).

## Verification

Tested against a live Oura account with ~90 days of ring data, self-hosted from source:

- Before: `summaries/body.averaged.resting_heart_rate_bpm` was `null`.
- After: it returns **59 bpm**, matching Oura's native `usercollection/sleep` `lowest_heart_rate` mean of **59.1** over the same 7-day window.

No schema or migration changes; `resting_heart_rate` already exists in `SeriesType` and is consumed by the body summary's averaged-vitals path.

## Summary by CodeRabbit

* **New Features**
  * Resting heart rate from Oura sleep sessions (the lowest heart rate per session) is now captured and stored for review.
  * Respiration rate from Oura sleep sessions (average breathing rate) is now captured and stored alongside your sleep metrics.
* **Documentation**
  * Updated the wearable coverage documentation to reflect the new Oura support for resting heart rate, min heart rate, and respiration rate during sleep.


KaliszS (approved):
Hi @andrewle21. I have adjusted entire code to the new meta in our streategy pattern. We store all timeseries types in coverage.py for ech provider and classes like data247 just fetch from that place.

PR #1187, opened 2026-06-16
fix(integrations): recover missing Suunto sleep windows safely

bwickner:
Suunto sleep payloads can omit the documented BedtimeStart/BedtimeEnd window across REST backfill and SLEEP_CREATED webhook ingestion. Without a usable window, normalized sessions are skipped and sync stats can make those drops look like successful ingestion.

Keep documented bedtimes as authoritative, reconstruct absent ends from entryData.DateTime plus Duration, and reject degenerate windows so only persistable sleep sessions are counted as saved. Do not infer sleep starts from the wrapper timestamp because Suunto documents it as the sample measurement timestamp.

### General

### Backend Changes

### Frontend Changes

- [ ] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1.
2.
3.

**Expected behavior:**

## Release Notes

* **Bug Fixes**
  * Improved Suunto sleep data processing by reconstructing missing sleep start or end times from available duration information.
  * Improved handling of mixed timezone formats in sleep records.
  * Sleep entries with missing or invalid time windows are now safely skipped instead of being saved incorrectly.
* **Tests**
  * Added coverage for sleep-window reconstruction, timezone handling, and invalid-duration safeguards.


bwickner:
Origin of Suunto API behaviour is not officially explained, might be device dependent - I am using a Vertical 2.

This is also publicly documented / discussed at intervals.icu, see https://forum.intervals.icu/t/suunto-wellness-data-now-integrated/91204/5

KaliszS (approved):
Hi, sorry for the late reply. I have done a small refactor, becasue `DateTime` field was always existing in our payloads (I use the same device btw). So that fallback was redundant in my opinion.

PR #1203, opened 2026-06-18
fix: Mark connection as revoked on failed token refresh + connection.revoked outgoing webhook

czajkub:
## Description

When we try to refresh token and get `400` / `401` response, the connection is still active and each periodic sync tries to refresh an inactive access token, causing constant errors. This way, the connection will be considered revoked during syncs, showing up on frontend.

Also adding `connection.revoked` outgoing webhook type to notify of that

 
### General

### Backend Changes

## Summary by CodeRabbit

* **New Features**
  * Added a new `connection.revoked` webhook event when a provider connection becomes invalid and requires re-authorization.
  * Webhook payload now includes connection identifiers plus `reason` and `revoked_at`, emitted on provider deregistration and certain refresh-token failures (HTTP 400/401).
* **Documentation**
  * Updated webhook API docs with the new `connection.revoked` event, payload schema, and an example.
* **Bug Fixes**
  * Improved error reporting by treating expected 4xx HTTP errors as client-side issues and avoiding unnecessary Sentry capture.

### DIFF
--- a/backend/app/services/providers/garmin/handlers/lifecycle.py
+++ b/backend/app/services/providers/garmin/handlers/lifecycle.py
@@ -10,6 +10,7 @@ from typing import Any
 
 from app.database import DbSession
 from app.repositories import UserConnectionRepository
+from app.services.outgoing_webhooks.events import on_connection_revoked
 from app.utils.structured_logging import log_structured
 
 logger = logging.getLogger(__name__)
@@ -105,6 +106,13 @@ def process_deregistrations(
             continue
 
         connection_repo.mark_as_revoked(db, connection)
+        on_connection_revoked(
+            user_id=connection.user_id,
+            provider="garmin",
+            connection_id=connection.id,
+            reason="deregistration",
+            revoked_at=connection.updated_at.isoformat(),
+        )
 
         log_structured(
             logger,
--- a/backend/app/services/providers/templates/base_oauth.py
+++ b/backend/app/services/providers/templates/base_oauth.py
@@ -11,13 +11,13 @@ from uuid import UUID
 import httpx
 from fastapi import HTTPException
 from redis import Redis
-from starlette.status import HTTP_400_BAD_REQUEST, HTTP_500_INTERNAL_SERVER_ERROR
+from starlette.status import HTTP_400_BAD_REQUEST, HTTP_401_UNAUTHORIZED, HTTP_500_INTERNAL_SERVER_ERROR
 
 from app.database import DbSession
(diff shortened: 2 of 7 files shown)


backend/app/services/providers/templates/base_oauth.py
```diff
@@ -178,18 +178,39 @@ def refresh_access_token(self, db: DbSession, user_id: UUID, refresh_token: str)
                 user_id=str(user_id),
                 status_code=e.response.status_code,
             )
+            # 400/401 = refresh token is dead so revoke + notify
+            if e.response.status_code in (HTTP_400_BAD_REQUEST, HTTP_401_UNAUTHORIZED):
+                self._revoke_connection(db, user_id, reason="refresh_failed")
+                raise HTTPException(
```
bartmichalak, line 184: Won't this accidentally go to Sentry if it's called within the Celery task?

If it was called within the /sync endpoint, it shouldn't, but I'm not sure about Sentry.

backend/app/services/providers/garmin/handlers/lifecycle.py
```diff
@@ -105,6 +106,13 @@ def process_deregistrations(
             continue
 
         connection_repo.mark_as_revoked(db, connection)
+        on_connection_revoked(
+            user_id=connection.user_id,
+            provider="garmin",
+            connection_id=connection.id,
+            reason="deregistration",
+            revoked_at=connection.updated_at.isoformat(),
+        )
```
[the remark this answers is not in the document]
czajkub: `get_by_provider_user_id` already filters connections by active

backend/app/services/providers/templates/base_oauth.py
czajkub: true, it still goes to sentry because of `log_and_capture_error`

changed it so that on `HTTPException` (what is thrown here) it just logs instead of propagating to sentry - we basically only manually raise those exceptions as a handled response from our endpoints, and verified in sentry that with `HTTPException` it is only errors related to refresh tokens or user errors

backend/app/services/providers/garmin/handlers/lifecycle.py
```diff
@@ -105,6 +106,13 @@ def process_deregistrations(
             continue
 
         connection_repo.mark_as_revoked(db, connection)
+        on_connection_revoked(
```
KaliszS, line 109: don't we need to implement similar hooks for other providers?

KaliszS (changes requested):

KaliszS (approved):

PR #1256, opened 2026-06-30
del

revtukh:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sleep-time handling so summaries and records better reflect local dates when time zone offsets are present.
  * Added support for capturing and using sleep time zone offsets across connected data sources.
  * Updated user-management access handling to work with API key authentication.
  * Improved database session behavior to make post-save data available more reliably.


revtukh:
Closed: intended for rhiseai fork internal sync, not upstream contribution.

PR #1263, opened 2026-07-03
fix: emit connection.created on reconnect (existing connection branch)

Ri087:
## Problem

`base_oauth.py:_save_connection()` only emitted `connection.created` on the very first connect. On a reconnect (existing row — typically after a revoked/expired connection, which is a soft-delete so the row survives), the `existing_connection` branch updated the tokens and flipped the status back to `ACTIVE` **without emitting any Svix event**. Consumers that reacted to `connection.revoked` never learned the connection was usable again (downstream UI stuck on "pending/disconnected").

## Fix

- `_save_connection`: the `existing_connection` branch now emits `on_connection_created` after `update_connection_info`, with the updated connection's id and `updated_at` timestamp.
- `on_connection_created`: the idempotency key now includes `connected_at` (mirroring `on_connection_revoked`'s `revoked_at`). This is required, not cosmetic: the key becomes the Svix `event_id`, which deduplicates **forever** — the old key (`connection.created.{user}.{provider}`) would have silently swallowed every re-emit for the same pair.

## Tests

- `test_reconnect_emits_connection_created_and_reactivates`: revoked connection + `_save_connection` → event emitted with the existing connection id, row back to `ACTIVE` with fresh tokens.
- `test_first_connect_still_emits_connection_created`: regression guard on the historical branch.

`uv run pytest` (templates + outgoing_webhooks + garmin oauth: 45 passed) · `ruff check` ✅ · `ruff format` ✅ · `ty check` ✅

Bazard: BAZ-453 (W4, Sprint Beta). The api-side `POST /me/wearables/sync` reconciliation stays as a safety net.

## Summary by CodeRabbit

* **New Features**
  * Added live activity-stream access for supported workouts, including optional stream-key selection.
  * User deletion now supports admin API-key authentication and includes a best-effort data purge step.

* **Bug Fixes**
  * Reconnection events are now reliably emitted when a previously revoked or expired connection is restored.
  * Production services are more tightly restricted from public exposure.

* **Documentation**
  * Expanded setup and API docs for local development, activity streams, provider coverage, and protected endpoint behavior.


Ri087:
Opened by mistake against upstream — this targets our fork's integration branch. Apologies for the noise; a clean upstream proposal may follow separately.

PR #1264, opened 2026-07-04
docs: sync .env.example with config.py + CodeRabbit drift check

### LINKED ISSUE #1222: `.env.example` out of sync with `config.py`
After looking at both `.env.example` and `config.py`, I noticed that while the config file was updated, the example remained a bit stale. There are a lot of variables that can only be added manually or changed in config.py itself (for example `raw_payload_storage: str = "disabled"` or `store_fit_files: bool = False`).

Some of these are explained in the docs, but it would be better to include them in example env as well, it can even be commented initially as it is with some system config files

Also added note for coderabbit to check if newly added envs are synchronized accross both places

I guess not everything has to be in both places - some niche config settings could remain just there and explicitly added to example so as not to make it too long - like `xml_chunk_size: int = 50_000`?

aks129:
# Sync `backend/config/.env.example` with `config.py`

Fixes #1222.

## What changed

Added every user-tunable `Settings` field from `backend/app/config.py` that was missing from `backend/config/.env.example`, as commented-out entries with their default values (per the suggestion in the issue), grouped into the file's existing sections:

- **App:** `CORS_ALLOW_ALL`, `API_NAME`, `PAGING_LIMIT`
- **Sentry:** `GIT_SHA`
- **Auth:** `ALGORITHM`, `ACCESS_TOKEN_EXPIRE_MINUTES`, `MIN_PASSWORD_LENGTH`
- **Email:** `INVITATION_EXPIRE_DAYS`, `EMAIL_MAX_RETRIES`, `USER_INVITATION_CODE_EXPIRE_DAYS`
- **Raw payload storage** (new sub-section above the replay one): `RAW_PAYLOAD_STORAGE`, `RAW_PAYLOAD_MAX_SIZE_BYTES` — the main `disabled | log | s3` switch called out in the issue was missing while its `RAW_PAYLOAD_S3_*` companions were present
- **Sync:** `SLEEP_SYNC_INTERVAL_SECONDS`, `SCORE_BACKFILL_DAYS`
- **Sleep session tracking** (new sub-section): `REDIS_SLEEP_TTL_SECONDS`, `SLEEP_END_GAP_MINUTES`
- **Svix:** `SVIX_JWT_SECRET`, `SVIX_AUTH_TOKEN`
- **Providers:** `SUUNTO_DEFAULT_SCOPE`, `SUUNTO_WEBHOOK_SECRET`, `POLAR_DEFAULT_SCOPE`, `GARMIN_DEFAULT_SCOPE`, `STRAVA_WEBHOOK_SIGNATURE_TOLERANCE_SECONDS`, `STRAVA_EVENTS_PER_PAGE`

Comments were written from the actual usage sites (e.g. the Strava tolerance is the webhook signature timestamp skew check in `webhook_handler.py`), not just the field names.

Also added a `.coderabbit.yaml` path instruction for `backend/app/config.py` so CodeRabbit flags future drift between the two files, as suggested in the issue.

## Deliberately left out

- `XML_CHUNK_SIZE` — the issue itself suggests niche knobs like this can stay config-only
- `API_V1` / `API_LATEST` — route-path constants, not deployment config
- `*_REDIRECT_URI` provider vars — deprecated in favor of `API_BASE_URL`
- `TOKEN_LIFETIME` — defined in `Settings` but not referenced anywhere in `app/`; possibly dead (happy to remove it from `config.py` instead if that's preferred)
- `MASTER_KEY` / `fernet_decryptor` — read directly from `os.environ`, but no current `Settings` field uses `EncryptedField`, so documenting it seemed premature

## Verification

- AST-based cross-check: parsed the `Settings` class fields and the `.env.example` keys — every non-excluded field now appears in the example (case-insensitive), and every example key maps to a `Settings` field (`OPEN_WEARABLES_API_KEY` is the replay-script exception already documented in the file). 104 fields / 92 example keys, zero drift.
- Boot check: `cp config/.env.example config/.env` and instantiated `Settings()` (pydantic-settings) — loads cleanly; `raw_payload_storage="disabled"`, `store_fit_files=False`, derived Svix/webhook secrets all resolve.

Context: we run open-wearables in our production docker-compose alongside our FHIR guardrail stack (healthclaw.io), which is how we hit the example/config drift in practice.

## Summary by CodeRabbit

* **Documentation**
  * Expanded the sample environment configuration with more commented options and guidance for common settings.
  * Added new optional configuration entries for app behavior, authentication, storage, sync timing, sleep session tracking, webhooks, and provider integrations.
  * Improved review guidance so configuration changes should be kept in sync with the environment example file.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -26,6 +26,13 @@ reviews:
         the PR to state whether a data migration or backfill is needed for
         historical records, or explicitly note that only new data is affected.
 
+    - path: "backend/app/config.py"
+      instructions: |
+        If settings fields are added, removed, or renamed, or their defaults
+        change, verify that backend/config/.env.example is updated to match.
+        New optional settings may be added commented-out with their default
+        value. Flag any drift between the two files.
+
     - path: "backend/app/schemas/**"
       instructions: |
         If data models or enums change (e.g. new provider, new data type),
--- a/backend/config/.env.example
+++ b/backend/config/.env.example
@@ -4,12 +4,18 @@
 ENVIRONMENT="local"
 API_PORT=8000
 CORS_ORIGINS=["http://localhost:3000"]
+# CORS_ALLOW_ALL=false  # Allow any origin instead of the CORS_ORIGINS list
+# API_NAME="Open Wearables API"  # FastAPI title
+# PAGING_LIMIT=100  # Default page size for list endpoints
 
 # EMAIL SETTINGS (Resend)
 RESEND_API_KEY=your-resend-api-key
 EMAIL_FROM_ADDRESS="onboarding@resend.dev"
 EMAIL_FROM_NAME="Open Wearables"
 FRONTEND_URL=http://localhost:3000
+# INVITATION_EXPIRE_DAYS=7  # Developer invitation link lifetime
+# EMAIL_MAX_RETRIES=5  # Retries for failed email sends
+# USER_INVITATION_CODE_EXPIRE_DAYS=7  # SDK user invitation code lifetime
 
 #--- DB ---#
 DB_HOST=db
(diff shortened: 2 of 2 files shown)


KaliszS (approved):
Thank you for your time, and congratulations on your first contribution!

PR #1265, opened 2026-07-06
refactor(backend): update get method to accept string as object_id

curiousmoonnadin:
## Description
Widens the `object_id` type hint on `CrudRepository.get()` from `UUID | int` to `UUID | int | str`, since some models (e.g. `ApiKey`) use string primary keys rather than UUID/int.

### General

- [ ] My code follows the project's code style

### Backend Changes

### Frontend Changes

- [ ] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1.
2.
3.

**Expected behavior:**

## Summary by CodeRabbit

* **Enhancements**
  * Broadened record lookup support so IDs can now be provided as a string, integer, or UUID when fetching items by identifier.
* **Chores**
  * Removed an inline type-check suppression comment from the application’s item retrieval path (no behavioral change).

### DIFF
--- a/backend/app/repositories/repositories.py
+++ b/backend/app/repositories/repositories.py
@@ -32,7 +32,7 @@ class CrudRepository[
     def exists_any(self, db_session: DbSession) -> bool:
         return db_session.query(exists().where(getattr(self.model, "id").isnot(None))).scalar()
 
-    def get(self, db_session: DbSession, object_id: UUID | int) -> ModelType | None:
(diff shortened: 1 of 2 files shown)


KaliszS (approved):
You’re absolutely right. We’d missed that because there was a comment in the code instructing ty to ignore this error. I’ve also removed the comment and am approving it. Congratulations on your first contribution!

PR #1298, opened 2026-07-15
perf(integrations): skip Strava stream fetch when workout samples already exist

slavarosin:
## Problem

`_ingest_workout_streams` fetches `/api/v3/activities/{id}/streams` for every workout that passes through `load_data` or `process_push_activity`, with no check whether the samples are already in the DB. Workouts get re-upserted whenever a sync window re-covers them, the common case being a webhook arrival followed by the periodic pull over the same days. Each re-cover then re-fetches the full stream payload and re-upserts identical rows.

Cost per re-covered workout: one extra Strava API call (against the 100/15min and 1000/day budget) plus the memory of materializing the full per-second sample list. On our deployment one pull sync re-covering 24 recent activities re-fetched all 24 streams it already had.

## Fix

Streams are immutable once the activity is uploaded, so a cheap existence probe is enough: before hitting the API, check whether any `strava` sample exists for the user within `[record.start_datetime, record.end_datetime)` and return early if so.

- `DataPointSeriesRepository.has_samples_in_range` — single EXISTS query over `data_point_series` joined to `data_source`
- `TimeSeriesService.has_samples_in_range` — thin passthrough
- `StravaWorkouts._ingest_workout_streams` — early return when the probe hits

Records without `end_datetime` keep the old always-fetch behaviour. Workouts whose stream fetch failed previously have no samples in the window, so they are still retried on the next sync.

## Tests

`tests/providers/strava/test_strava_workout_samples.py`: two new wiring tests (probe hit skips fetch and save; missing `end_datetime` bypasses the probe), existing wiring tests updated to stub the probe as False. 16/16 pass.

https://claude.ai/code/session_01UwvKzR4F85gS6NBZAngfge

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate time-series samples during Strava workout reprocessing.
  * Skips unnecessary data retrieval and saving when samples already exist for the workout period.
  * Preserves normal ingestion when no end time is available or when samples are missing.

* **Tests**
  * Added coverage for existing samples, missing end times, fetch failures, save failures, and empty results.

### DIFF
--- a/backend/app/repositories/data_point_series_repository.py
+++ b/backend/app/repositories/data_point_series_repository.py
@@ -336,6 +336,27 @@ class DataPointSeriesRepository(
         limit = params.limit or 50
         return query.limit(limit + 1).all(), total_count  # ty:ignore[invalid-return-type]
 
+    def has_samples_in_range(
+        self,
+        db_session: DbSession,
+        user_id: UUID,
+        source: str,
+        start_datetime: datetime,
+        end_datetime: datetime,
+    ) -> bool:
+        """Whether any sample from `source` exists for the user in [start, end)."""
+        query = (
+            db_session.query(self.model.id)
+            .join(DataSource, self.model.data_source_id == DataSource.id)
+            .filter(
+                DataSource.user_id == user_id,
+                DataSource.source == source,
+                self.model.recorded_at >= start_datetime,
+                self.model.recorded_at < end_datetime,
+            )
+        )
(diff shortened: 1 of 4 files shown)


czajkub (approved):
Looks good to me, just one question - is it possible that there are some overlapping workout windows, which would prevent one of the workouts from having its samples ingested? or is that not gonna happen, would Strava not allow overlapping workouts?

PR #1307, opened 2026-07-16
fix(backend): make date-only end bounds inclusive of the whole day

nilzen:
## Summary by Sebastian
One running workout and one heart-rate sample, both at `2026-07-13T07:01:00Z`.

**Single day as a bare date now returns the day** (`main` returned nothing):

```
GET /events/workouts?start_date=2026-07-13&end_date=2026-07-13
```
```json
{ "data": [ { "type": "running", "start_time": "2026-07-13T07:01:00Z", "distance_meters": 8200.0 } ],
  "pagination": { "has_more": false, "total_count": 1 } }
```

Same result as the explicit `end_date=2026-07-14T00:00:00Z`, which is what a bare date now means.

**Bucketed timeseries with bare dates** (`main` returned `500`):

```
GET /timeseries?start_time=2026-07-13&end_time=2026-07-13&types=heart_rate&resolution=1min
```
```json
{ "data": [ { "timestamp": "2026-07-13T07:01:00Z", "type": "heart_rate", "value": 128.0, "unit": "bpm" } ] }
```

**`/summaries/data/timeline`** — the one call site the original diff missed:

```
GET /summaries/data/timeline?start_date=2026-07-13&end_date=2026-07-13&bucket=day
```
```json
{ "series": [ { "key": "garmin", "metric": "data_points", "buckets": [ [ "2026-07-13", 1 ] ] } ] }
```

**`/health-scores` with `start == end`** (`main` returned `500` from the `start < end` validator):

```
GET /health-scores?start_date=2026-07-13&end_date=2026-07-13
```
```json
{ "data": [], "pagination": { "has_more": false, "total_count": 0 } }
```

### Changes on top of the original diff

- A bare end date resolves to the start of the next day, not `23:59:59.999999`. Every repository compares the upper bound with `<`, so `time.max` would drop the last microsecond of the day.
- `_inclusive_end` removed from the timeseries repository: it bumped *any* midnight by a day, including an explicit `T00:00:00Z`, and after this change it would have bumped a bare date twice.
- `/summaries/data/timeline` end bound switched to the new parser.
- No `strip()` in the end parser, so both parsers treat whitespace the same way.

### Behaviour changes

- All range endpoints: a bare end date now includes that day. Clients that worked around this by passing the next day will get one extra day.
- `/timeseries`: an explicit `end_time=...T00:00:00Z` now stops at midnight instead of silently including the whole day.

## Problem

All range endpoints accept date-only strings (`YYYY-MM-DD`), but both bounds are parsed with `datetime.fromisoformat`, which normalizes a date-only string to **midnight**. For the end bound this silently excludes the entire end day:

- `GET /api/v1/users/{id}/events?start_date=2026-07-13&end_date=2026-07-13` returns nothing even when a workout started at 07:01 UTC that day — the range collapses to a zero-length window at midnight.
- Any date-only range excludes events on its last day.

This is especially visible through the MCP server, whose tool docs instruct clients to pass `YYYY-MM-DD`.

## Fix

- New `parse_query_end_datetime()` in `app/utils/dates.py`: date-only strings normalize to end-of-day (`23:59:59.999999`); full ISO datetimes and Unix timestamps pass through unchanged.
- Used for the end bound in `events`, `summaries`, `timeseries`, and `health_scores` routes (9 call sites). Start bounds keep midnight semantics.
- Query param description updated; unit tests added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Date-only end-date filters now include the entire selected day across event, health score, summary, and timeseries queries.
  * Date and time values are handled consistently, while invalid date formats and calendar dates continue to return errors.

* **Tests**
  * Added coverage for full-day date ranges, timestamps, datetime values, and invalid inputs.

* **Documentation**
  * Clarified that date-only ranges include both boundary days and that explicit times are honored.

### DIFF
--- a/backend/app/api/routes/v1/events.py
+++ b/backend/app/api/routes/v1/events.py
@@ -13,7 +13,7 @@ from app.schemas.responses.activity import (
 from app.schemas.utils import PaginatedResponse
 from app.services import ApiKeyDep
 from app.services.event_record_service import event_record_service
-from app.utils.dates import DateTimeQueryParam, parse_query_datetime
+from app.utils.dates import DateTimeQueryParam, parse_query_datetime, parse_query_end_datetime
 
 router = APIRouter()
 
@@ -32,7 +32,7 @@ def list_workouts(
     """Returns workout sessions."""
     params = EventRecordQueryParams(
         start_datetime=parse_query_datetime(start_date),
-        end_datetime=parse_query_datetime(end_date),
+        end_datetime=parse_query_end_datetime(end_date),
         cursor=cursor,
         limit=limit,
         record_type=record_type,
@@ -60,7 +60,7 @@ def list_sleep_sessions(
     """Returns sleep sessions (including naps)."""
     params = EventRecordQueryParams(
         start_datetime=parse_query_datetime(start_date),
-        end_datetime=parse_query_datetime(end_date),
+        end_datetime=parse_query_end_datetime(end_date),
         cursor=cursor,
         limit=limit,
     )
@@ -80,7 +80,7 @@ def list_menstrual_cycles(
     """Returns menstrual cycle records."""
     params = EventRecordQueryParams(
         start_datetime=parse_query_datetime(start_date),
-        end_datetime=parse_query_datetime(end_date),
+        end_datetime=parse_query_end_datetime(end_date),
         cursor=cursor,
         limit=limit,
     )
--- a/backend/app/api/routes/v1/health_scores.py
+++ b/backend/app/api/routes/v1/health_scores.py
@@ -10,7 +10,7 @@ from app.schemas.utils import PaginatedResponse, Pagination
 from app.schemas.utils.metadata import TimeseriesMetadata
 from app.services import ApiKeyDep
 from app.services.health_score_service import health_score_service
-from app.utils.dates import DateTimeQueryParam, parse_query_datetime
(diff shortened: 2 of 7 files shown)


nilzen:
Any updates on this? When using the MCP in an LLM and asking for the last three days of data it feels very unnatural to ask it for 2026-09-10 to 2026-09-14 (when I really want 2026-09-10 to 2026-09-13)

KaliszS:
Sorry to keep you waiting so long. The 1.0 release is coming up, and we’re trying to use this opportunity to streamline our bug-finding and data model verification processes. I’ve had a quick look at the PR and I can see a few minor issues. I’ll fix them myself tomorrow and merge the changes.

KaliszS (approved):
My notes after review

### Changes on top of the original diff

- A bare end date resolves to the start of the next day, not `23:59:59.999999`. Every repository compares the upper bound with `<`, so `time.max` would drop the last microsecond of the day.
- `_inclusive_end` removed from the timeseries repository: it bumped *any* midnight by a day, including an explicit `T00:00:00Z`, and after this change it would have bumped a bare date twice.
- `/summaries/data/timeline` end bound switched to the new parser.
- No `strip()` in the end parser, so both parsers treat whitespace the same way.

### Behaviour changes

- All range endpoints: a bare end date now includes that day. Clients that worked around this by passing the next day will get one extra day.
- `/timeseries`: an explicit `end_time=...T00:00:00Z` now stops at midnight instead of silently including the whole day.

PR #1314, opened 2026-07-16
perf(backend): remove event_record_detail table

### LINKED ISSUE #229: Remove EventRecordDetail table while preserving polymorphic relationship structure
**Tl;dr:** Investigate removing EventRecordDetail table redundancy while maintaining polymorphism. 

The EventRecordDetail table currently acts as a 1:1 duplicate of two columns from another table, which creates data redundancy in the database. While the polymorphic relationship structure in SQLAlchemy serves valid purposes (automatic relationship population and unified repository CRUD methods), the actual table creation appears unnecessary.

**Goal:**
SQLAlchemy models support maintaining the polymorphic relationship structure in Python code without generating the actual EventRecordDetail table in the database. SQLAlchemy documentation suggests this is possible and might be relatively easy.

**Hint:**
It is worth to test `detail` field from `EventRecord` model within Python code, whether it works after the changes.

itsnotashtonkutcher:
resolves: #229

There are 3 polymorphism options:
1. Single Table Inheritance - all subclasses share one table with all related columns. Unused ones are NULL.
Problem here: sleep_details, workout_details, menstrual_cycle_details have wildly different columns — one table would be mostly NULLs.
2. (current solution) Joined Table Inheritance - parent table + one child table per subclass, joined via FK. 
3. Concrete Table Inheritance - each subclass has its own fully independent table. No shared parent table. Can be implemented with AbstractConcreteBase (proposed in: https://github.com/the-momentum/open-wearables/pull/1156 ) or __abstract__ = True (this solution). AbstractConcreteBase  is not the best option as he UNION mapper breaks PostgreSQL GROUP BY in the summaries aggregation queries. What it does behind the hood is huge UNION, which makes it impossible to perform more complex queries (also in future). The only notable trade-off is that queries without a known category must fan out across all three tables, though each lookup is a single-column PK scan so the overhead is minimal in practice.

## Summary by CodeRabbit

* **Improvements**
  * Event records now provide dedicated details for sleep, workouts, and menstrual cycles.
  * Responses, sleep-session merging, and sleep scoring now consistently use the correct category-specific data.
* **Bug Fixes**
  * Deleting event records now reliably cascades to remove the associated category detail records.
  * Detail creation, lookups, and updates more accurately distinguish between categories.
* **Tests**
  * Updated API and repository/service tests to reflect the new detail structure.

### DIFF
--- a/backend/app/mappings.py
+++ b/backend/app/mappings.py
@@ -44,10 +44,6 @@ FKEventRecord = Annotated[
     UUID,
     mapped_column(ForeignKey("event_record.id", ondelete="CASCADE"), primary_key=True),
 ]
-FKEventRecordDetail = Annotated[
-    UUID,
-    mapped_column(ForeignKey("event_record_detail.record_id", ondelete="CASCADE"), primary_key=True),
-]
 FKDataSource = Annotated[
     UUID,
     mapped_column(ForeignKey("data_source.id", ondelete="CASCADE")),
--- a/backend/app/models/__init__.py
+++ b/backend/app/models/__init__.py
@@ -2,25 +2,32 @@ from .api_key import ApiKey
 from .application import Application
 from .archival_setting import ArchivalSetting
 from .data_point_series import DataPointSeries
-from .health_score import HealthScore
 from .data_point_series_archive import DataPointSeriesArchive
 from .data_source import DataSource
 from .developer import Developer
 from .device_type_priority import DeviceTypePriority
 from .event_record import EventRecord
-from .event_record_detail import EventRecordDetail
+from .event_record_detail import DetailType, EventRecordDetail
+from .health_score import HealthScore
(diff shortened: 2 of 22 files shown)


KaliszS (changes requested):
`detail_type` field in detailed event record tables (like sleep details) seems to be redundant (we take that value from business layer, so this field be be just calculated on python side). If we remove that, we can avoid costly db operations within migration. We have to take into consideration some OW clients might have a huge user base and such a migration might be an over kill for them.

There is also issue with data integratity. Looks like `created_at` is missing in modified tables, so opening data tabs in frontend crashes webpagae. The reason is behind inheritance. `BaseDbModel` includes that field by default, but `SleepDetails` etc inherit from `EventRecordDetail`.

KaliszS (approved):

PR #1315, opened 2026-07-17
fix(frontend): emit root stylesheet from client build

### LINKED ISSUE #1292: Production frontend image omits referenced hashed stylesheet
## Problem

The production frontend build on current `main` serves root HTML that references a stylesheet which is absent from the deployed output.

- Commit: `3a1ea230819b965f9dee8611d2826f39b3a49885`
- Railway deployment ID: `0287b775-d64f-4513-9c03-61bb9c035be3`
- Root HTML status: `200`
- Referenced asset: `/assets/styles-cunT8RDB.css`
- Asset response: `404` (`text/html; charset=utf-8`)

## Reproduction

`GET /` returns HTML containing:

`href="/assets/styles-cunT8RDB.css"`

Requesting that exact path returns HTTP 404, while the other referenced JS/static assets return HTTP 200.

## Expected

Every asset referenced by the generated root HTML is copied into the production image and served with HTTP 200. The missing stylesheet blocks correct rendering of the admin dashboard.

The deployment uses the repository's current `frontend/Dockerfile` and upstream `main` directly; no custom image, source fork, or Railway-side patch is involved.

aritsyg:
## Summary

- import the root stylesheet as CSS instead of a `?url` asset
- let Vite/TanStack Start inject the client-build stylesheet link into SSR HTML
- avoid Linux production builds referencing an SSR-only stylesheet that is absent from `.output/public`

Fixes #1292.

## Root cause

The existing `styles.css?url` import is resolved independently in the client and SSR environments:

- client output: `.output/public/assets/styles-DSq6jfP6.css`
- SSR intermediate output: `node_modules/.nitro/vite/services/ssr/assets/styles-DYfE_Pvq.css`

On Linux (`node:22-alpine`), the generated server route bundle references the SSR hash, but the production image copies only `.output`; the root HTML therefore requests `/assets/styles-DYfE_Pvq.css`, which returns 404. A native macOS build happened to reference the client hash, which explains why the issue was not reproducible there.

Using a side-effect CSS import makes the generated Linux server HTML link the emitted client stylesheet instead.

## Verification

- `pnpm run lint`
- `pnpm run format:check`
- clean `node:22-alpine` Docker build using the repository `frontend/Dockerfile`
- ran the resulting image and requested every asset referenced by `GET /`
- root stylesheet: `/assets/index-DSq6jfP6.css` → HTTP 200
- all 12 root-referenced assets → HTTP 200

## Summary by CodeRabbit

* **Style**
  * Updated global styling integration to ensure application styles are applied directly.
  * Removed the redundant stylesheet link from the page head.

### DIFF
--- a/frontend/src/routes/__root.tsx
+++ b/frontend/src/routes/__root.tsx
@@ -12,7 +12,7 @@ import { queryClient } from '@/lib/query/client';
 import { runtimeConfigScript } from '@/lib/api/runtime-config';
 import { Toaster } from '@/components/ui/sonner';
 
-import appCss from '../styles.css?url';
+import '../styles.css';
 
 export const Route = createRootRoute({
   head: () => ({
@@ -47,10 +47,6 @@ export const Route = createRootRoute({
       {
         rel: 'stylesheet',
         href: 'https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700;800&display=swap',
-      },
-      {
-        rel: 'stylesheet',
-        href: appCss,
       },
       // Fallback for browsers that don't support media queries
       {


KaliszS (approved):
Good catch and thank you for the contribution!

Issue #1380, opened 2026-08-03
Garmin webhook accepts any non-empty garmin-client-id header; docker-compose.prod.yml publishes Postgres + Redis on 0.0.0.0


elfrost:
Two items from a read of the webhook and deployment paths at `87f5893`. Both are unauthenticated and both have small fixes. Happy to open a PR for either if you'd rather review a diff — I held off because each involves a choice that's yours to make.

---

## 1. The Garmin webhook accepts any request that sends a non-empty `garmin-client-id` header

`app/services/providers/garmin/webhook_handler.py:95-106`:

```python
def verify_signature(self, request: Request, body: bytes) -> bool:
    client_id = request.headers.get("garmin-client-id")
    if not client_id:
        log_structured(..., "Missing garmin-client-id header", ...)
        return False
    return True
```

The header is read and tested for presence, but its value is never compared to anything — `settings.garmin_client_id` (`app/config.py:143`) is configured and unused here. Any non-empty string passes.

I ran the shipped method against a stub request to be sure I wasn't misreading it:

| `garmin-client-id` sent | `verify_signature()` |
| --- | --- |
| *(absent)* | `False` → 401 |
| `x` | **`True`** |
| `0` | **`True`** |
| `attacker-invented` | **`True`** |

Your own test suite shows the same thing from the outside — `tests/api/v1/test_garmin_webhooks.py` calls the endpoint with `"test-client-id"` and `"x"`, neither of which is configured anywhere.

**Why it matters.** `BaseWebhookHandler.handle()` treats `verify_signature` as the authentication boundary — it's the only gate before `parse_payload` → `dispatch`, and the base class docstring describes it as verifying "the authenticity of an incoming webhook request" (it also lists Garmin as using a *shared-secret* `Authorization` header, which is a different scheme from what's implemented). Nothing else guards these routes: neither `POST /api/v1/providers/garmin/webhooks` nor the compat routes in `deprecated_webhooks.py` carry an auth dependency, and there's no router-level `dependencies=`. The compat comment notes the legacy path "was in production".

So an unauthenticated request reaches `dispatch()`, which enqueues `process_webhook_push`. For any user whose Garmin-side `userId` is known, that reaches:

- `handlers/lifecycle.py:process_deregistrations` → `mark_as_revoked` on the connection, plus an outgoing `connection.revoked` webhook
- `handlers/lifecycle.py:process_user_permissions` → overwrites `connection.scope`
[...]

PR #1393, opened 2026-08-04
fix(backend): verify Strava webhook HMAC-SHA256 signature

Sirius1616:
## Description

Strava's `X-Strava-Signature` header was only checked for timestamp freshness — the `v1` HMAC-SHA256 digest was never verified. Anyone able to reach `POST /providers/strava/webhooks` could forge activity create/update/delete and athlete deauthorize events, injecting attacker-controlled data and activity records.

`verify_signature` now validates the digest against `HMAC-SHA256(strava_client_secret, "{t}.{raw_request_body}")` using the shared `_verify_hmac_sha256` base helper (same pattern as the Whoop and Oura handlers), rejects requests when `strava_client_secret` is unset, and keeps the existing timestamp-freshness check.

Related to #1380 (same class of webhook signature gap on Garmin).

### General

### Backend Changes

## Testing Instructions

**Steps to test:**
1. `cd backend`
2. `uv run pytest tests/providers/strava/ -q` (all 22 tests pass, 8 new)
3. `uv run ruff check app/services/providers/strava/webhook_handler.py tests/providers/strava/test_strava_webhook_handler.py`
4. `uv run --group code-quality ty check app/services/providers/strava/webhook_handler.py tests/providers/strava/test_strava_webhook_handler.py`

**Expected behavior:**
- A correctly signed Strava webhook (fresh timestamp + valid digest) is accepted.
- Forged signatures, tampered bodies, stale timestamps, missing headers, malformed headers, an unset client secret, or a wrong client secret are all rejected.

## Summary by CodeRabbit

* **Security**
  * Strengthened Strava webhook validation using HMAC-SHA256 signatures.
  * Webhooks are rejected when the client secret or signature is missing, invalid, stale, or does not match the request body.
* **Bug Fixes**
  * Prevented acceptance of tampered webhook requests that previously passed timestamp checks alone.
* **Tests**
  * Added coverage for valid signatures and common invalid or tampered request scenarios.

### DIFF
--- a/backend/app/services/providers/strava/webhook_handler.py
+++ b/backend/app/services/providers/strava/webhook_handler.py
@@ -73,11 +73,23 @@ class StravaWebhookHandler(BaseWebhookHandler):
     # ------------------------------------------------------------------
 
     def verify_signature(self, request: Request, body: bytes) -> bool:
-        """Validate timestamp from X-Strava-Signature; skip HMAC.
+        """Verify X-Strava-Signature using HMAC-SHA256 over the raw body.
 
-        Security relies on the hub.challenge handshake at subscription time.
-        Replays are rejected using the timestamp in X-Strava-Signature.
+        Strava signs the message ``{t}.{raw_request_body}`` with
+        HMAC-SHA256(strava_client_secret) and sends the hex digest as ``v1``.
+        The timestamp is also checked for freshness to bound replay windows.
         """
+        secret_setting = settings.strava_client_secret
+        if not secret_setting:
+            log_structured(
+                logger,
+                "error",
+                "STRAVA_CLIENT_SECRET not configured; rejecting webhook",
+                provider="strava",
+                action="webhook_signature_missing_secret",
+            )
+            return False
+
         header = request.headers.get("X-Strava-Signature", "")
         if not header:
             return False
@@ -85,6 +97,7 @@ class StravaWebhookHandler(BaseWebhookHandler):
         try:
             parts = dict(p.split("=", 1) for p in header.split(","))
             timestamp = int(parts["t"])
+            provided_signature = parts["v1"]
         except (KeyError, ValueError):
             return False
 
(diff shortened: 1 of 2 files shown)


backend/app/services/providers/strava/webhook_handler.py
```diff
@@ -73,18 +73,31 @@ def __init__(self, workouts: StravaWorkouts) -> None:
...
-        Replays are rejected using the timestamp in X-Strava-Signature.
+        Strava signs the message ``{t}.{raw_request_body}`` with
+        HMAC-SHA256(strava_client_secret) and sends the hex digest as ``v1``.
+        The timestamp is also checked for freshness to bound replay windows.
         """
+        secret_setting = settings.strava_client_secret
+        if not secret_setting:
+            log_structured(
+                logger,
+                "error",
```
czajkub, line 86: I would rather have this as a `warning`, to prevent this leaking into things like sentry

backend/app/services/providers/strava/webhook_handler.py
```diff
@@ -98,7 +111,13 @@ def verify_signature(self, request: Request, body: bytes) -> bool:
             )
             return False
 
-        return True
+        secret = secret_setting.get_secret_value()
```
czajkub, line 114: are you sure that the `client_secret` for strava is the correct value here for verification?

czajkub (changes requested):
Hello, did you verify this change yourself? I'm not quite certain that the `client_secret` for Strava is the correct way to verify the incoming signature - I recall this being a large issue with the webhooks, with Strava's docs not explicitly telling what value to use for verification, and forum posts asking what to use.

If this is verified, then good - but could you please provide proof (like screenshots of logs with verified signature coming from Strava's webhooks) that this works?

backend/app/services/providers/strava/webhook_handler.py
KaliszS: It's obviously incorrect. I have seen similar PR for the Suunto and it didn't work, Claude just tends to add this into every provider strategy.

KaliszS:
I’m closing this due to a lack of response to the review. What’s more, we had a similar PR for Suunto, where Claude was also seeing things and the change simply didn’t work; it’s most likely the same here.

PR #1376, opened 2026-07-30
fix(frontend): load theme colors under Tailwind v4

MichalZak17:
## Description

The frontend runs Tailwind v4 (`tailwindcss@4.3.3` + `@tailwindcss/vite`), but the semantic color palette is declared in `tailwind.config.ts`. Tailwind v4 does not auto-load that file, and `src/styles.css` had no `@config` directive, so every utility built on those tokens generated **no CSS at all**: `bg-popover`, `bg-card`, `bg-muted`, `bg-accent`, `bg-sidebar-*`, `text-muted-foreground`, `border-input`, `ring-ring`. Probing the running app's stylesheet for those selectors returned zero rules; only Tailwind's built-in palette (`bg-white`, `bg-red-500`) worked.

The visible symptom reported in #1359 is the **Sync History** dropdown rendering with no background, so page content behind it shows through and overlaps the menu options. The same defect affects every other dropdown (the connection card ⋮ menu, the user page ⋯ menu) and every tooltip (`bg-card/90` in `tooltip.tsx`). Dialogs, alert dialogs and sheets were unaffected only because they had already been worked around with `bg-[hsl(var(--x))]` arbitrary values — there are 56 such workarounds across 25 files.

**Fix:** reference the existing config from the stylesheet so the palette is generated again. No color values change — they still resolve from the same CSS variables already defined in `styles.css`.

**Follow-on change:** with the palette live again, the status badge variants started rendering their intended solid neon backgrounds, which were unreadable against the foreground colors they had been paired with while the classes were dead (e.g. `bg-success` + `text-green-300`). `success`/`warning`/`destructive` are restyled as tinted chips — `bg-*/10`, `border-*/30`, `text-*-glow` — consistent with each other and with the tinted treatment used elsewhere in the app. Three hard-coded icon colors in `connection-card.tsx` were removed so icons inherit the badge color.

Fixes #1359

### General

- [x] My code follows the project's code style
- [x] I have performed a self-review of my code
- [ ] I have added tests that prove my fix/feature works (if applicable)
- [x] New and existing tests pass locally
- [x] I have updated relevant documentation in `docs/` (or no docs update needed)

### Frontend Changes

- [x] `pnpm run lint` passes
- [x] `pnpm run format:check` passes
- [x] `pnpm run build` succeeds

## Testing Instructions

**Steps to test:**
1. Open a user's page that has a connected provider with `max_historical_days` unset (e.g. Polar or WHOOP, status Active).
2. Click the **Sync History** button on the connection card.
[...]

PR #1397, opened 2026-08-05
refactor(frontend): replace hsl(var(--x)) workarounds with theme utilities

### LINKED ISSUE #1396: Replace bg-[hsl(var(--x))] workarounds with theme utilities
## Problem Statement

Until #1376, Tailwind v4 was not loading `tailwind.config.ts`, so every utility built on the semantic palette (`bg-card`, `text-muted-foreground`, `border-input`, …) generated no CSS. Components that needed those colors worked around it by writing the CSS variable inline as an arbitrary value:

```tsx
className="bg-[hsl(var(--card))] text-[hsl(var(--foreground-muted))]"
```

#1376 added the `@config` directive, so the palette is generated again and the plain utilities work. The workarounds still function, but they are now redundant indirection. There are **172 of them across 36 files**:

| Prefix | Count |
|---|---|
| `text-[hsl(var(--…))]` | 90 |
| `bg-[hsl(var(--…))]` | 56 |
| `border-[hsl(var(--…))]` | 24 |
| `border-b-[hsl(var(--…))]` | 2 |

This matters beyond tidiness:

- **The pattern spreads.** It is now the majority spelling in the codebase, so new components copy it, and each copy re-teaches the workaround as if it were the house style.
- **Two spellings for one color.** `bg-card` and `bg-[hsl(var(--card))]` produce identical CSS, so reviewers have no way to tell which is intended, and a search for `bg-card` misses most of its real usages.
- **Arbitrary values bypass the theme.** Renaming or restructuring a token in `tailwind.config.ts` will not surface these call sites — they reference the raw CSS variable directly, so they break silently rather than at build time.

## Proposed Solution

Mechanically convert the arbitrary values back to the theme utilities:

```diff
-className="bg-[hsl(var(--card))]"
+className="bg-card"

-className="text-[hsl(var(--foreground-subtle))]"
+className="text-foreground-subtle"

-className="bg-[hsl(var(--warning-muted)/0.1)] border-[hsl(var(--warning-muted)/0.2)]"
+className="bg-warning-muted/10 border-warning-muted/20"
```

All 12 distinct tokens currently used this way already have a matching entry in `tailwind.config.ts`, so every occurrence has a direct equivalent — no config changes are needed:

`--success-muted` (31), `--destructive-muted` (30), `--warning-muted` (25), `--secondary-muted`, `--primary-muted`, `--primary`, `--popover`, `--foreground-subtle`, `--foreground-muted`, `--accent-muted`, `--destructive`, `--card`

### Verification

The change should be provably a no-op visually:

1. `pnpm run build` before and after, then diff the emitted stylesheet in `.output/public/assets/*.css`. Rule-level differences should be nil or explainable.
2. `pnpm run lint`, `pnpm run format:check`, `pnpm run build` all pass.
3. Visual pass over Dashboard, Users, user detail tabs, Syncs, Settings, Webhooks, and the widget connect flow.

### Caveat worth checking

The ~45 occurrences carrying an alpha suffix (`hsl(var(--x)/0.1)`) are the only ones that are not a pure textual swap. Tailwind v4 implements the `/10` opacity modifier with `color-mix(in oklab, …)` rather than plain HSL alpha. The result should be visually identical, but these are worth an explicit before/after comparison rather than a blind find-and-replace.

## Use Case

Anyone touching the frontend. New components get one obvious way to reference a theme color, reviewers stop seeing two spellings for the same thing, and the palette becomes refactorable — a token rename would fail loudly instead of leaving dead call sites behind.

## Alternatives Considered

- **Leave them.** They work, and this produces a large diff touching 36 files. But the pattern is actively spreading, and the cost of removal grows with every new component.
- **Convert opportunistically, file by file.** Lower risk per PR, but leaves the codebase in a mixed state indefinitely — which is the ambiguity the change is meant to remove.
- **Add a lint rule banning `[hsl(var(--…))]` instead.** Useful as a follow-up to prevent regressions, but it does not clean up what is already there, and it would fail CI on 172 existing occurrences on day one.

## Additional Context

Flagged in the Additional Notes of #1376, which fixed the underlying cause (#1359).

The diff is large but shallow — every hunk is a class-string substitution with no logic change — so it should review quickly despite the file count. Splitting by directory (`components/ui`, `components/user`, `components/pages`, `routes`) would keep individual PRs reviewable if a single PR is judged too big.

Good candidate for `good first issue`: the work is mechanical, the correct output is unambiguous, and it is verifiable by diffing the built CSS.

MichalZak17:
## Description

While Tailwind v4 was not loading `tailwind.config.ts` (fixed in #1376), the semantic palette generated no CSS, so components wrote the CSS variables inline as arbitrary values. Now that the palette is generated again, those occurrences are redundant indirection: two spellings for the same color, and call sites that reference the raw variable, so a token rename in the config would leave them silently broken.

This converts all **172 occurrences across 36 files** back to the theme utilities:

```diff
-className="bg-[hsl(var(--card))]"
+className="bg-card"

-className="text-[hsl(var(--foreground-muted))]"
+className="text-foreground-muted"

-className="bg-[hsl(var(--warning-muted)/0.1)] border-[hsl(var(--warning-muted)/0.2)]"
+className="bg-warning-muted/10 border-warning-muted/20"
```

All 12 tokens used this way already have entries in `tailwind.config.ts`, so no config change was needed. No occurrences remain in `src/`.

Closes #1396

### General

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

This should be a visual no-op. The most efficient review is to diff the built stylesheet rather than read all 36 files.

**Steps to test:**
1. `cd frontend && pnpm run build` on `main`, keep a copy of `.output/public/assets/*.css`.
2. Check out this branch, build again, and diff the two stylesheets.
3. Spot-check the pages with the densest changes: Dashboard (stat cards), Settings → Data Lifecycle (early-access banner, storage overview), Webhooks (attempts table), and a user detail page (status badges, sleep/body sections).

**Expected behavior:**

The 61 arbitrary-value rules (`.bg-\[hsl\(var\(--card\)\)\]` and friends) are gone, replaced one-for-one by the corresponding utility rules (`.bg-card`). No other rules change, and the UI renders identically.

## Additional Notes

**The opacity-modifier cases are the only ones that are not a pure textual swap.** Tailwind v4 compiles `bg-warning-muted/10` to `color-mix(in oklab, hsl(var(--warning-muted)) 10%, transparent)` rather than plain HSL alpha. I measured these rather than assuming: composited over white and read back as pixels, five of six sampled colors are byte-identical, and `bg-warning-muted/10` differs by 1/255 in two channels (`250,246,232` vs `250,245,231`) from oklab rounding. Not perceptible, but worth stating precisely rather than claiming byte-for-byte identity.

A follow-up worth considering: a lint rule banning `[hsl(var(--…))]` in `className` would stop the pattern from being reintroduced. It only makes sense after this lands, since it would otherwise fail CI on all 172 existing occurrences.

## Summary by CodeRabbit

* **Style**
  * Standardized interface colors across dashboards, settings, user profiles, webhooks, dialogs, alerts, and forms.
  * Improved consistency for success, warning, destructive, informational, and status indicators.
  * Preserved existing layouts, messaging, interactions, validation, and component behavior.


bartmichalak (approved):

bartmichalak (changes requested):
Please resolve conflicts

KaliszS (approved):

PR #1406, opened 2026-08-06
fix: remove `handle_duplicates` decorator from `event_record_detail_service`'s `create`

### LINKED ISSUE #1375: `EventRecordDetailRepository.create()` fails on duplicate inserts after abstract detail models
## Related

- [#730](https://github.com/the-momentum/open-wearables/issues/730) / #1047 added savepoint + idempotent handling for `create_and_flush()` only.
- `create()` still relied on `@handle_duplicates`, which breaks once `EventRecordDetail` is an abstract base (concrete rows live in `workout_details` / `sleep_details` / etc.).

## Problem

On vendor re-sync (e.g. Whoop workouts), inserting an event detail for a `record_id` that already exists raises a unique constraint violation. Recovery via `@handle_duplicates` then fails while trying to look up the existing row, so the sync errors instead of returning the existing detail.

`create_and_flush()` already handles this pattern; `create()` does not.

## Error

Primary failure during duplicate recovery:

```text
NoInspectionAvailable: No inspection system is available for object of type <class 'app.utils.mappings_meta.AutoRelMeta'>
```

Raised from:

```text
File "app/utils/duplicates.py", in wrapper
    mapper = inspect(model)
```

with `model` set to abstract `EventRecordDetail`.

Underlying insert conflict:

```text
IntegrityError: (psycopg.errors.UniqueViolation) duplicate key value violates unique constraint "workout_details_pkey"
DETAIL:  Key (record_id)=(...) already exists.
[SQL: INSERT INTO workout_details (...) VALUES (...) RETURNING workout_details.created_at]
```

Call path (abbreviated):

```text
sync_vendor_data_task.sync_vendor_data
  → whoop.workouts.load_data
    → event_record_service.create_detail
      → EventRecordDetailRepository.create
        → handle_duplicates → inspect(EventRecordDetail)
```

## Root cause

1. Production constructs `EventRecordDetailRepository(EventRecordDetail)` where `EventRecordDetail` is `__abstract__ = True`.
2. `create()` inserts a concrete subclass from `DETAIL_MODELS[detail_type]` (e.g. `WorkoutDetails`).
3. On `UniqueViolation`, `@handle_duplicates` calls `sqlalchemy.inspect(self.model)` and queries that model. Abstract classes are not inspectable / queryable, so recovery raises `NoInspectionAvailable` instead of returning the existing row.

## Expected Behavior

Same as `EventRecordRepository.create` and `EventRecordDetailRepository.create_and_flush`: on duplicate `record_id`, return the existing detail without failing the sync.

## Actual Behavior

Duplicate insert triggers unique violation; duplicate handler crashes with `NoInspectionAvailable`; workouts sync fails for the provider.

## Environment

- Python 3.13
- PostgreSQL + SQLAlchemy 2.0
- Triggered by Celery vendor sync re-processing overlapping workout windows

## Proposed Fix

Remove `@handle_duplicates` from `EventRecordDetailRepository.create`. Catch `IntegrityError`, roll back, and return `get_by_record_id(..., detail_type)` (mirror `EventRecordRepository.create`). Keep savepoints on `create_and_flush` only.

### Before

```python
@handle_exceptions
@handle_duplicates
def create(
    self,
    db_session: DbSession,
    creator: EventRecordDetailCreate,
    detail_type: DetailType = "workout",
) -> EventRecordDetail:
    """Create a detail record using the appropriate polymorphic model."""
    detail = self._build_detail(creator, detail_type)
    db_session.add(detail)
    db_session.commit()
    db_session.refresh(detail)
    return detail
```

### After

```python
@handle_exceptions
def create(
    self,
    db_session: DbSession,
    creator: EventRecordDetailCreate,
    detail_type: DetailType = "workout",
) -> EventRecordDetail:
    """Create a detail record using the appropriate polymorphic model.

    Idempotent on record_id: if the detail already exists, returns the existing row.
    """
    detail = self._build_detail(creator, detail_type)
    try:
        db_session.add(detail)
        db_session.commit()
        db_session.refresh(detail)
        return detail
    except IntegrityError:
        db_session.rollback()
        if existing := self.get_by_record_id(db_session, creator.record_id, detail_type):
            return existing
        raise
```

czajkub:
## Description

After we changed our approach to the `event_record_detail` abstract table (with pr #1314), the decorator starts throwing uncaught errors because of the lack of model associated - we also noticed this on our sentry instance:
```
NoInspectionAvailable
No inspection system is available for object of type <class 'app.utils.mappings_meta.AutoRelMeta'>
```

The change is that the scope is smaller than the decorator used before (now the only unique constraint on the table is queried - which doesn't necessarily remove any overhead, but simplifies the code a bit)

The functionality stays the same as before the changes, but errors should not be raised further up anymore

Also added some tests to prevent this from firing in the future (hopefully)

Resolves #1375

### General

### Backend Changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of duplicate event detail submissions.
  * Existing matching details are now reused without overwriting their original values.
  * Genuine database errors continue to be reported appropriately.
  * Improved consistency when creating event details through different workflows.
  * Event detail types are now validated more reliably, reducing invalid submissions and unexpected errors.

### DIFF
--- a/backend/app/repositories/event_record_detail_repository.py
+++ b/backend/app/repositories/event_record_detail_repository.py
@@ -1,6 +1,7 @@
 from typing import Any, cast
 from uuid import UUID
 
+from psycopg.errors import UniqueViolation
 from sqlalchemy import Table, update
 from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.exc import IntegrityError
@@ -55,9 +56,12 @@ class EventRecordDetailRepository(
             db_session.commit()
             db_session.refresh(detail)
             return detail
-        except IntegrityError:
+        except IntegrityError as exc:
             db_session.rollback()
-            if existing := self.get_by_record_id(db_session, creator.record_id, detail_type):
+            # record_id is the only unique index here; FK and other violations are real errors
+            if isinstance(exc.orig, UniqueViolation) and (
+                existing := self.get_by_record_id(db_session, creator.record_id, detail_type)
+            ):
                 return existing
             raise
 
@@ -79,9 +83,11 @@ class EventRecordDetailRepository(
             db_session.flush()
             nested.commit()
             return detail
-        except IntegrityError:
+        except IntegrityError as exc:
             nested.rollback()
-            if existing := self.get_by_record_id(db_session, creator.record_id, detail_type):
+            if isinstance(exc.orig, UniqueViolation) and (
+                existing := self.get_by_record_id(db_session, creator.record_id, detail_type)
+            ):
                 return existing
             raise
 
--- a/backend/app/repositories/event_record_detail_repository.py
+++ b/backend/app/repositories/event_record_detail_repository.py
@@ -17,7 +17,6 @@ from app.schemas.model_crud.activities import (
     EventRecordDetailCreate,
     EventRecordDetailUpdate,
 )
-from app.utils.duplicates import handle_duplicates
 from app.utils.exceptions import handle_exceptions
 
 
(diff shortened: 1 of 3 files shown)


bartmichalak (commented):
There’s still one pending comment from 🐰, please address or decline it

czajkub:
sure @bartmichalak - addressed the comment

bartmichalak:
@czajkub one more question. Was the outcome of this issue as follows?

When the first duplicate workout appeared during a any re-sync (for example: Whoop), it crashed the entire sync, so every record after it (not before) - including those not-duplicated records - was dropped and never saved.

?

czajkub:
@bartmichalak - not sure if it would crash the entire sync, my guess is that it depends on the implementation - whether there was a try/except implemented somewhere to scope down the error and continue further.

In principle this should have affected mostly the workout/sleep path. Although if there was a duplicate detected, it was likely that the rest may have been duplicates too - so this may have been unnoticed for some time (excluding the sentry error)

bartmichalak (approved):

PR #1423, opened 2026-08-11
feat: Emit `connection.created` and `connection.revoked` events for SDK

czajkub:
## Description

our SDK doesnt emit the `connection.x` events currently, so it is implemented this way:

for `created`:
- arrives with first chunk of data sent from SDK (when `ensure_sdk_connection` is called)
- also emitted when a previously revoked connection is marked as active again

for `revoked`:
- Not implemented, but SDK should implement hitting the `DELETE /users/{id}/connections/apple` endpoint when disconnecting - not a backend change though.

### General

### Backend Changes

## Summary by CodeRabbit

- **New Features**
  - Upload results now show separate counts for newly inserted and updated records, including sync metadata.
  - SDK connections are identified as newly created, reactivated, or already active.
  - SDK uploads and resumed connections provide clearer connection events and status details.

- **Bug Fixes**
  - Improved connection handling to prevent duplicate events and support repeated or resumed uploads.

- **Documentation**
  - Updated sync status and webhook documentation to clarify record counts, SDK connection events, and revocation reasons.


backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
```diff
@@ -0,0 +1,81 @@
...
+
+def _is_sdk_fed(connection: UserConnection) -> bool:
+    """True when the SDK upload endpoint is this connection's only way in.
+
+    Only then is revoking safe: the next upload reactivates it. Decided per connection,
+    not per provider - Google has both an SDK and an OAuth integration under one slug,
+    and no tokens is what marks the SDK-provisioned rows.
+    """
+    try:
+        caps = ProviderFactory().get_provider(connection.provider).capabilities
```
czajkub: i dont think this would be a big improvement in performance, but we still create the strategy each time with no global instance somewhere in the codebase - maybe we could add some form of caching the strategies and global factory for cohesion?

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
```diff
@@ -0,0 +1,81 @@
...
+            action="stale_connection_sweep_start",
+            threshold_days=threshold_days,
+            stale_total=len(stale_connections),
+            stale_sdk=len(candidates),
+            skipped_non_sdk=len(stale_connections) - len(candidates),
+        )
+
+        for connection in candidates:
+            try:
+                if user_connection_service.revoke_as_stale(db, connection):
```
czajkub: here it is marked as revoked, since this is the most fitting connection status - I thought about adding a `STALE` status, but we would need to handle this status accross a lot of branches in the codebase, and would be more fitting if also implemented for the rest of the providers as well, not SDK only

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
```diff
@@ -0,0 +1,81 @@
...
+    """True when the SDK upload endpoint is this connection's only way in.
+
+    Only then is revoking safe: the next upload reactivates it. Decided per connection,
+    not per provider - Google has both an SDK and an OAuth integration under one slug,
+    and no tokens is what marks the SDK-provisioned rows.
+    """
+    try:
+        caps = ProviderFactory().get_provider(connection.provider).capabilities
+    except Exception:
+        return False
```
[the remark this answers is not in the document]
czajkub: this is never gonna happen, this is just a catch for a very edge case of removing a provider from codebase (which will not happen), or drastic changes

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
```diff
@@ -0,0 +1,81 @@
...
+        logger,
+        "info",
+        f"Revoked {len(revoked)} stale connection(s)",
+        action="stale_connection_sweep_complete",
+        threshold_days=threshold_days,
+        revoked_count=len(revoked),
+        revoked=revoked,
+    )
+
+    return {"threshold_days": threshold_days, "revoked_count": len(revoked), "revoked": revoked}
```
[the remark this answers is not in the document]
czajkub: this is for debug purposes and will never be exposed further, also logging to sentry doesnt contain sensitive data. this will stay locally

backend/app/repositories/user_connection_repository.py
```diff
@@ -266,6 +267,48 @@ def get_expiring_tokens(self, db_session: DbSession, minutes_threshold: int = 5)
             .all()
         )
 
+    def revoke_stale_sdk_connections(
```
bartmichalak: I agree, as you wrote in the description, that it would be better to mark this connection as stale rather than revoke.

revoke suggests that the mobile SDK will no longer be able to reconnect using the same token, which is not true (access / refresh token might be still active and kept on the mobile side)

I think we should leave this revoke implementation out for now (less code == better) and create an issue for handling the stale status.

WDYT?

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
```diff
@@ -0,0 +1,59 @@
...
+            caps = ProviderFactory().get_provider(provider.value).capabilities
+        except ValueError:
+            continue
+        if caps.client_sdk:
+            providers.append(provider.value)
+    return providers
+
+
+@shared_task
+def revoke_stale_connections() -> dict:
```
bartmichalak: I'm opting for removing this task completely for now, but general note to keep in mind for the future: the name of the task suggests that it applies to all providers, while the logic already speaks of cloud-based

bartmichalak (changes requested):

bartmichalak:
also please create relevant issues in our Flutter and RN SDKs 🙏

backend/app/repositories/user_connection_repository.py
czajkub: fair point, we can leave it unimplemented for now and handle the stale connection issue separately

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
czajkub: true, also if we switch the mechanism to work with a new `stale` connection status then we can implement it the same way for all providers, not only sdk-based

navFooh:
This computes the fact #1426 needs, then drops it.

`ensure_sdk_connection` now reports whether the SDK created, reactivated, or found the row, but that only drives a webhook. No column is added, so nothing records that an SDK ever fed a connection.

The `EXISTING` branch is where that shows. `(user_id, provider)` is unique, and Health Connect uploads under `setProvider("google")`, the same slug as the cloud OAuth provider. So an account that connects Google Health over OAuth and then installs the Android SDK has a single row that hits `EXISTING` on every upload and comes back untouched, OAuth tokens intact, no trace of the phone.

That undercuts the inference #1426 plans on:

```python
sdk_fed = connection.access_token is None and connection.refresh_token is None
```

It reads False there, so the guard would reject the SDK's own disconnect against a connection the SDK actually feeds. Both paths share one `google` identity by design, so nothing token shaped can separate them.

Since you are already in this function, would you consider persisting it? A nullable flag on `UserConnection`, stamped on first SDK touch and left alone after, keeps `EXISTING` free of writes. Additive rather than an origin enum, since a `google` row can be both.

If it lands, could it also go on `UserConnectionRead`? Tokens are stripped there, so an API consumer cannot derive it at all, and `DELETE /users/{user_id}/connections/{provider}` currently means either "ends it" or "ends it until the next upload" with nothing on the wire to say which.

bartmichalak (changes requested):
I think we're generally good to go, but first the PR title and description need to be updated, once we stop marking stale connections as revoked 😄

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
bartmichalak: I think @KaliszS  will have the best perspective on this? But anyway, it's something to introduce separately and doesn't block this change

bartmichalak:
@navFooh thanks for raising this, good catch. I've created a [separate issue](https://github.com/the-momentum/open-wearables/issues/1491) to track this so this one doesn't need to be blocked

bartmichalak (approved):

backend/app/integrations/celery/tasks/revoke_stale_connections_task.py
KaliszS: Not a significant perf improvement, but a good point anyway. I've added an issue:  #1514

KaliszS: I will implment that later.

KaliszS (approved):
I have added two more issues to RN and Dart SDK repos:
- https://github.com/the-momentum/open_wearables_health_sdk/issues/54
- https://github.com/the-momentum/open-wearables-react-native-sdk/issues/12

PR #1440, opened 2026-08-12
fix: resolve race condition in token refresh using distributed lock

### LINKED ISSUE #1433: Concurrent OAuth refresh can revoke healthy connections (rotating refresh tokens)
## Related

- [#1308](https://github.com/the-momentum/open-wearables/issues/1308) — connections revoked after failed token refresh (Garmin). Same `refresh_failed` revoke path in `BaseOAuthTemplate.refresh_access_token`; root cause there may differ.

## Problem

On OAuth token refresh HTTP 400/401, `BaseOAuthTemplate.refresh_access_token` immediately marks the connection `REVOKED` and emits `connection.revoked` with `reason=refresh_failed`.

That can drop a still-healthy connection when the 400/401 is not a true end-of-life refresh token, but the result of another worker already having spent and rotated it.

## Possible cause

Hypothesis (not confirmed as the only path):

1. Some providers rotate refresh tokens on every successful refresh (notably WHOOP) and invalidate the previous refresh token immediately.
2. Multiple Celery workers can load the same stored refresh token near access-token expiry — e.g. pull sync and webhooks both calling through `_get_valid_token` (5-minute expiry buffer).
3. Both workers call the provider token endpoint with that refresh token. Only one refresh succeeds and persists a new access + refresh token pair.
4. The loser receives HTTP 400/401 for the already-spent refresh token.
5. Today the loser immediately revokes the connection — even though the winner has already written valid tokens to `user_connection`.

Call path that leads to revoke:

```text
BaseOAuthTemplate.refresh_access_token
  → httpx.post(token_url) returns 400/401
  → _revoke_connection(reason="refresh_failed")
```

Relevant upstream-shaped handling:

```python
if e.response.status_code in (HTTP_400_BAD_REQUEST, HTTP_401_UNAUTHORIZED):
    self._revoke_connection(db, user_id, reason="refresh_failed")
    raise HTTPException(...)
```

## Expected behavior

A transient “refresh token already used / invalid” from a concurrent refresh race should not revoke a connection that another worker has already successfully refreshed and persisted.

## Actual behavior

The losing worker revokes the connection and emits `connection.revoked` (`reason=refresh_failed`). The user must reconnect even though a valid token pair may already be stored.

## Workaround Resolution

We hit this in production (WHOOP under concurrent sync/webhook load) and mitigated it locally. On refresh 400/401 we remember the pre-attempt access token, wait briefly, re-read `user_connection` from the DB, and if the access token changed we adopt those tokens instead of revoking. That worked for us.

The wait-and-re-read approach is a bit hacky and may require a more robust fix than what we did.

sehmbimanvir:
## Description

This PR resolves a race condition during concurrent OAuth token refresh requests (e.g., when multiple incoming webhooks for the same user arrive simultaneously).
Previously, concurrent webhooks could trigger redundant token rotation requests with stale refresh tokens, leading to HTTP 400 errors from provider OAuth servers and spurious connection revocations.

**Solution:**
1. **Distributed Lock (`redis_client.lock`)**: Scopes a Redis lock per user and provider (`token_refresh_lock:{provider_name}:{user_id}`) to serialize token refresh attempts.
2. **Session Eviction (`db.expire_all()`)**: Invalidates stale SQLAlchemy identity map objects before fetching the connection inside the lock.
3. **Double-Checked Refresh**: Reloads connection data inside the lock and verifies whether another worker refreshed the token while waiting. If refreshed, it reuses the newly issued access token immediately.
4. **Safety**: Replaced custom polling and deletion with `redis-py`'s built-in lock context manager to ensure safe Lua-backed token-verified release.
### General

### Backend Changes

### Frontend Changes

- [ ] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. Connect any provider. e.g **Whoop**.
2. Go to **user_connections** table and set token expiration date to past.
3. Update anything in provider (e.g I updated my "Whoop Sleep")
4. Now, that Webhook should trigger/update data in open-wearables db.
5. Check in **user_connections** table tokens and expiry should be updated while connection status should be active only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication reliability when multiple requests refresh the same provider connection simultaneously.
  * Prevented unnecessary token refreshes by reusing a valid token refreshed by another process.
  * Added clearer handling for missing or invalid connections and unavailable refresh tokens during authentication.
  * Ensured token refresh operations use the latest connection information before proceeding.

### DIFF
--- a/backend/app/services/providers/api_client.py
+++ b/backend/app/services/providers/api_client.py
@@ -10,10 +10,10 @@ import httpx
 from fastapi import HTTPException, status
 
 from app.database import DbSession
+from app.integrations.redis_client import get_redis_client
 from app.repositories import UserConnectionRepository
 from app.services.providers.templates.base_oauth import BaseOAuthTemplate
 from app.utils.structured_logging import log_structured
-from app.integrations.redis_client import get_redis_client
 
 logger = logging.getLogger(__name__)
 
@@ -59,27 +59,39 @@ def _get_valid_token(
         redis_client = get_redis_client()
         lock_key = f"token_refresh_lock:{provider_name}:{user_id}"
 
-        with redis_client.lock(lock_key, timeout=60, sleep=0.2):
-          # Fetch and refresh connection instance inside the lock
-          connection = connection_repo.get_by_user_and_provider(db, user_id, provider_name)
-          if not connection or not connection.access_token:
-              raise ValueError(f"No valid connection found for {provider_name}:{user_id}")
+        with redis_client.lock(lock_key, timeout=60, blocking_timeout=10, sleep=0.2):
+            # Fetch and refresh connection instance inside the lock
+            connection = connection_repo.get_by_user_and_provider(db, user_id, provider_name)
+            if not connection:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail=f"User not connected to {provider_name}",
+                )
+            if not connection.access_token:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail=f"No access token available for {provider_name} (SDK-based provider?)",
+                )
 
-          db.refresh(connection)
+            db.refresh(connection)
 
-          # 1. If the token never expires, return it directly
-          # 2. If it expires in the future (> 5 min buffer), return existing token
-          now = datetime.now(timezone.utc)
-          if not connection.token_expires_at or connection.token_expires_at >= now + timedelta(minutes=5):
-              return connection.access_token
+            # 1. If the token never expires, return it directly
+            # 2. If it expires in the future (> 5 min buffer), return existing token
+            now = datetime.now(timezone.utc)
+            if not connection.token_expires_at or connection.token_expires_at >= now + timedelta(minutes=5):
+                return connection.access_token
 
-          # Guard against missing refresh token before attempting OAuth refresh
-          if not connection.refresh_token:
-              raise ValueError(f"Missing refresh token for {provider_name}:{user_id}")
+            # Guard against missing refresh token before attempting OAuth refresh
+            if not connection.refresh_token:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail=f"Token expired and no refresh token available for {provider_name}",
+                )
 
-          token_response = oauth.refresh_access_token(db, user_id, connection.refresh_token)
-          return token_response.access_token
-          return connection.access_token
+            token_response = oauth.refresh_access_token(db, user_id, connection.refresh_token)
+            return token_response.access_token
+
+    return connection.access_token
 
 
 def make_authenticated_request(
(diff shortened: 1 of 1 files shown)


backend/app/services/providers/api_client.py
```diff
@@ -53,10 +54,32 @@ def _get_valid_token(
...
+          if not connection.token_expires_at or connection.token_expires_at >= now + timedelta(minutes=5):
+              return connection.access_token
+
+          # Guard against missing refresh token before attempting OAuth refresh
+          if not connection.refresh_token:
+              raise ValueError(f"Missing refresh token for {provider_name}:{user_id}")
+
+          token_response = oauth.refresh_access_token(db, user_id, connection.refresh_token)
+          return token_response.access_token
+          return connection.access_token
```
czajkub: Pretty sure indent here is wrong. With this, the happy path returns None instead of the correct token

backend/app/services/providers/api_client.py
```diff
@@ -53,10 +54,32 @@ def _get_valid_token(
...
             )
-        token_response = oauth.refresh_access_token(db, user_id, connection.refresh_token)
-        return token_response.access_token
+        # Scope distributed lock per user/provider to avoid concurrent refresh race conditions.
 
-    return connection.access_token
+        redis_client = get_redis_client()
+        lock_key = f"token_refresh_lock:{provider_name}:{user_id}"
+
+        with redis_client.lock(lock_key, timeout=60, sleep=0.2):
```
czajkub: shouldnt we have a `blocking_timeout` here alongside `timeout`?

backend/app/services/providers/api_client.py
```diff
@@ -53,10 +54,32 @@ def _get_valid_token(
...
+          # Fetch and refresh connection instance inside the lock
+          connection = connection_repo.get_by_user_and_provider(db, user_id, provider_name)
+          if not connection or not connection.access_token:
+              raise ValueError(f"No valid connection found for {provider_name}:{user_id}")
```
czajkub: are those handled somewhere higher, or will they raise up to the top?

czajkub: also, instead of ValueError we could raise here 400/401 http error, as above in the function

czajkub (changes requested):
The change looks good, although left a few notes to check. Also - looks like your pipeline is failing. For the PR title - im pretty sure oauth is not recognized, so if you change it to `fix: ` it'll be fine, as for the code-quality hook - you can `cd backend && uv run pre-commit run --all-files` to auto-format / see what's wrong.

As for the locking - I'm pretty sure the locks are on /0 db in redis, we might want it changed to /2 later (/1 is used by svix), but thats outside the scope of this PR

backend/app/services/providers/api_client.py
sehmbimanvir: Yes, correct, this was an indentation issue

sehmbimanvir:
> The change looks good, although left a few notes to check. Also - looks like your pipeline is failing. For the PR title - im pretty sure oauth is not recognized, so if you change it to `fix: ` it'll be fine, as for the code-quality hook - you can `cd backend && uv run pre-commit run --all-files` to auto-format / see what's wrong.
> 
> As for the locking - I'm pretty sure the locks are on /0 db in redis, we might want it changed to /2 later (/1 is used by svix), but thats outside the scope of this PR

@czajkub , Thank you for your review, please note that all the comments are resolved.

czajkub (approved):
Thank you for the contribution, looks good now!

PR #1444, opened 2026-08-14
feat(backend): raise upload limit to 5 GiB and add upload progress dialog for uploading Apple's XML

### LINKED ISSUE #1413: Apple Health XML 1GB file limit
I am getting this error when I try to upload my large Apple Health XML export

Image

Is there a way to increase the file upload size?

MichalZak17:
## Description

Frontend and backend enforced a 1 GiB cap on Apple Health XML uploads, while the API docs stated 500 MB — and a user needs to import a ~4 GB export (#1413). This raises the maximum to **5 GiB**, the ceiling for a single S3 presigned POST, so large exports upload without needing multipart. It also aligns the docs and the stale `500MB`/`1GB` code comments with the real value, and fixes an outdated backend-schema path referenced from the frontend constants.

**Changes**
- `MAX_FILE_SIZE`: `1 GiB` → `5 GiB` — enforced by the backend `PresignedURLRequest.max_file_size` `le=` bound and the frontend upload validation
- Backend field description and API docs now read `1KB - 5GiB`
- No change to the 50 MB default `max_file_size` or the 10 MB direct-vs-S3 routing threshold

Closes #1413

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes

## Testing Instructions

**Steps to test:**
1. On a user's detail page, use the **Upload Apple Health XML** action and select an export larger than 1 GiB (up to 5 GiB).
2. Confirm it is accepted — the previous `File is too large … Maximum size is 1GB` rejection no longer fires.
3. Try a file larger than 5 GiB and confirm it is still rejected against the new limit.
4. (API) `POST /api/v1/users/{user_id}/import/apple/xml/s3` now accepts `max_file_size` up to `5368709120`; a value above that returns a 422.

**Expected behavior:**

Files up to 5 GiB are accepted end to end; the presigned POST's `content-length-range` upper bound reflects the requested size. Behavior for files ≤ 1 GiB is unchanged.

## Screenshots

Screenshot 2026-08-14 at 12 29 55

## Additional Notes

5 GiB is the hard limit for a single S3 presigned POST; anything larger would require multipart uploads as explained in [Amazon S3 multipart upload limits](https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html), which is out of scope here. Full suite verified locally before opening: backend `pytest` (1995 passed, 2 skipped) and frontend `vitest` (12 passed), plus `lint`/`format:check`/`build`. The `pre-commit` hook still needs to be run in a local shell.

## Summary by CodeRabbit

- **New Features**
  - Apple Health XML uploads now display real-time progress, file details, transfer status, and success or error messages.
  - Upload progress can be dismissed after completion, while active uploads prevent accidental closure.
- **Improvements**
  - Supported Apple Health XML file size increased to 5 GiB.
  - Progress tracking now works for both standard and S3-based uploads.
- **Documentation**
  - Updated upload size guidance to reflect the new 5 GiB limit.

### DIFF
--- a/backend/app/schemas/providers/apple/apple_xml/aws.py
+++ b/backend/app/schemas/providers/apple/apple_xml/aws.py
@@ -4,7 +4,7 @@ MIN_EXPIRATION_SECONDS = 60  # 1 minute
 MAX_EXPIRATION_SECONDS = 3600  # 1 hour
 DEFAULT_EXPIRATION_SECONDS = 300  # 5 minutes
 MIN_FILE_SIZE = 1024  # 1KB
-MAX_FILE_SIZE = 1024 * 1024 * 1024  # 500MB
+MAX_FILE_SIZE = 5 * 1024 * 1024 * 1024  # 5GiB (S3 presigned POST ceiling for a single upload)
 DEFAULT_FILE_SIZE = 50 * 1024 * 1024  # 50MB
 
 
@@ -20,7 +20,7 @@ class PresignedURLRequest(BaseModel):
         default=DEFAULT_FILE_SIZE,
         ge=MIN_FILE_SIZE,
         le=MAX_FILE_SIZE,
-        description="Maximum file size in bytes (1KB - 500MB)",
+        description="Maximum file size in bytes (1KB - 5GiB)",
     )
 
 
--- a/docs/api-reference/guides/apple-xml-import.mdx
+++ b/docs/api-reference/guides/apple-xml-import.mdx
@@ -135,7 +135,7 @@ const presignedData = await response.json();
 </ParamField>
 
 <ParamField body="max_file_size" type="integer" default="52428800">
-  Maximum file size in bytes (1KB - 500MB). Default is 50MB.
+  Maximum file size in bytes (1KB - 5GiB). Default is 50MB.
 </ParamField>
 
 <ResponseExample>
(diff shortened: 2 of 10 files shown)


bartmichalak (commented):
This feels like a perfect fit for GitHub's new Stacked PRs feature (gh stack) - https://github.com/the-momentum/open-wearables/pull/1445 already builds directly on top of this one, so a stack would give us clean per-layer diffs and let us merge the whole chain in one click.

But since these PRs come from an external fork, I'm not sure the stacked-PRs feature actually works in that setup?

You can take a look @MichalZak17 - if it doesn't work, we'll merge this change first, since it can live independently of the second PR and actually resolves an existing problem and inconsistency.

MichalZak17:
Thanks @bartmichalak! I looked into `gh stack` and stacked PRs won't work for us — GitHub requires all branches to live in the same repository and explicitly notes that cross-fork stacks aren't supported ([docs](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs)), so it's out for a fork contribution.

So let's go with your suggestion: feel free to merge this one first — it's self-contained and resolves the upload-limit/docs inconsistency on its own. I'll rebase #1445 onto `main` afterward and give you a heads-up once its changes are done and ready for merge.

bartmichalak (approved):
Pushed a small change: the progress-based upload now handles a 401 like apiClient does (clears the session and redirects to login), to keep auth handling consistent with the rest of the app.

PR #1445, opened 2026-08-14
feat(backend): multipart upload + MinIO parity for Apple's XML imports

### LINKED ISSUE #439: XML direct upload fails for large files (Redis message size limit)
## Description

The `/api/v1/users/{user_id}/import/apple/xml/direct` endpoint reads the entire uploaded file into memory and passes the raw bytes as a Celery task argument through Redis. For large files (e.g., a typical Apple Health export of ~670MB), this fails because Redis has a message size limit.

## Steps to Reproduce

1. Export Apple Health data from an iPhone (typically 500MB-7GB)
2. Upload via the direct XML endpoint:
   ```bash
   curl -X POST "http://localhost:8000/api/v1/users/{user_id}/import/apple/xml/direct"      -H "X-Open-Wearables-API-Key: {api_key}"      -F "file=@export.xml"
   ```
3. The upload completes but the backend returns `Internal Server Error`

## Error

```
kombu.exceptions.OperationalError: Error 104 while writing to socket. Connection reset by peer.
```

The error occurs at `import_xml.py:35` when calling `process_xml_upload.delay(file_contents=file_contents, ...)` — the entire file content is serialized and sent through Redis as a Celery task argument.

## Suggested Fix

Save the uploaded file to a shared volume accessible by both the `app` and `celery-worker` containers, then pass only the **file path** to the Celery task instead of the raw bytes.

1. Add a shared Docker volume (e.g., `xml_uploads`) mounted at `/tmp/xml_uploads` on both `app` and `celery-worker` services
2. In the endpoint, stream the file to disk using `shutil.copyfileobj()` instead of `await file.read()`
3. Pass `file_path` instead of `file_contents` to `process_xml_upload.delay()`
4. The Celery task already writes to a temp file internally, so it just needs to accept a path instead of bytes

## Environment

- Docker Compose local setup
- Apple Health XML export (~670MB)

### LINKED ISSUE #735: docs(aws-setup): missing S3 CORS config, Celery Worker AWS credentials, and SNS IAM permissions
## Summary

  The AWS setup guide (`docs/dev-guides/aws-setup.mdx`) is missing three required
  configuration steps for the Apple XML upload flow to work end-to-end. All three
  were discovered during a real setup after the SQS→SNS migration (#599).

  ## Missing 1: S3 CORS Configuration

  Browser-based uploads via presigned POST are cross-origin requests. Without a
  CORS policy on the S3 bucket, the browser's preflight OPTIONS request gets a 403
  and the upload never starts — the user sees a generic "Load Failed" error.

  The docs currently have no mention of CORS at all.

  **Required CORS policy** (AWS Console → S3 → `<bucket>` → Permissions → CORS):

  \```json
  [
    {
      "AllowedHeaders": ["*"],
      "AllowedMethods": ["POST"],
      "AllowedOrigins": [
        "https://your-frontend-domain.com"
      ],
      "ExposeHeaders": []
    }
  ]
  \```

  ---

  ## Missing 2: Celery Worker Needs AWS Credentials

  Since #599, the new `process_aws_upload` task downloads the XML file directly
  from S3 (`s3_client.download_file(...)`). The Celery Worker now also needs AWS
  credentials — previously it received file contents directly and needed none.

  In a multi-service deployment (Docker Compose, Railway, etc.) the Celery Worker
  is a separate service and needs the same variables as the backend:

  AWS_ACCESS_KEY_ID
  AWS_SECRET_ACCESS_KEY
  AWS_BUCKET_NAME
  AWS_REGION

  Without these, the task fails with:
  RuntimeError: S3 client not configured — cannot process AWS upload

  ---

  ## Missing 3: IAM User Needs `SNS:ConfirmSubscription`

  The backend calls `sns_client.confirm_subscription()` to confirm the SNS
  subscription, which requires the `SNS:ConfirmSubscription` IAM permission.
  Without it the subscription stays on "Pending confirmation" and the error is not
  mentioned in the troubleshooting section.

  Attach `AmazonSNSFullAccess` to the IAM user, or add a targeted policy:

  \```json
  {
    "Effect": "Allow",
    "Action": "SNS:ConfirmSubscription",
    "Resource": "arn:aws:sns:<region>:<account-id>:<topic-name>"
  }
  \```

  Users who set up before #599 (SQS only) have `AmazonSQSFullAccess` but not
  `AmazonSNSFullAccess` and will silently fail on subscription confirmation.

  ---

  ## Context

  All three issues became relevant with the SQS→SNS migration in #599. The docs
  were not fully updated to reflect the new requirements.

MichalZak17:
## Description

Continuation of #1444 (raise Apple XML upload limit to 5 GiB + progress dialog). Now that #1444 has merged, this targets `main` directly.

Makes the Apple Health XML upload flow work identically against **AWS S3 and a self-hosted MinIO server**, and uploads large exports via **S3 multipart upload**.

Previously MinIO was only wired up for *raw payload storage*; the Apple XML path used a hard-coded AWS S3 client with no custom-endpoint support and relied on **AWS SNS** (which MinIO cannot emit) to trigger processing. This PR closes that gap with a shared endpoint-aware S3 client, a full multipart flow (create → sign → complete → abort), and a storage-agnostic client-driven completion trigger — so a local, AWS-free deployment can ingest multi-gigabyte Apple Health exports through the portal.

### General

### Backend Changes

### Frontend Changes

- [x] `pnpm run lint` passes 

## Testing Instructions

**Steps to test (against local MinIO):**
1. Ensure `backend/config/.env` has `AWS_ENDPOINT_URL=http://minio:9000`, `AWS_PUBLIC_ENDPOINT_URL=http://localhost:9000`, `AWS_ACCESS_KEY_ID=minioadmin`, `AWS_SECRET_ACCESS_KEY=minioadmin`, `AWS_BUCKET_NAME=open-wearables`, and `APPLE_XML_UPLOAD_COMPLETION_MODE=client` (default). `AWS_PUBLIC_ENDPOINT_URL` lets the browser reach the presigned URLs directly, so no `/etc/hosts` edit is needed.
2. Start the stack (`make watch`) and open the portal; upload a large Apple Health `export.xml` (>10 MB) for a user.
3. Watch the parts upload with live progress; confirm the object lands in the MinIO console (`open-wearables` bucket, `<user_id>/raw/…`) and a `process_aws_upload` task runs (Flower/logs).

**Expected behavior:**
The file uploads in parallel parts directly to MinIO, is assembled server-side, and import processing is dispatched — identical to the AWS S3 path, with no AWS account required.

**Automated:**
- Backend: full suite green (2075 passed, 2 skipped); new tests for the S3 client factory, multipart service, multipart API endpoints, SNS dispatch gating, and part-size helper. `ruff` + `ty` clean.
- Frontend: `tsc` clean, `oxlint` 0/0, prettier clean, part-planner vitest 5/5.
- Live E2E against a real MinIO (production code paths + real HTTP PUTs): create → sign → PUT 2 parts → complete → download round-tripped 11 MiB byte-for-byte; MinIO's default CORS already exposes `ETag`.

## Additional Notes

- **Completion mode:** keep `client` (default) for S3 and MinIO; use `sns` only for AWS event-driven setups. The two modes are mutually exclusive so processing never runs twice.
- **CORS:** MinIO exposes `ETag` and allows origins by default — no bucket CORS config needed. On AWS S3, the bucket must allow `PUT` and expose the `ETag` header for browser multipart (see #735).


bartmichalak (commented):
@MichalZak17  first of all, nice job!
I really like what you're suggesting here, because we did make the default setup for importing the XML file quite complex, and in most cases it's not needed, I think. Also, as you noticed, setting this up on localhost wasn't easy because of SNS - and Railway, for example, doesn't offer it either. But it's also good that you kept the SNS config for backward compatibility.

I tested this on our Railway instance and noticed one thing - Celery takes a lot of time to process batches, 30 seconds per batch on average. I think it shouldn't take more than a second. On this instance we have ~26M records, which isn't that much I think. Could you take a look and try to find the bottlenecks? I'd suggest checking either locally (and first seeding some records to get a similar database size) or on Railway.

Image

bartmichalak:
Also, I see backend tests didn't pass

MichalZak17:
~I will take care of that.~

MichalZak17:
@bartmichalak I fixed the backend tests and dug into this locally, using the production XML parser and repository code.

One important detail is that the time between the Yielding chunk messages isn’t Celery scheduling time. That message is logged just before the parser yields a chunk, so the gap includes writing the previous 50k records, committing them, and parsing the next 50k records.

Here are the timings I got:

Test case | Time between equivalent chunk logs
-- | --
Local Docker, small table | ~5.7 seconds
Local Docker, 26M-row table | ~6.4 seconds
Application limited to 0.3 vCPU | ~22.6 seconds
Application limited to 0.25 vCPU | ~34.7 seconds
Railway trace from your screenshot | 20.6–31.3 seconds

The CPU-limited results are interesting because they’re in roughly the same range as the Railway trace. This doesn’t prove that Railway was CPU-constrained, but it does show that limited CPU or contention can reproduce the behaviour you saw. You should be able to confirm that by accessing the metrics. 

As far as I can see 26M rows are not the main problem here. I tested with an isolated 26M-row table, around 6.7 GB including indexes, and the interval only increased from about 5.7 to 6.4 seconds.

I might be wrong, but the main bottleneck seems to be the SQLAlchemy bulk upsert itself. For a 50k-row duplicate batch, I measured roughly:
- 4.7 seconds for the complete repository write
- 1.7 seconds inside PostgreSQL/the driver
- 3 seconds building the statements and processing the results in Python
- Less than 0.01 seconds for the commit

The current code generates seven large `INSERT ... ON CONFLICT` statements. Six of them are around 2.1 million characters long, and then we process 50k returned values in Python.

A few other things can make it worse:
- Unchanged records are updated again, generating around 22–33 MB of WAL per chunk and potentially causing table bloat if autovacuum falls behind.
- Preparing the webhooks generated 44 payloads totalling around 19 MB for one 50k chunk. In my test, this slowed parsing of the next chunk by about 25% (if my math is right).
- XML imports share the same 10-thread Celery worker with the other queues.
- If the same import runs twice concurrently, both processing time and memory usage increase.

Out of curiosity, I tried loading the records with PostgreSQL COPY into a temporary staging table and then doing one merge into the destination table. That reduced the database-write part from around 4.9 seconds to 0.37 seconds for 50k new records. An unchanged duplicate chunk took 0.34 seconds and didn’t rewrite any rows.

Saying so, my suggestion would be to move the import to a COPY + staging-table approach, skip updates when nothing has changed, and give XML imports their own Celery queue with controlled concurrency. It would also be useful to add separate timings for parsing, database writes, commits, and webhook preparation so we can see exactly where the time goes in production.

For your reference, I ran this on a MacBook Air M5 with 16 GB RAM. Docker Desktop had 10 CPUs and 8 GB RAM available, and PostgreSQL 18 was running locally in Docker. The CPU quotas were applied only to the application container; PostgreSQL was left unrestricted and I used a synthetic 500k-record Apple Health XML export.

bartmichalak:
@MichalZak17 thanks for digging into this further. I also took a look at it locally, and together with Claude I basically confirmed what you found and created https://github.com/the-momentum/open-wearables/issues/1453.

@czajkub please take a look at the implementation from a semantic point of view. At a high level, this is a good direction.

BTW, after spending a while with this functionality, I think we could add `minio` to our docker compose.

MichalZak17:
Minio is great, but I just checked and it's no longer maintained. I will take a look later of any alternative that we could use long term for that, but the code itself should handle that anyway.

backend/app/api/routes/v1/import_xml.py
```diff
@@ -28,6 +39,83 @@ def import_xml_presigned_url(
...
+            bucket_name=AWS_BUCKET_NAME,
+            object_key=key,
+            user_id=user_id,
+        )
+        task_id = task.id
+
+    return MultipartCompleteResponse(
+        status="processing" if task_id else "uploaded",
+        key=key,
+        bucket=AWS_BUCKET_NAME,  # ty:ignore[invalid-argument-type]
```
czajkub: It looks like at this point we aren't "sure" the `AWS_BUCKET_NAME` exists - maybe a check if it is `not None` would remove the need for this qa ignore comment?

backend/app/schemas/providers/apple/apple_xml/multipart.py
```diff
@@ -0,0 +1,104 @@
...
+parts are 5 MiB-5 GiB (last part may be smaller), at most 10,000 parts per upload,
+objects up to 5 TiB.
+"""
+
+from math import ceil
+from typing import Annotated
+
+from pydantic import BaseModel, Field
+
+MIN_PART_SIZE = 5 * 1024 * 1024  # 5 MiB (S3 minimum, except the final part)
```
czajkub, line 16: That's a lot of settings. I would move most of this to `config.py` (and maybe `.env.example` by extension) - though it might create a lot of spam, but definitely something like min/max part and object size would belong there.

backend/app/schemas/providers/apple/apple_xml/multipart.py
```diff
@@ -0,0 +1,104 @@
...
+MAX_PART_SIZE = 5 * 1024 * 1024 * 1024  # 5 GiB
+MAX_PARTS = 10_000
+DEFAULT_PART_SIZE = 100 * 1024 * 1024  # 100 MiB
```
czajkub: Is this default based on any benchmarks, or recommendations from some docs? It looks like it would make sense at this size, just asking for good measure

backend/app/services/apple/apple_xml/aws_service.py
```diff
@@ -12,16 +14,16 @@
 
 
 def get_s3_client():  # noqa: ANN201
```
czajkub: Is this function still necessary, if it only checks if settings are filled (which is already done in the `create_s3_client` function below)?

Also ,calling this would create a new instance of a client each time. Maybe we could implement some caching mechanism for it, or one single global instance of the client? You could look at how `get_redis_client()` is implemented for reference

backend/app/services/apple/apple_xml/multipart_upload_service.py
```diff
@@ -0,0 +1,144 @@
...
+                ContentType=content_type,
+            )
+        except ClientError as e:
+            raise self._client_error(e, "Failed to create multipart upload") from e
+
+        self.log.debug(f"Created multipart upload {response['UploadId']} for {key}")
+        return MultipartCreateResponse(
+            upload_id=response["UploadId"],
+            key=key,
+            bucket=AWS_BUCKET_NAME,  # ty:ignore[invalid-argument-type]
```
czajkub: same issue as before - here, I recognise that the `BUCKET_NAME` will never be `None`, but still defensive checks before will never hurt

backend/app/services/apple/apple_xml/aws_service.py
czajkub, line 1: It looks like there are some features implemented here, in `aws_service.py`, and some in `multipart_upload_service.py`. Is there some logical separation between these two, or do they serve similar purposes? Maybe they could be logically separated in some clear way

backend/app/services/apple/apple_xml/multipart_upload_service.py
```diff
@@ -0,0 +1,144 @@
...
+        try:
+            response = client.create_multipart_upload(
+                Bucket=AWS_BUCKET_NAME,
+                Key=key,
+                ContentType=content_type,
+            )
+        except ClientError as e:
+            raise self._client_error(e, "Failed to create multipart upload") from e
+
+        self.log.debug(f"Created multipart upload {response['UploadId']} for {key}")
```
czajkub: I wouldn't use `self.log`, rather `log_structured` from utils, same as in rest of codebase

backend/app/services/apple/apple_xml/multipart_upload_service.py
```diff
@@ -0,0 +1,144 @@
...
+    @staticmethod
+    def _verify_owner(user_id: str, key: str) -> None:
+        """Reject keys that don't belong to the requesting user."""
+        if not key.startswith(f"{user_id}/"):
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail="Object key does not belong to this user",
+            )
+
+    def create(self, user_id: str, filename: str, content_type: str, file_size: int) -> MultipartCreateResponse:
```
czajkub: Maybe we could use more descriptive names, like `create_upload`, `complete_upload` for example? It would make it clearer from a reading standpoint and not mix with other methods from repositories for example.

backend/app/services/apple/apple_xml/multipart_upload_service.py
```diff
@@ -0,0 +1,144 @@
...
+            client.complete_multipart_upload(
+                Bucket=AWS_BUCKET_NAME,
+                Key=key,
+                UploadId=upload_id,
+                MultipartUpload={"Parts": multipart_parts},
+            )
+        except ClientError as e:
+            raise self._client_error(e, "Failed to complete multipart upload") from e
+
+        self.log.debug(f"Completed multipart upload {upload_id} for {key}")
```
czajkub: Same issue with `self.log` here - and below as well

backend/app/services/raw_payload_storage.py
```diff
@@ -64,24 +64,10 @@ def configure(
...
-            kwargs["aws_secret_access_key"] = settings.aws_secret_access_key.get_secret_value()
-        if endpoint_url:
-            kwargs["endpoint_url"] = endpoint_url
-
-        return boto3.client("s3", **kwargs)
-    except (NoCredentialsError, AttributeError, Exception) as e:
-        logger.error("Cannot create S3 client for raw payload storage: %s", e)
-        return None
+    """Create a boto3 S3 client using app AWS settings (shared factory)."""
+    from app.services.s3_client import create_s3_client
```
czajkub: inline import here - can we move it to the top of the module?

backend/app/services/s3_client.py
```diff
@@ -0,0 +1,47 @@
...
+    When ``endpoint_url`` is provided (e.g. a MinIO server), the client is configured
+    with path-style addressing and SigV4 signing so presigned URLs and multipart
+    uploads behave identically to AWS S3.
+
+    Explicit credentials are used when configured; otherwise boto3 falls back to its
+    default credential chain (env vars, instance role, ...). Returns ``None`` if the
+    client cannot be created.
+    """
+    try:
+        import boto3
```
czajkub: again, inline imports - I'd rather have them at the top of the file

backend/app/services/s3_client.py
```diff
@@ -0,0 +1,47 @@
...
+            kwargs["aws_secret_access_key"] = settings.aws_secret_access_key.get_secret_value()
+
+        if endpoint_url:
+            kwargs["endpoint_url"] = endpoint_url
+            # MinIO and most self-hosted S3 servers only support path-style addressing.
+            kwargs["config"] = Config(signature_version="s3v4", s3={"addressing_style": "path"})
+
+        return boto3.client("s3", **kwargs)
+    except (NoCredentialsError, AttributeError) as e:
+        logger.warning("Cannot create S3 client: %s", e)
```
czajkub: another `log_structured` instance here

backend/app/config.py
```diff
@@ -234,6 +234,15 @@ class Settings(BaseSettings):
...
     aws_sns_topic_arn: SecretStr | None = None
+    # Optional custom endpoint for S3-compatible object storage (e.g. MinIO). Also read
+    # natively by boto3 from AWS_ENDPOINT_URL; declaring it here lets the Apple XML flow
+    # force path-style addressing + SigV4 so presigned/multipart URLs work against MinIO.
+    aws_endpoint_url: str | None = None
+    # How a completed S3/MinIO XML upload triggers processing:
+    #   "client" — the frontend calls the /complete endpoint (works for S3 and MinIO alike)
+    #   "sns"    — an S3 bucket event fans out through SNS to /sns/notification (AWS only)
+    # Only the selected mode dispatches the import task, so processing never runs twice.
+    apple_xml_upload_completion_mode: Literal["client", "sns"] = "client"
```
czajkub, line 252: In case we allow more methods in the future, maybe we could put an enum with these values somewhere in the schema for xml? Obviously for now with just `client` and `sns`

czajkub (changes requested):
Hi, great work! I haven't tested it myself but I believe @bartmichalak did - just added some semantic and maintainability comments to address.

For the `minio` storage/setup itself - that probably is in scope for another PR, but as you mentioned - minio is unmaintained for over 6 months and renamed to `AIStor` - maybe we could find an alternative, but that is a topic up to discussion later.

MichalZak17:
Heads-up on where this landed, since it grew a bit beyond the original "multipart + parity" scope while I was working through the review.

First, the storage backend. As I mentioned earlier, [MinIO isn't really maintained anymore](https://github.com/minio/minio), so I didn't want to bake it in as the default for a health-data project we'd have to support long-term. I ended up going with [SeaweedFS](https://github.com/seaweedfs/seaweedfs) as it's actively maintained, S3-compatible, and drops into `docker-compose` as an all-in-one service with the bucket auto-created. The important part is that none of this is MinIO- or SeaweedFS-specific: everything now goes through one endpoint-aware S3 client that forces path-style + SigV4 when a custom endpoint is set, so **AWS S3, SeaweedFS, and MinIO all behave the same**. If someone already has MinIO running, it still works. It's just no longer what we ship by default.

The multipart flow itself is the core of the PR: create → sign → complete → abort, driven from the backend so the browser only ever gets short-lived presigned part URLs. Every call checks that the object key belongs to the requesting user, so nobody can touch someone else's upload. One thing that tripped me up locally was that the browser can't resolve the backend's internal Docker hostname, so I added `AWS_PUBLIC_ENDPOINT_URL` — the backend talks to storage over the internal endpoint, but the presigned URLs handed to the browser use the public one. **That also means no more `/etc/hosts` hack in dev.**

For triggering processing, I kept it behind `APPLE_XML_UPLOAD_COMPLETION_MODE`. The default is `client` (the frontend calls `/complete`), which works everywhere including AWS-free setups. `sns` is still there for AWS event-driven deployments, but the two are mutually exclusive so a file never gets processed twice. Worth flagging for anyone on AWS today: if you currently rely on the SNS bucket event to kick off processing, you'll need to explicitly set `APPLE_XML_UPLOAD_COMPLETION_MODE=sns`, otherwise the client path takes over.

I also added encryption at rest, since we're handling health data. There are two independent layers, both opt-in: `AWS_SSE_ALGORITHM` for storage-side SSE on the browser-direct uploads, and `DATA_ENCRYPTION_KEY` (Fernet, rotation-aware) for the backend-written raw/FIT payloads, so the object store only ever sees ciphertext for those. I deliberately kept both off by default and made them upgrade-safe, so there's no runtime decrypt path in the app (those payloads are write-only), and older plaintext objects stay readable because encrypted ones carry a metadata marker the replay tooling checks. So you can turn it on without any data-migration step; only new writes get encrypted.

On encryption *in transit* I left that out on purpose, and I don't think it belongs in this PR. It's a deployment concern, not app code: you terminate TLS with a reverse proxy in front of the API and the storage endpoint. "Self-hosted" doesn't get us out of it for HIPAA/GDPR once the browser is uploading health data across a network, but the hook is already here. Just point `AWS_PUBLIC_ENDPOINT_URL` at an `https://` host in prod and the presigned URLs inherit that automatically. The only hop that stays on plain HTTP is backend↔storage over the private Docker network. I added a short note about this to the docs.

Last thing: browser multipart uploads that get abandoned halfway leave orphaned parts sitting in the bucket costing money, so there's now a periodic task that aborts incomplete `<user_id>/raw/` uploads older than `APPLE_XML_MULTIPART_MAX_AGE_HOURS`. It only ever looks at incomplete uploads and only within our namespace, so it can't touch finished objects.

### What a migration actually looks like

If you're on **AWS S3**, upgrading is a no-op as everything defaults to the previous behaviour. The one exception is the SNS note above.

For a **self-hosted / AWS-free** setup, it's just config:

```
AWS_ENDPOINT_URL=http://seaweedfs:8333          # internal, backend-side
AWS_PUBLIC_ENDPOINT_URL=http://localhost:8333   # what the browser hits (https in prod)
AWS_ACCESS_KEY_ID=open-wearables
AWS_SECRET_ACCESS_KEY=open-wearables-secret
AWS_BUCKET_NAME=open-wearables
APPLE_XML_UPLOAD_COMPLETION_MODE=client
```

The bundled SeaweedFS service creates the bucket for you. To turn on encryption, add `AWS_SSE_ALGORITHM=AES256` and a `DATA_ENCRYPTION_KEY`. You can generate one with the following command:

```bash
python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"`
```
Existing data stays readable and only new writes are encrypted.

Verification-wise, the backend suite is green locally and `ruff` + `ty` are clean, with new tests around the client factory, the multipart service and endpoints, the SNS gating, the encryption/replay path, and the cleanup task. No DB migration as all the new settings are additive and default off.

backend/app/integrations/celery/core.py
```diff
@@ -150,6 +150,14 @@ def create_celery() -> Celery:
             "args": (),
             "kwargs": {},
         },
+        "cleanup-stale-apple-xml-multipart-uploads": {
+            "task": (
+                "app.integrations.celery.tasks.cleanup_multipart_uploads_task.cleanup_stale_apple_xml_multipart_uploads"
+            ),
+            "schedule": crontab(minute=17),  # Hourly, offset from other periodic work
```
czajkub, line 157: Why 17 here?

czajkub: also - should be configured with a setting, either a previously existing one or a new one if nothing fits well enough

docker-compose.yml
czajkub, line 1: Could you please remove changes related to the SeaweedFS setup, in docker compose and related files? This IMO is a scope for another PR with this forming the base for further work, and we are still discussing the trade-offs between object stores.

czajkub (changes requested):
Hi, updating the review based on the new commit - just a few things this time :P

Also, coderabbit pointed out more issues if you haven't noticed - these should be closed as well

MichalZak17:
Hi. I saw the comments, but didn't have time to finish the changes yet. It should be done today/tomorrow. Also as you requested I will remove the SeaweedFS for now.

docker-compose.yml
bartmichalak: Yeah, something strange happened here - it seems like there are a lot of new changes since I last looked at this PR. @MichalZak17 can you make sure you didn't change anything out of scope after the last review and only responded to @czajkub's comments on the PR?

MichalZak17: Yeah, the scope drifted. While I was working through @czajkub's review I did end up going past the original multipart and parity scope. A few things pulled me there. I started looking for alternatives for MinIO, since it is not maintained anymore, so rather than bake it in I swapped to SeaweedFS, and since we're moving health data around I also added the at-rest encryption and an orphan cleanup task for abandoned uploads. Each of those felt like it followed naturally from the review at the time, but I take the point that together they turned this into a much bigger diff than what was actually asked for.

I will just pull it back. I'll strip out the SeaweedFS setup like you asked and move the encryption and cleanup pieces into their own follow up PRs where they can be reviewed on their own terms later.

backend/app/integrations/celery/core.py
MichalZak17: This is actually gone in the latest commit. I dropped the whole periodic cleanup task in the MinIO-parity rework, so there's no more minute=17 schedule to configure. I will make a separate PR for that like you recommended.

backend/app/services/apple/apple_xml/multipart_upload_service.py
MichalZak17: I see that current version of `xml_service.py` still has ~11 `self.log.info/warning/debug`. Should I migrate those to `log_structured too` as well?

backend/app/schemas/providers/apple/apple_xml/multipart.py
MichalZak17: It's a pragmatic default derived from S3's multipart constraints. The two hard limits are the 5 MiB minimum part size and the 10,000-part cap. At 100 MiB, a typical import sits at a very low part count (the 5 GiB default max works out to ~52 parts), which keeps request overhead and the number of presign round-trips low, while still being small enough that a failed part is cheap to retry. It also lines up with the "~100 MB" part size that AWS's own multipart guidance uses as a general-purpose starting point.

backend/app/schemas/providers/apple/apple_xml/multipart.py
MichalZak17: I did move the operator-tunable knobs to `config.py`: `apple_xml_max_file_size_bytes` (the object-size cap) and `apple_xml_multipart_part_size_bytes` (the default part size), both with the same bounds validation. So the settings someone would actually want to tune per-deployment now live there (and in .env.example).

What I deliberately kept as constants in `multipart.py` are the values that aren't ours to tune like `MIN_PART_SIZE` (5 MiB), `MAX_PART_SIZE` (5 GiB) and `MAX_PARTS` (10,000) are fixed by the S3 multipart protocol itself. Exposing them as settings felt like it'd invite someone to set an invalid value that S3 would just reject, so I treated them as protocol constants and left a comment.

If you'd rather have min/max part in config too for consistency/visibility, I'm happy to move them. I'd keep them clamped to the S3-legal range so they can't be misconfigured. Let me know which you prefer and I'll adjust it.

backend/app/services/apple/apple_xml/aws_service.py
MichalZak17: Actually there's a deliberate separation between the two:

`aws_service.py` holds the stateless storage plumbing that every Apple-XML flow shares like bucket/config validation (`require_bucket_name`), the cached S3/SNS client factories (`get_s3_client`, `get_public_s3_client`, `get_sns_client`), and object-key generation (`build_object_key`). It doesn't run any upload operations itself.

`multipart_upload_service.py` on the other hand is the domain service that orchestrates the multipart lifecycle, so `create_upload`, `sign_upload_parts`, `complete_upload`, `abort_upload` consuming the clients/keys from `aws_service` rather than building them. `presigned_url_service.py `sits at the same layer and reuses the same plumbing essentially.

MichalZak17:
Just fixed the formatting issue that was failing the CI pipeline.

bartmichalak (approved):
I added / modified a few more things, especially around docs, and also cleaned up env.example & the config files.

For the future, please don't change the scope this much after the first review without checking in first 🙏  It adds a massive load on our side during the CR process.

Also, try to keep the PR description concise. These days we're drowning in LLM-generated text that no one actually reads, so we try to avoid that here as much as we can. That's why I went ahead and trimmed the PR description a bit.

Overall, solid chunk of work, appreciate it.

PR #1447, opened 2026-08-16
fix(integrations): retry Strava stream ingestion when the webhook precedes stream processing

slavarosin:
Hi!

Follow-up to #1298, found while debugging why workouts synced without per-sample streams in our production.

## Problem

Strava fires the activity webhook seconds after upload, often before its stream processing has finished. The webhook path saves the workout, but the immediate `/streams` fetch comes back empty - silently, because an empty stream set is a normal response. And since webhook processing bumps `last_synced_at` past the activity's start time, the periodic pull never re-covers that workout. Net effect: every webhook-delivered workout stays permanently without samples. In our prod not a single webhook workout got streams since we enabled `INGEST_WORKOUT_SAMPLES`.

## Fix

When the webhook-path stream ingest lands zero samples (and none exist for the workout window), schedule a delayed retry task with backoff (5m, 15m, 45m, 3 attempts). The task re-checks `has_samples_in_range` before fetching so it is idempotent, handles a deleted workout, and gives up with a warning log after the last attempt.

Stacked on #1298 (uses its `has_samples_in_range` guard) - the first commit here is that PR, only the second one is new. Happy to rebase once #1298 lands.

## Verification

- `uv run pytest tests/providers/strava/` - 25 passed (9 new)
- `ruff check` / `ruff format --check` clean
- Running in our production: previously stream-less workouts now get samples on the delayed retry

Please let me know if you want the backoff timings configurable.

https://claude.ai/code/session_01EY3bZLVL41Q5SHG4uW39Xo

## Summary by CodeRabbit

- **New Features**
  - Added automatic delayed retries when Strava workout streams are temporarily unavailable.
  - Retry timing and maximum attempts can now be configured through application settings.
- **Bug Fixes**
  - Prevented redundant stream fetching when workout samples already exist.
  - Improved handling for missing workouts, empty stream responses, and ingestion failures.
  - Retries now follow the standard workout processing flow for more consistent results.
- **Tests**
  - Added coverage for retry scheduling, backoff behavior, duplicate-fetch prevention, and failure scenarios.

### DIFF
--- a/backend/app/integrations/celery/tasks/strava_stream_retry_task.py
+++ b/backend/app/integrations/celery/tasks/strava_stream_retry_task.py
@@ -18,7 +18,6 @@ from celery import shared_task
 
 from app.database import DbSession, SessionLocal
 from app.schemas.providers.strava import ActivityJSON as StravaActivityJSON
-from app.services.providers.factory import ProviderFactory
 from app.services.timeseries_service import timeseries_service
 from app.utils.sentry_helpers import log_and_capture_error
 from app.utils.structured_logging import log_structured
@@ -37,10 +36,18 @@ def retry_strava_stream_ingest(user_id: str, activity_id: str, attempt: int = 1)
         with SessionLocal() as db:
             _retry_ingest(db, UUID(user_id), str(activity_id), attempt)
     except Exception as e:
+        # A transient failure (detail fetch, token refresh, DB hiccup) must not
+        # end the retry sequence — keep rescheduling until MAX_ATTEMPTS.
         log_and_capture_error(e, logger, "Strava stream retry failed", extra={"activity_id": activity_id})
+        if attempt < MAX_ATTEMPTS:
+            schedule_stream_retry(user_id, str(activity_id), attempt + 1)
 
 
 def _retry_ingest(db: DbSession, user_uuid: UUID, activity_id: str, attempt: int) -> None:
+    # Imported lazily: factory -> strategies -> strava modules; a module-level
+    # import here would participate in that cycle via the tasks package init.
+    from app.services.providers.factory import ProviderFactory
+
     workouts = ProviderFactory().get_provider("strava").workouts
     record = workouts.workout_repo.get_by_external_id(db, user_uuid, activity_id, source="strava")
     if record is None:
--- a/backend/app/integrations/celery/tasks/strava_stream_retry_task.py
+++ b/backend/app/integrations/celery/tasks/strava_stream_retry_task.py
@@ -1,0 +1,122 @@
+"""Delayed retry for Strava per-sample stream ingestion.
+
+Strava fires the activity webhook seconds after upload, before stream
+processing on their side has finished — the immediate /streams fetch in the
+webhook path then comes back empty and the workout is saved without samples.
+The hourly pull can't repair it either: the webhook bumps last_synced_at past
+the activity's start time, so the pull window never re-covers the workout.
+
+This task re-fetches the streams a few minutes later, when Strava has
+finished processing. It is idempotent (skips if samples already exist) and
+self-limits to MAX_ATTEMPTS.
+"""
+
+from logging import getLogger
+from uuid import UUID
+
+from celery import shared_task
+
+from app.database import DbSession, SessionLocal
+from app.schemas.providers.strava import ActivityJSON as StravaActivityJSON
+from app.services.providers.factory import ProviderFactory
+from app.services.timeseries_service import timeseries_service
+from app.utils.sentry_helpers import log_and_capture_error
+from app.utils.structured_logging import log_structured
+
+logger = getLogger(__name__)
+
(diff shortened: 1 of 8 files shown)


backend/app/services/providers/strava/workouts.py
```diff
@@ -432,6 +445,23 @@ def process_push_activity(
...
+        # Schedule a delayed retry so the samples land once Strava is done.
+        if (
+            ingested == 0
+            and settings.ingest_workout_samples
+            and record.end_datetime is not None
+            and not timeseries_service.has_samples_in_range(
+                db, user_id, "strava", record.start_datetime, record.end_datetime
+            )
+        ):
+            from app.integrations.celery.tasks.strava_stream_retry_task import schedule_stream_retry
```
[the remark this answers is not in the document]
slavarosin: Moving it to module scope breaks import: anything imported from the celery tasks package runs tasks/__init__, which imports tasks that import the provider factory, and the factory imports this module. Kept the local import and added a comment explaining the cycle (same pattern as outgoing_webhooks/events._dispatch). The retry rescheduling on transient failures is fixed in 38aa0911.

backend/app/integrations/celery/tasks/strava_stream_retry_task.py
```diff
@@ -0,0 +1,129 @@
...
+
+from app.database import DbSession, SessionLocal
+from app.schemas.providers.strava import ActivityJSON as StravaActivityJSON
+from app.services.timeseries_service import timeseries_service
+from app.utils.sentry_helpers import log_and_capture_error
+from app.utils.structured_logging import log_structured
+
+logger = getLogger(__name__)
+
+MAX_ATTEMPTS = 3
```
czajkub, line 27: maybe this could be somehow integrating with already existing retry logic? (if any exists in codebase as of now, I'm not sure). Either way - I believe these sorts of constants belong in some setting anyways.

backend/app/integrations/celery/tasks/strava_stream_retry_task.py
```diff
@@ -0,0 +1,129 @@
...
+            logger,
+            "info",
+            "Strava stream retry: samples already ingested, nothing to do",
+            provider="strava",
+            action="stream_retry_already_done",
+            activity_id=activity_id,
+        )
+        return
+
+    activity_data = workouts.get_workout_detail_from_api(db, user_uuid, activity_id)
```
czajkub, line 77: Do we have to rewrite the entire flow here? Can't we use the `workouts` module methods instead of this? Doing it this way could pose issues in the future, when we apply changes to Strava but this task will stay the same way

czajkub (changes requested):
Looks fine enough, with some issues raised - once these are addressed and #1298 is merged, I'm happy to merge this one too.

Also, one more thing - for cases like these, where you build on top of another PR, you can use stacked PRs - a new github feature, Claude should also be able to use it properly via the `gh` CLI

backend/app/integrations/celery/tasks/strava_stream_retry_task.py
slavarosin: Done in 7413a9f - moved to settings: STRAVA_STREAM_RETRY_COUNTDOWN_SECONDS (default [300, 900, 2700], accepts JSON or comma-separated). The attempt budget is just the list length, so it's one knob.

On existing retry logic: celery-native autoretry (send_email_task, renew_oura_webhooks_task use it) fits exception-driven retries, but here the first schedule comes from the webhook request path (no task context), and 'streams not ready yet' is a normal empty response rather than an exception. So the explicit countdown scheduling stays, now settings-driven.

backend/app/integrations/celery/tasks/strava_stream_retry_task.py
slavarosin: Done in 7413a9f - the task no longer reimplements the flow. It just calls get_workout_detail_from_api + process_push_activity, i.e. exactly what the webhook handler does. process_push_activity got a stream_retry_attempt parameter and owns the reschedule/give-up decision itself, so any future change to the Strava flow applies to retries automatically.

slavarosin:
Hi!

Both points addressed in 7413a9f:

- backoff constants -> settings (STRAVA_STREAM_RETRY_COUNTDOWN_SECONDS, attempt budget = list length)
- the retry task now reuses the regular webhook flow (get_workout_detail_from_api + process_push_activity) instead of poking internals; process_push_activity owns the reschedule/give-up decision via a stream_retry_attempt parameter

Tried the stacked PRs tip via gh stack - nice workflow, but it hit two walls here: the feature isn't enabled on this repository yet, and a fork-based PR can't take another fork branch as base (gh reported "Stacked PRs are not enabled for this repository" plus a 422 on the base update). If you enable the preview on the repo I'm happy to restack next time.

Ready for another look. Once #1298 lands I'll rebase this one onto main.

PR #1452, opened 2026-08-18
fix: split HRV into RMSSD/SDNN and deprecate recovery_score in recovery summaries endpoint

LouisFerger-Andrews:
## Description

Fixes a semantic mismatch in the recovery summaries API for WHOOP data.

WHOOP provides HRV as `hrv_rmssd_milli`, and Open Wearables already classifies this internally as `heart_rate_variability_rmssd`. However, the recovery-summary response exposed the same value through `avg_hrv_sdnn_ms`.

RMSSD and SDNN are distinct HRV measures. This change adds `avg_hrv_rmssd_ms` to the response and maps WHOOP RMSSD data to that field instead.

### General

- [x] No docs update needed

### Backend Changes

### Frontend Changes

- [x] Not applicable

## Testing Instructions

**Steps to test:**

1. From the repository root, enter the backend directory:

   ```bash
   cd backend
   ```

2. Install the locked development environment:

   ```bash
   uv sync --group dev
   ```

3. Run the new regression test:

   ```bash
   uv run pytest tests/services/test_summaries_service.py -k whoop_rmssd
   ```

4. Run the full backend test suite:

   ```bash
   uv run pytest
   ```

5. Run project checks:

   ```bash
   uv run pre-commit run --all-files
   ```

**Expected behavior:**

- A WHOOP recovery summary with `hrv_rmssd_milli: 63.2` returns:

  ```json
  {
    "avg_hrv_sdnn_ms": null,
    "avg_hrv_rmssd_ms": 63.2
  }
  ```

- No WHOOP RMSSD value is exposed as SDNN.

## Screenshots

Not applicable; this is a backend API response correction.

## Additional Notes

This does not change WHOOP ingestion or stored source data. It corrects the recovery-summary API contract so that RMSSD and SDNN remain semantically distinct.

## Summary by CodeRabbit

* **New Features**
  * Recovery summaries now include average HRV RMSSD measurements.

* **Bug Fixes**
  * Corrected HRV data mapping so RMSSD values appear in the appropriate field instead of being reported as SDNN.

* **Documentation**
  * Clarified provider limitations and when SDNN may be unavailable.
  * Marked recovery scores as deprecated and directed users to the health scores endpoint.

### DIFF
--- a/backend/app/api/routes/v1/summaries.py
+++ b/backend/app/api/routes/v1/summaries.py
@@ -67,7 +67,17 @@ def get_recovery_summary(
     cursor: str | None = None,
     limit: Annotated[int, Query(ge=1, le=100)] = 50,
 ) -> PaginatedResponse[RecoverySummary]:
-    """Returns daily recovery metrics (recovery score, HRV, resting HR, SpO2)."""
+    """Returns daily recovery metrics (recovery score, HRV, resting HR, SpO2).
+
+    **Warning - known limitation:** this endpoint currently returns data **only for WHOOP**.
+    Metrics are read from stored recovery-score records, which today are produced solely by
+    WHOOP, so users connected only to other providers (e.g. Apple Health) receive an empty
+    result even when the underlying resting HR, HRV and SpO2 are available. This is a bug.
+
+    These values will soon be computed from the timeseries we already store in the database
+    rather than from what a single provider reports, at which point recovery metrics will be
+    returned for all providers that supply the underlying data.
+    """
     start_datetime = parse_query_datetime(start_date)
     end_datetime = parse_query_datetime(end_date)
     return summaries_service.get_recovery_summaries(db, user_id, start_datetime, end_datetime, cursor, limit)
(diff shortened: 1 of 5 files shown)


bartmichalak (approved):
Hey, thanks for the PR - great catch. 

Verifying it surfaced a deeper issue, so I pushed a follow-up commit onto your branch:

The endpoint currently returns rows only for WHOOP. A row is only returned if a recovery-score record exists, and only WHOOP produces one - so Apple/other users get an empty result even though their resting HR, HRV and SpO2 are already in our timeseries. Same root cause as the mismatch you fixed: metrics are hostages of the recovery-score record.

What I added to it:
- Kept both HRV fields - avg_hrv_rmssd_ms carries the value; avg_hrv_sdnn_ms stays as a (null for now) placeholder for real SDNN once metrics come from timeseries.
- OpenAPI warning that the endpoint is WHOOP-only today (known bug) and will soon be timeseries-based.
- Deprecated recovery_score (only WHOOP has one, we don't compute our own recovery score) -> pointed to GET /users/{user_id}/health-scores.
- Renamed the test to be provider-neutral (again, we tend to be unified layer)

PR #1455, opened 2026-08-19
chore: Remove old data migrations ran during startup

czajkub:
Remove old data migrations ran during startup - just about it

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated startup migration handling to retain the Oura HRV relabeling step.
  * Failed Oura HRV relabeling attempts remain non-blocking and can retry during a later startup.

### DIFF
--- a/backend/scripts/start/app.sh
+++ b/backend/scripts/start/app.sh
@@ -25,17 +25,6 @@ uv run python scripts/init/seed_admin.py
 echo 'Initializing series type definitions...'
 uv run python scripts/init/seed_series_types.py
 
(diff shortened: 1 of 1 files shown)


bartmichalak (approved):
Yeah, let's do it.

I also noticed yesterday on one instance that some of it takes a long time (around a minute?) on the instance where we have ~20M records.

BTW I noticed that for some reason we don't have them described in this section of the docs:

Image

(which I'm not sure is valuable in the public docs TBH)

czajkub:
Thats weird, but now the script wont be on startup and affect times at least. also the data migration in docs topic is probably something that should be touched upon later.

Either way, merging this for now

PR #1460, opened 2026-08-20
feat(backend): offload SDK upload payload to S3, enqueue only a reference

bartmichalak:
Passing the full upload body as a Celery task argument serializes it into the Redis broker queue, so every queued message carries several MB. A backlog therefore grows in proportion to payload size and steadily eats broker memory. How quickly it reaches `maxmemory` - and whether the worker drops into an OOM crash-loop, as it did in production - depends not only on the Redis instance's RAM but on the load driving the queue: 
- how large a range of historical data is being backfilled, 
- for how many users at once. 

Enqueuing only an S3 reference keeps each broker message at ~1 KB, so a backlog no longer scales with payload size.

## What this changes

The SDK sync endpoint now enqueues an S3 reference to the payload instead of the payload itself, and the worker loads the body from S3 when the task runs. The payload was already being written to S3, so this removes a duplicate copy from the broker.

**This is just a draft for now, still needs some thinking.**

## Summary by CodeRabbit

* **New Features**
  * Added optional S3 offloading for SDK sync payloads.
  * Deferred processing can retrieve and clean up offloaded payloads.
  * Added dedicated bucket configuration with startup validation.
  * Added clearer handling for missing payloads and storage configuration issues.

* **Bug Fixes**
  * SDK sync requests now return a service-unavailable response when payload persistence fails.
  * Payload references are passed correctly for deferred processing.
  * Offloaded payloads are retained when processing fails.

* **Documentation**
  * Documented S3 offloading requirements, behavior, cleanup, and retry handling.

### DIFF
--- a/backend/app/api/routes/v1/sdk_sync.py
+++ b/backend/app/api/routes/v1/sdk_sync.py
@@ -7,7 +7,7 @@ from fastapi import APIRouter, HTTPException, status
 from app.integrations.celery.tasks.process_sdk_upload_task import process_sdk_upload
 from app.schemas.providers.mobile_sdk import SyncRequest
 from app.schemas.responses.upload import UploadDataResponse
-from app.services.raw_payload_storage import store_raw_payload
+from app.services.raw_payload_storage import s3_enabled, store_raw_payload
 from app.utils.api_utils import inline_schema_defs
 from app.utils.auth import SDKAuthDep
 from app.utils.structured_logging import log_structured
@@ -107,7 +107,10 @@ def sync_sdk_data(
 
     content_str = json.dumps(body)
 
-    store_raw_payload(
+    # Persist the raw payload and pass only a reference to the worker so the (potentially
+    # multi-MB) body never travels through the Celery/Redis broker - that broker bloat is
+    # what drove Redis to maxmemory. When S3 storage is off (dev), fall back to inline content.
+    payload_ref = store_raw_payload(
         source="sdk",
         provider=provider,
         payload=content_str,
@@ -115,12 +118,31 @@ def sync_sdk_data(
         trace_id=batch_id,
     )
 
+    if s3_enabled() and payload_ref is None:
+        # S3 is the payload channel here; a missing key means the write failed or the payload
+        # exceeded the size limit. Fail loudly rather than silently falling back to inline
+        # content (which would re-introduce the broker bloat for exactly the large payloads).
+        log_structured(
+            logger,
+            "error",
+            "Failed to persist SDK payload to S3; rejecting batch",
+            action="sdk_payload_persist_failed",
+            batch_id=batch_id,
+            user_id=user_id,
+            provider=provider,
+        )
+        raise HTTPException(
+            status_code=status.HTTP_503_SERVICE_UNAVAILABLE,
+            detail="Failed to persist payload; please retry.",
+        )
+
     process_sdk_upload.delay(
-        content=content_str,
+        content=None if payload_ref else content_str,
         content_type="application/json",
         user_id=user_id,
         provider=provider,
         batch_id=batch_id,
+        payload_ref=payload_ref,
     )
 
     return UploadDataResponse(status_code=202, response="Import task queued successfully", user_id=user_id)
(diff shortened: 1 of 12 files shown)


Amirzhz:
Field report in support of this change — we hit exactly the failure mode the code comment describes ("falling back to an inline body is what fills the broker until Redis hits maxmemory"), on a self-hosted deployment running 0.7.0, which predates this PR by two weeks.

An Android phone re-exported its full Health Connect history: 29 batches, 644,724 records, in five minutes. The SDK chunks at 2,000 Health Connect *records* per round, but a `HeartRateRecord` flattens to one unified record per sample, so the batches that actually arrive are 14k–52k records / ~10 MB each. With the payload inline in the task message, the broker went from 22 MB to 407 MB. Redis hit maxmemory, and because the queue declaration does an `LLEN`, the worker then failed at boot on every restart — so nothing could drain it. All SDK syncs for all users 500'd for 24 hours, and three batches were lost.

Two numbers that might be useful for sizing:

- **Import throughput ≈ 445 records/s.** Measured on 0.7.0 — 62 records in 0.20 s, 50,000 records in 112 s, single worker, `concurrency: 2`, Postgres on a small instance. So 645k records is ~24 minutes of import time delivered in 5 minutes of upload time. A ~5× producer/consumer overrun is enough to fill a 0.5 GB broker from one phone.
- **The queue survived the eviction, the results did not.** With the default `volatile-lru`, Redis evicted the ~20k keys that had a TTL (task results) and left the queue lists, which have none. Worth knowing for anyone tempted to switch to `allkeys-lru` for headroom — that would silently delete queued health data instead.

Question: is this targeted at a tagged release? `latest` on Docker Hub is from 2026-08-26 09:42 UTC, about four hours before this merged, so the newest published non-nightly image doesn't carry it. We pin by digest rather than tracking `nightly`, so we're waiting on a tag before we can turn `SDK_PAYLOAD_S3_OFFLOAD` on. Happy to report back once we're running it with the offload enabled.

KaliszS:
> Field report in support of this change — we hit exactly the failure mode the code comment describes ("falling back to an inline body is what fills the broker until Redis hits maxmemory"), on a self-hosted deployment running 0.7.0, which predates this PR by two weeks.
> 
> An Android phone re-exported its full Health Connect history: 29 batches, 644,724 records, in five minutes. The SDK chunks at 2,000 Health Connect _records_ per round, but a `HeartRateRecord` flattens to one unified record per sample, so the batches that actually arrive are 14k–52k records / ~10 MB each. With the payload inline in the task message, the broker went from 22 MB to 407 MB. Redis hit maxmemory, and because the queue declaration does an `LLEN`, the worker then failed at boot on every restart — so nothing could drain it. All SDK syncs for all users 500'd for 24 hours, and three batches were lost.
> 
> Two numbers that might be useful for sizing:
> 
>     * **Import throughput ≈ 445 records/s.** Measured on 0.7.0 — 62 records in 0.20 s, 50,000 records in 112 s, single worker, `concurrency: 2`, Postgres on a small instance. So 645k records is ~24 minutes of import time delivered in 5 minutes of upload time. A ~5× producer/consumer overrun is enough to fill a 0.5 GB broker from one phone.
> 
>     * **The queue survived the eviction, the results did not.** With the default `volatile-lru`, Redis evicted the ~20k keys that had a TTL (task results) and left the queue lists, which have none. Worth knowing for anyone tempted to switch to `allkeys-lru` for headroom — that would silently delete queued health data instead.
> 
> 
> Question: is this targeted at a tagged release? `latest` on Docker Hub is from 2026-08-26 09:42 UTC, about four hours before this merged, so the newest published non-nightly image doesn't carry it. We pin by digest rather than tracking `nightly`, so we're waiting on a tag before we can turn `SDK_PAYLOAD_S3_OFFLOAD` on. Happy to report back once we're running it with the offload enabled.

Hi, `latest` tag will be pinned to the newest release (`0.7` right now), we are goign to push new release (`0.8`) next Friday. If you want to test those changes earlier, just use `nightly` tag, it's updated every night based on main branch :) Let us know whether it solved your issues after testing with nightly release!

PR #1464, opened 2026-08-22
fix(backend): fix recovery endpoint and resolve cycle_id for recovery.updated

sehmbimanvir:
## Description

**Problem**
- Webhook Payload Identifier: According to Whoop's developer documentation, the recovery.updated webhook payload contains a sleep_id in its resource field rather than a direct recovery or cycle ID.
- Incorrect Recovery API Endpoint: Whoop does not expose a /v2/recovery/{id} endpoint for fetching a single recovery record. The correct API path is /v2/cycle/{cycle_id}/recovery.
- Previously, WhoopWebhookHandler attempted to call load_single_recovery directly with the webhook's resource_id against the non-existent /v2/recovery/{cycle_id} endpoint, resulting in failed webhook processing.

**Solution**
- Resolve cycle_id via Sleep Record: When handling WhoopWebhookNotificationType.RECOVERY_UPDATED, fetch the sleep record using the webhook's resource_id (sleep_id) first to obtain its associated cycle_id.
- Correct Recovery Endpoint: Updated Whoop data247.get_recovery_record to request /v2/cycle/{cycle_id}/recovery.

### General

- [X] My code follows the project's code style
- [X] I have performed a self-review of my code
- [X] I have added tests that prove my fix/feature works (if applicable)
- [X] New and existing tests pass locally
- [X] No Documentation Update Required 

### Backend Changes

- [X] `uv run pre-commit run --all-files` passes

## Testing Instructions

**Steps to test:**
1. Connect Whoop Provider
2. In Whoop App update your sleep or see when new sleep record is being added
3. in data_point_series table check any data_definition related to recovery e.g resting_heart_rate.

**Expected behavior:**
A record for that date for which you updated the sleep should exists in data_point_series table.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery updates from Whoop webhooks by associating recovery data with the correct sleep cycle.
  * Corrected recovery retrieval to use the appropriate cycle-based endpoint and matching trace information.
  * Added validation for missing or malformed sleep cycle information to prevent invalid recovery updates.

### DIFF
--- a/backend/app/services/providers/whoop/data_247.py
+++ b/backend/app/services/providers/whoop/data_247.py
@@ -858,14 +858,14 @@ class Whoop247Data(Base247DataTemplate):
         user_id: UUID,
         cycle_id: str,
     ) -> dict[str, Any]:
-        """Fetch a single recovery record by cycle_id from /v2/recovery/{cycle_id}."""
-        response = self._make_api_request(db, user_id, f"/v2/recovery/{cycle_id}")
+        """Fetch a single recovery record by cycle_id from /v2/cycle/{cycle_id}/recovery."""
+        response = self._make_api_request(db, user_id, f"/v2/cycle/{cycle_id}/recovery")
         store_raw_payload(
             source="api_response",
             provider="whoop",
             payload=response,
             user_id=str(user_id),
-            trace_id=f"/v2/recovery/{cycle_id}",
+            trace_id=f"/v2/cycle/{cycle_id}/recovery",
         )
         return response if isinstance(response, dict) else {}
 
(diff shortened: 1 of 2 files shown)


czajkub (commented):
Hi, this looks fine semantically - but I'd like some clarification here.

I assume you've tested this a bit and it works - but when exactly are the `recovery_updated` webhooks sent? In the docs, it says it's related to a sleep id - so should we only sync the sleep record, or the cycle as well? (I see that here the cycle is also collected). Does the cycle span the time period when the sleep occured, or are they separate from each other?

sehmbimanvir:
@czajkub, 

> Hi, this looks fine semantically - but I'd like some clarification here.
> 
> I assume you've tested this a bit and it works - but when exactly are the `recovery_updated` webhooks sent? In the docs, it says it's related to a sleep id - so should we only sync the sleep record, or the cycle as well? (I see that here the cycle is also collected). Does the cycle span the time period when the sleep occured, or are they separate from each other?

recovery.update will always be called whenever sleep will be updated or created. If you see, I attached my logs from server which are listening Webhooks directly from whoop.
image


And answer to your another question is that, cycle data doesn't change if I sleep/recovery is being updated So, I don't think so we need to keep it with recovery or sleep update.
And anyways, we are not consuming cycle/{cycleId} endpoint in webhooks for Whoop.

czajkub (approved):

PR #1469, opened 2026-08-23
fix(backend): report real synced item counts for Ultrahuman

### LINKED ISSUE #1468: Ultrahuman sync reports 0 items despite saving data
Ultrahuman Live Syncs complete successfully but always show "0 items", even when records are saved (see screenshot below).

Image

bartmichalak:
## Description

Ultrahuman syncs were reporting "0 items" even when data was saved, because the implementation returned a different structure than the sync orchestrator expects to count (`WriteCounts`). Now, the Ultrahuman implementation follows the same shared `bulk_create_samples` upsert path as the other providers, which gives us real insert/update counts and deduplication on re-sync for free.

### General

### Backend Changes

## Testing Instructions

**Steps to test:**
1. Run `tests/providers/ultrahuman/test_ultrahuman_integration.py::TestUltrahumanSyncCountContract`
2. Or trigger a live Ultrahuman sync and check the reported item count

**Expected behavior:**
A successful sync reports the actual number of saved records instead of `0 items`, and re-syncing the same day upserts in place rather than duplicating rows.

**Verified, and what's more, it's now way faster!** 

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved activity and sleep data synchronization to update existing records instead of creating duplicates.
  * Sync results now accurately distinguish inserted and updated records.
  * Failed days and errors are excluded from saved-record totals.
  * Malformed activity entries are safely skipped, improving synchronization reliability.

* **Tests**
  * Added coverage for synchronization counts and repeated activity imports.
  * Verified that resynchronizing data does not create duplicate records.

### DIFF
--- a/backend/app/services/providers/ultrahuman/data_247.py
+++ b/backend/app/services/providers/ultrahuman/data_247.py
@@ -9,9 +9,9 @@ from fastapi import HTTPException
 
 from app.config import settings
 from app.database import DbSession
-from app.models import DataPointSeries, EventRecord
+from app.models import EventRecord
 from app.repositories import EventRecordRepository, UserConnectionRepository
-from app.repositories.data_point_series_repository import DataPointSeriesRepository
+from app.repositories.data_point_series_repository import WriteCounts
 from app.schemas.enums import daily_total_flag
 from app.schemas.enums.series_types import SeriesType
 from app.schemas.model_crud.activities.data_point_series import TimeSeriesSampleCreate
@@ -23,6 +23,7 @@ from app.services.providers.templates.base_247_data import Base247DataTemplate
 from app.services.providers.templates.base_oauth import BaseOAuthTemplate
 from app.services.providers.ultrahuman.coverage import ACTIVITY_SAMPLE_SERIES
 from app.services.raw_payload_storage import store_raw_payload
+from app.services.timeseries_service import timeseries_service
 
 
 class Ultrahuman247Data(Base247DataTemplate):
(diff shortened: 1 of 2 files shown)


backend/app/services/providers/ultrahuman/data_247.py
```diff
@@ -374,50 +374,46 @@ def normalize_activity_samples(
...
+                            is_daily_total=daily_total_flag(series_type, is_daily=False),
+                        )
                     )
-
-                    self.data_point_repo.create(db, ts_sample)
-                    count += 1
                 except Exception as e:
-                    # Log but continue for other samples
-                    # Use warning level for first few errors to help debug issues
                     self.logger.warning(
```
czajkub: maybe we could as well update that to `log_structured`, and in other places that use raw logger? Though it is inconsequential, it might as well be another issue/pr scope

backend/app/services/providers/ultrahuman/data_247.py
```diff
@@ -565,6 +567,9 @@ def load_and_save_all(
 
             current_date += timedelta(days=1)
 
+        results["sleep_sessions_synced"] = WriteCounts(results["sleep_sessions_synced"], 0)
+        results["activity_samples"] = WriteCounts(activity_inserted, activity_updated)
```
czajkub, line 623: I see that the return of `load_and_save_all` is still a raw dict and not normalized accross all providers (at least not to a pydantic schema) - this is out of scope here, but I'll flag it to look at it later and if anything could change there

czajkub (approved):
Looks good - flagged some separate issues, but everything here is fine to merge IMO

backend/app/services/providers/ultrahuman/data_247.py
bartmichalak: very good find, thanks 💯  I added instructions to agents.md, it should eliminate similar inconsistencies in the future
https://github.com/the-momentum/open-wearables/pull/1481

I also fixed the way of logging in this PR - I think there is no need to break it down further

Issue #1453, opened 2026-08-18
perf(backend): Accelerate Apple XML import


bartmichalak:
## Summary

Importing large Apple Health `export.xml` files (`process_aws_upload` task) is slow:
each 50k-record chunk takes ~15-30s to persist. Profiling shows the bottleneck is **not** PostgreSQL, table size, or XML parsing - it is the client-side cost of the multi-row `INSERT ... ON CONFLICT` upsert in `DataPointSeriesRepository._insert_data_points` (`backend/app/repositories/data_point_series_repository.py`).

This issue captures a reproduction + profiling and proposes a fix. It confirms and builds on @MichalZak17's investigation in https://github.com/the-momentum/open-wearables/pull/1445#issuecomment-5332572470.

Sample XML file (~1GB): https://www.kaggle.com/competitions/kaggle-pog-series-s01e04/data

## Reproduction

- File: real 1.8 GiB `export.xml`, chunk size 50k (`XML_CHUNK_SIZE`)
- Table `data_point_series`: 1.25M rows, 2 unique btree indexes, 332 MB total
- Driver: `psycopg` (v3.3.4); Postgres in Docker

Per-phase timing over 8 chunks (production code path):

| Phase | Share |
|-------|-------|
| XML parse (`parse_xml`) | ~15% |
| `bulk_create_samples` (DB write) | **~85%** |
| commit | ~0% |

Drill-down of the DB write:

| Sub-phase | Share |
|-----------|-------|
| `_resolve_data_sources` | ~1% |
| `_insert_data_points` | **~99%** |

## Root cause

Splitting `_insert_data_points` into server vs client time for one ~8k-row batch:

| Measurement (8000 rows) | Time |
|-------------------------|------|
| **Postgres server** (EXPLAIN ANALYZE, DO UPDATE upsert) | **0.32s** (~24.7k rows/s) |
| **Python / SQLAlchemy** - same upsert, execute only | **2.2-4.4s** |

The engine does the upsert in ~0.3s. The remaining ~2-4s is **client-side overhead**: each statement is a single `INSERT ... VALUES` carrying **~64,000 bound parameters** (8000 rows × 8 columns; ~2.1M characters of SQL), which psycopg3 must adapt, serialize and ship. `RETURNING (xmax = 0)` adds ~0.7s more to transfer and materialize 8000 rows back into Python. A 50k chunk = 7 such statements (batch size 8191).

[...]

PR #1492, opened 2026-08-27
perf(backend): rewrite DataPointSeries bulk upsert to use COPY + staging merge

McArthur955:
## Description

Before `DataPointSeriesRepository._insert_data_points` was creating a `INSERT ... VALUES (...)` statement per chunk (≤8191 rows) with huuuuuge amount of bind parameters, which was super slow.
Now we create a single `COPY` statement into a temporary staging table, avoiding those bind parameters, to then merge into the target table.
Addresses #1453.

## What changed

- `_insert_data_points` now: `COPY`s the batch into a `TEMP` staging table, then does one AND ONLY ONE `INSERT INTO data_point_series SELECT ... FROM staging ON CONFLICT DO UPDATE ... RETURNING (xmax = 0)` to merge and split inserted/updated counts - no loop over chunks needed.
- Removed `_INSERT_COLUMNS_PER_ROW` / `BATCH_INSERT_CHUNK_SIZE` from the repository — `COPY` has no per-statement bind-parameter limit, so we don't need those anymore.
- New celery worker for xml imports has been added

#### Local test
| Batch size | Variant | average | median | speedup |
|---|---|---|---|---|
| 50,000 | old code (INSERT) | 6.10s | - | 1.0x |
| 50,000 | new code (COPY) | 1.67s | 1.44s | 3.7x |

#### Railway test (with COPY)
| Batch size | Variant | average | median | min | max | standard deviation |
|---|---|---|---|---|---|---|
| 50,000 | one celery worker | 29.54s | 13.00s | 3.00s | 124.00s | 34.55 |
| 50,000 | dedicated celery worker | 6.38s | 3.57s | 2.72 | 49.72 | 8.93 |
| 50,000 | dedicated celery worker & container | 3.23s | 2.19s | 0.148 | 17.38 | 2.64 |

## Additional improvements from the issue thread

The issue comment listed 4 follow-up improvements as "worth splitting out." **1 of 4 is implemented in this PR:**

- ✅ **Skip no-op updates** — Unchanged duplicate rows are currently re-UPDATEd, generating ~22-33 MB WAL per chunk and risking bloat if autovacuum lags. Add WHERE ... IS DISTINCT FROM ... to the merge so identical rows aren't rewritten.
- ✅ Dedicated Celery queue for XML imports — with bounded concurrency - today they share the 10-thread worker with other queues, and two concurrent imports compound CPU/memory.
- ⬜ emit separate parse / DB-write / commit / webhook-prep timings so the split is visible in prod (not just local benchmarks).
- ⬜ building svix payloads for a 50k chunk produced ~44 payloads / ~19 MB and slowed next-chunk parsing ~25% in @MichalZak17's test; revisit whether large-import webhooks should be batched/skipped.

## Bug Fixes
* Improved bulk data-point imports for large batches.
* Duplicate records that do not change existing values are now handled correctly.
* Genuine updates are accurately reflected in import results.
* NULL and decimal values are preserved during bulk imports.
* Import results now consistently account for duplicate records, including records requiring no changes.

## Reliability
* Scheduled background tasks now run on a dedicated processing queue, improving workload isolation and execution reliability.

### DIFF
--- a/backend/app/repositories/data_point_series_repository.py
+++ b/backend/app/repositories/data_point_series_repository.py
@@ -1,10 +1,13 @@
 import contextlib
 from datetime import datetime, time, timedelta
+from decimal import Decimal
+from typing import LiteralString, NamedTuple
+from typing import cast as typing_cast
 from uuid import UUID
 
+from psycopg import Connection as PGConnection
 from psycopg.errors import UniqueViolation
 from sqlalchemy import ColumnElement, Date, Interval, String, and_, asc, case, cast, func, literal_column, text, tuple_
-from sqlalchemy.dialects.postgresql import insert
 from sqlalchemy.exc import IntegrityError as SQLAIntegrityError
 
 from app.database import DbSession
@@ -60,12 +63,6 @@ class DataPointSeriesRepository(
     CrudRepository[DataPointSeries, TimeSeriesSampleCreate, TimeSeriesSampleUpdate],
 ):
     """Repository for unified device data point series."""
-
-    # PostgreSQL/psycopg cap of 65535 bind params per query. Derive the row chunk
-    # from the column count in _insert_data_points so adding a column can't silently
-    # push a full chunk over the limit (8 cols -> 8191 rows).
-    _INSERT_COLUMNS_PER_ROW = 8
-    BATCH_INSERT_CHUNK_SIZE = 65_535 // _INSERT_COLUMNS_PER_ROW
 
     def __init__(self, model: type[DataPointSeries]):
         super().__init__(model)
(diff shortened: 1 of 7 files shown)


KaliszS (changes requested):
I've added a commit with celery refactor.

I think it's better to avoid creating separate containers for XML sync, because it's not something most users would use anyway. My approach is based on running two celery wrokers in one container, it's still going to use more RAM, but still not as much as two docker containers.

What's changed:
- added `&` at the end of the first worker, because it's daemon, so we need to start that in background in order to run another command after that
- added `echo`s command to describe workers destintions
- reduce concurrency from 3 to 2 (you may increase that number, but we know some clients have just 2 cores on their infra, so switching between extra cores might be costly)
- changed pool for CPU bound queues from threads to `prefork`, becasue GIL won't let us here to make use of threads anyway, while prefork will spawn more proccesses

backend/app/repositories/data_point_series_repository.py
```diff
@@ -149,78 +146,122 @@ def _resolve_data_sources(
...
+
+        # Raw psycopg connection sharing this Session's transaction - COPY has no
+        # SQLAlchemy Core equivalent, and using a separate connection would commit
+        # outside this transaction.
+        raw_conn: PGConnection | None = db_session.connection().connection.driver_connection
+        assert raw_conn is not None, "no DBAPI connection on an active Session"
+        with raw_conn.cursor() as cursor:
+            cursor.execute(
+                """
+                CREATE TEMP TABLE IF NOT EXISTS data_point_series_staging (
```
KaliszS: Now temp table is hardcoded and can desync with sqlalchemy model, we should handle this somehow.

backend/app/repositories/data_point_series_repository.py
```diff
@@ -149,78 +146,122 @@ def _resolve_data_sources(
...
+                    id uuid,
+                    external_id varchar(100),
+                    data_source_id uuid,
+                    recorded_at timestamptz,
+                    zone_offset varchar(10),
+                    value numeric(10, 3),
+                    series_type_definition_id integer,
+                    is_daily_total boolean
+                ) ON COMMIT DROP
```
KaliszS: Maybe `ON COMMIT DELETE ROWS`?

After tests with 2000 chunks, `pg_attribute` grew from 0.9 to 6.1MB. The reason is probably the `autovacuum` process. Dropping tables sets them as dead tables, not removing them (it's autovacuum responsibility). And it might just be too slow, because we create multiple tables for group of chunks using drop.

Of course it will leave the staging table empty, but it's ok in my opinion.

KaliszS: This approach should be more efficient for small chunks and about as quick as drop approach for larger ones.

backend/app/repositories/data_point_series_repository.py
```diff
@@ -149,78 +146,122 @@ def _resolve_data_sources(
...
+            )
+            cursor.execute("TRUNCATE data_point_series_staging")
+
+            with cursor.copy(f"COPY data_point_series_staging ({self._COPY_COLUMNS_SQL}) FROM STDIN") as copy:
+                for row in rows:
+                    copy.write_row(row)
+
+            cursor.execute(
+                f"""
+                    INSERT INTO data_point_series ({self._COPY_COLUMNS_SQL})
```
KaliszS: Maybe we should add `ORDER BY` statement?

`ORDER BY data_source_id, series_type_definition_id, recorded_at`

backend/app/repositories/data_point_series_repository.py
```diff
@@ -149,78 +146,122 @@ def _resolve_data_sources(
...
+                        zone_offset = excluded.zone_offset,
+                        is_daily_total = excluded.is_daily_total
+                    WHERE data_point_series.value IS DISTINCT FROM excluded.value
+                       OR data_point_series.external_id IS DISTINCT FROM excluded.external_id
+                       OR data_point_series.zone_offset IS DISTINCT FROM excluded.zone_offset
+                       OR data_point_series.is_daily_total IS DISTINCT FROM excluded.is_daily_total
+                    RETURNING (xmax = 0)
+                """
+            )
+            result_rows = cursor.fetchall()
```
KaliszS: That `fetchall` pulls one tuple per each row, just to count them. I suggest to wrap the merge in a CTE to keep `RETURNING` server-side and returns just a single number. Something like that:

```py
WITH merged AS (
      INSERT INTO data_point_series (...) SELECT ... FROM data_point_series_staging
      ON CONFLICT (...) DO UPDATE SET ... WHERE ...
      RETURNING (xmax = 0) AS was_insert
  )
  SELECT count(*) FILTER (WHERE was_insert) FROM merged

  inserted = cursor.fetchone()[0]
  return WriteCounts(inserted, len(rows) - inserted)
```

KaliszS (changes requested):
Should be also noted we are migrating from `Bind` protocol to `CopyData` protocol.

PR #1510, opened 2026-08-31
feat(backend): add name, intensity and recording source to the workouts API

### LINKED ISSUE #1494: Oura: per-workout `source` (manual vs auto-detected) is parsed then discarded in _normalize_workout
### Summary

`OuraWorkoutJSON` parses Oura's per-workout `source` field, but `_normalize_workout` overwrites it with the provider name, so the value never reaches consumers. The same happens to `intensity` and `label`.

### Where

[`backend/app/services/providers/oura/workouts.py#L191`](https://github.com/the-momentum/open-wearables/blob/main/backend/app/services/providers/oura/workouts.py#L191)

```python
workout_create = EventRecordCreate(
    ...
    source=self.provider_name,   # <- always "oura"
    user_id=user_id,
)
```

`raw_workout.source` appears **zero** times in that file, despite being declared in the schema:

[`backend/app/schemas/providers/oura/imports.py#L30`](https://github.com/the-momentum/open-wearables/blob/main/backend/app/schemas/providers/oura/imports.py#L30)

```python
class OuraWorkoutJSON(BaseModel):
    ...
    intensity: str | None = None  # easy / moderate / hard
    label: str | None = None
    source: str | None = None
```

`intensity` and `label` are likewise parsed and never read.

### Why it matters

Oura's workout `source` distinguishes a workout the ring **auto-detected** from one the user **typed in by hand**. Those are very different data for anything competitive or scoring-based — a self-reported workout shouldn't carry the same weight as a measured one.

Because `source` is collapsed to the provider name, a consumer has no way to tell them apart. On our side every Oura workout arrives as:

```json
"source": { "device": null, "provider": "oura" }
```

across 202 records, with no per-workout provenance available anywhere in the payload.

`intensity` (easy / moderate / hard) is a useful signal for the same reason.

### Suggestion

Carry Oura's per-workout `source` through to the normalized record rather than discarding it — either as a distinct field, or on the detail record alongside the metrics. `EventRecordCreate.source` is presumably meant to be the provider, so a separate field is probably the cleaner fix.

Happy to open a PR if you'd point me at where you'd prefer it to live.

### Separate question: heart rate on Oura workouts

`_build_metrics` maps `energy_burned`, `distance` and `moving_time_seconds` only, and `OuraWorkoutJSON` has no heart-rate field — which reads like Oura's `/v2/usercollection/workout` simply doesn't return HR, rather than anything being dropped here.

Is that right? We receive `avg_heart_rate_bpm` / `max_heart_rate_bpm` as keys with `null` in all 202 records, and I'd like to know whether that's an Oura limitation or something requestable, before I go looking in the wrong place.

### Environment

- Provider: Oura
- Path: workout ingestion via webhook
- Observed across 202 workout records

McArthur955:
Closes #1494

## Description

Three providers (Oura, Strava, Garmin) each parse a per-workout provenance & name field from their API but silently drop it during normalization

- **Oura** `source` (manual/autodetected/confirmed/workout_heart_rate) and `intensity` (easy/moderate/hard)
- **Strava** `manual` (bool) and `name`
- **Garmin** `manual`/`isWebUpload` (bool) and `activityName`

None of this reached storage or the `/events/workouts` API response, so consumers couldn't tell a measured workout from a manually-entered one, or see the workout's name.

This PR adds `entry_source`, `intensity`, and `label` to `WorkoutDetails` (new migration), wires all three providers to populate them, and surfaces them on `Workout`/`WorkoutDetailed`.

## Summary by CodeRabbit

- **New Features**
  - Workout records now include optional entry source, intensity, and label details.
  - Entry source and intensity use consistent, provider-independent values.
  - Workout responses display labels and normalized metadata.
  - Added metadata support for Apple, Garmin, Oura, and Strava, including Oura intensity levels.

- **Bug Fixes**
  - Overlong workout labels are truncated to 255 characters instead of rejected.

- **Documentation**
  - Updated integration examples and coverage documentation for the new fields and standardized values.

- **Validation**
  - Unsupported provider values are rejected, while recognized “unknown” values remain valid.


bartmichalak (changes requested):
I think `entry_source` needs a unified vocabulary before this goes out. Right now Strava and Garmin emit manual / auto, Oura passes through its raw values (manual / autodetected / confirmed / workout_heart_rate), and the field is a plain str - so what you get depends on which provider the workout came from.

Let's consider this: as a developer I want to drop manually-entered workouts from a user's history. What do I filter on? `entry_source == "manual"` works today only because Oura happens to use the same word - I'd have to read Oura's API docs to be sure. The inverse, entry_source == "auto", silently loses every Oura workout.

The point of Open Wearables is that you integrate once against a normalized model and never have to look at a provider's API. This field leaks provider semantics straight through. 

I'd make it a closed enum (manual / automatic / unknown, matching the existing RecordingMethod in the mobile SDK) with explicit per-provider mapping, and do the same for intensity. 

If you think unification is genuinely hard here, then at the very least we should define and document the vocabulary per provider - but a developer integrating with our API has to know what to expect from this field.

bartmichalak:
One more question while we're at it: both Oura and Whoop can import workouts from Apple Health
- https://support.ouraring.com/hc/pl/articles/360025438734-Integracja-z-Apple-Health
- https://support.whoop.com/s/article/Apple-Health-Integration?language=en_US

Do their APIs return those imported workouts at all, and if so, are they distinguishable from workouts recorded by the device itself? If they are, that's another case entry_source should cover (or we risk double-counting against our own HealthKit import)

backend/app/services/providers/garmin/workouts.py
```diff
@@ -195,6 +195,16 @@ def _build_metrics(self, raw_workout: GarminActivityJSON) -> EventRecordMetrics:
             "average_cadence": average_cadence,
         }
 
+        # isWebUpload means the activity was uploaded through Garmin Connect (e.g. a
+        # GPX/TCX file) rather than synced live from a device - treat it as manual too.
+        if raw_workout.manual is not None or raw_workout.isWebUpload is not None:
+            metrics["entry_source"] = "manual" if (raw_workout.manual or raw_workout.isWebUpload) else "auto"
```
bartmichalak: I don't think `isWebUpload` should map to manual. A FIT/GPX file uploaded to Garmin Connect (e.g. from a Wahoo or Coros device) is still measured data - it's just not synced live. 

Treating it as manual mislabels real recordings as self-reported, which is exactly the distinction consumers of this field care about. If we want to keep the signal, I'd add a separate value like imported for this case, and reserve manual for manual=true only.

bartmichalak:
I renamed the PR title to say what this gives API consumers, not what the new columns are added

backend/app/services/providers/garmin/workouts.py
McArthur955: * I have added enums to the json models for `entry_source ` & `intensity ` fields.
* You were right about the `isWebUpload` is now saved as automatic.
* The articles which you have mentioned are clear that both providers import workout data from `Apple Health`. (I'll show you screenshots below), but I didn't find any specific flag which can inform us that the date is coming from the `Apple Health`. I suggest creating internal issue in our linear to take care of duplicates, because they seem possible. Maybe we can match them by start and finish datetime?

image

image

KaliszS (approved):

PR #1515, opened 2026-09-01
security: ignore host-only compose override

Crich1187:
ADR-003 root-Secret-FleetCutover-001.22. Reviewed integration SHA: 7d9644a7b76421cddd5238d8b449ebc60a08afea. Gate 3 PASS artifact is recorded in the Bead. This is a one-path .gitignore guard; no credential values, rotations, or service changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,3 +174,5 @@ CLAUDE.local.md
 
 # Exception: AI tool configs that should be tracked
 !.gemini/settings.json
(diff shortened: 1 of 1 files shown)


KaliszS:
Hi. We have decided to reject this PR, as we may wish to add our own extensions to compose in the future, or our project’s forks may extend their configuration in this way without losing the main compose file. If the override is hidden now, this could cause problems when we make it public.

However, we recommend that you implement local ignoring using an exclude file (`.git/info/exclude`), which works in the same way as `.gitignore` but is never committed.

PR #1540, opened 2026-09-03
fix(backend): unify sleep data delivered by api & webhook

### LINKED ISSUE #1529: Sleep payload differs between API and webhook
The sleep object returned by the API carries more fields than the one delivered by the `sleep.created` webhook.

API (`GET /sleep-sessions`):

```json
{
  "id": "3c90c3cc-0d44-4b50-8888-8dd25736052a",
  "start_time": "2023-11-07T05:31:56Z",
  "end_time": "2023-11-07T05:31:56Z",
  "source": {
    "provider": "apple",
    "device_name": "iPhone 15",
    "source": "Connect",
    "device": "iPhone15,2",
    "device_type": "phone"
  },
  "duration_seconds": 123,
  "zone_offset": "+01:00",
  "sleep_duration_seconds": 123,
  "efficiency_percent": 123,
  "stages": { "awake_minutes": 123, "light_minutes": 123, "deep_minutes": 123, "rem_minutes": 123 },
  "sleep_stage_intervals": [
    { "stage": "in_bed", "start_time": "2023-11-07T05:31:56Z", "end_time": "2023-11-07T05:31:56Z" }
  ],
  "is_nap": false
}
```

Webhook (`sleep.created`):

```json
{
  "type": "sleep.created",
  "data": {
    "id": "b2c3d4e5-f6a7-8901-bcde-f12345678901",
    "user_id": "550e8400-e29b-41d4-a716-446655440000",
    "start_time": "2025-12-19T22:15:00+01:00",
    "end_time": "2025-12-20T06:45:00+01:00",
    "zone_offset": "+01:00",
    "duration_seconds": 30600.0,
    "source": { "provider": "oura", "device": "Oura Ring Gen3" },
    "efficiency_percent": 87.0,
    "stages": { "deep_minutes": 95, "rem_minutes": 80, "light_minutes": 215, "awake_minutes": 12 },
    "is_nap": false
  }
}
```

Missing from the webhook: 
-`sleep_duration_seconds`
- `sleep_stage_intervals`,
- `source.source` / `source.device_type` / `source.device_name`. 

Consumers that want a full sleep session still have to call the API after every webhook, which defeats the webhook-first flow. The same reduced `source` object is used by `workout.created` and `menstrual_cycle.created`.

**The cause is that the API response is built from the `SleepSession` schema (`backend/app/schemas/responses/activity/events.py`) while the webhook payload is a hand-written dict in `on_sleep_created` (`backend/app/services/outgoing_webhooks/events.py`), so the two shapes drift independently.**

Proposal: extract a shared payload builder used by both the API mapper and the webhook emitter so the shapes cannot diverge again. Additive change, backwards compatible, no Svix event-type re-registration required. Docs sample in https://openwearables.io/docs/api-reference/guides/webhooks needs the same update.

One thing to watch: including `sleep_stage_intervals` grows the payload by roughly 5-10 KB per night (40-100 intervals), well under the Svix 1 MB limit.

McArthur955:
## Description

So the GET `/sleep-sessions` response included `sleep_duration_seconds`, `sleep_stage_intervals`, and additional source metadata (`source`, `device_type`, `device_name`) that the `sleep.created` webhook never sent, forcing consumers to make an extra API call after every webhook :/

Now `on_sleep_created` (events.py) accepts and emits these fields. Both call sites (`create_or_merge_sleep`, `_emit_event_record_webhook` in `event_record_service.py`) were updated to pass them through - the data was already loaded in memory (EventRecordDetailCreate.sleep_stages/sleep_total_duration_minutes, DataSource), so no new queries were needed except one DataSource lookup (only when Svix is enabled) for the merge path that didn't already have it.

---

This is a hot fix of the [1529](https://github.com/the-momentum/open-wearables/issues/1529) issue.
But we should try to unify webhooks and api responses with the same schema / response building functions. But because the issue was a bug, I decided to move forward with the simplest solution asap, and make a new task on our linear to not forget to continue the work on making sure the data is unified.

## Summary by CodeRabbit

* **New Features**
  * Sleep-created webhook events now include sleep duration, sleep-stage intervals, and richer source and device details.
  * Source information can include the originating app, device type, and device name.
  * Device metadata is automatically inferred when it is not explicitly provided, including from recognized source information.

* **Documentation**
  * Updated the webhook reference with the expanded payload format and clarified the difference between time asleep and total time in bed.
  * Documented when source and sleep-stage details may be unavailable or abbreviated.

### DIFF
--- a/backend/app/constants/webhooks/test_payloads.py
+++ b/backend/app/constants/webhooks/test_payloads.py
@@ -14,7 +14,13 @@ _RECORD_ID = "00000000-0000-0000-0000-000000000001"
 _CONNECTION_ID = "00000000-0000-0000-0000-000000000003"
 
 _SOURCE_GARMIN = {"provider": "garmin", "device": "Garmin Fenix 7"}
-_SOURCE_OURA = {"provider": "oura", "device": "Oura Ring Gen3"}
+_SOURCE_OURA = {
+    "provider": "oura",
+    "source": "oura",
+    "device": "Oura Ring Gen3",
+    "device_type": "ring",
+    "device_name": "Oura Ring Gen3",
+}
 _SOURCE_APPLE = {"provider": "apple", "device": "Apple Watch Series 9"}
 
 
--- a/backend/app/repositories/event_record_repository.py
+++ b/backend/app/repositories/event_record_repository.py
@@ -66,6 +66,7 @@ class EventRecordRepository(
                 device_model=creator.device_model,
                 source=creator.source,
                 software_version=creator.software_version,
+                original_source_name=creator.source,
             )
             data_source_id = data_source.id
 
--- a/backend/app/services/event_record_service.py
+++ b/backend/app/services/event_record_service.py
@@ -254,6 +254,10 @@ class EventRecordService(
                     final_detail.sleep_rem_minutes,
                 ]
             )
+            device_type = None
+            if svix_service.is_enabled() and record.data_source_id is not None:
+                source_data_source = self.data_source_repo.get(db_session, record.data_source_id)
+                device_type = source_data_source.device_type if source_data_source else None
             on_sleep_created(
                 record_id=result.id,
                 user_id=user_id,
@@ -273,6 +277,18 @@ class EventRecordService(
                 if has_stages
                 else None,
                 is_nap=final_detail.is_nap,
+                source_app=record.source,
+                device_type=device_type,
+                sleep_duration_seconds=(
+                    final_detail.sleep_total_duration_minutes * 60
+                    if final_detail.sleep_total_duration_minutes is not None
+                    else None
+                ),
+                sleep_stage_intervals=(
+                    [stage.model_dump(mode="json") for stage in final_detail.sleep_stages]
+                    if final_detail.sleep_stages
+                    else None
+                ),
             )
         return result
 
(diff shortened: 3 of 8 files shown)


backend/app/constants/webhooks/test_payloads.py
```diff
@@ -14,7 +14,13 @@
 _CONNECTION_ID = "00000000-0000-0000-0000-000000000003"
 
 _SOURCE_GARMIN = {"provider": "garmin", "device": "Garmin Fenix 7"}
-_SOURCE_OURA = {"provider": "oura", "device": "Oura Ring Gen3"}
+_SOURCE_OURA = {
```
czajkub, line 17: From what I recall, we currently do not collect info on the device model (or even type for that matter, even though it's always gonna be a ring) for Oura.

It is available via the [API](https://cloud.ouraring.com/v2/docs#operation/Single_ring_configuration_Document_v2_usercollection_ring_configuration__document_id__get), but not currently consumed, due to the nature of having to extract the data source from a separate endpoint and keep track of possible changes to it, but it is definitely an issue worth tracking

czajkub: although it is still only a test payload - so not that severe of an issue :P

KaliszS (approved):
Tested on real data.

Issue #1539, opened 2026-09-03
Sleep stage intervals accumulate one copy per re-sync


navFooh:
## Description

Re-syncing a night appends a full copy of its stage timeline to `sleep_details.sleep_stages` instead of replacing it. After many syncs of a still-visible night, `GET /api/v1/users/{id}/events/sleep` returns the hypnogram repeated N times: one of our Polar sessions serves 5829 intervals of which 67 are distinct, each appearing exactly 87 times. The `stages` minute summary stays correct, so the bug is invisible there and only shows in `sleep_stage_intervals`.

## Steps to Reproduce

1. Connect a Polar account and let a night sync
2. Trigger sync for the same range a few more times
3. `GET /api/v1/users/{user_id}/events/sleep?start_date=...&end_date=...`
4. Group each session's `sleep_stage_intervals` by `(stage, start_time, end_time)`: every interval appears once per sync

## Expected Behavior

Each interval appears once; re-syncing an unchanged night leaves the timeline as is.

## Actual Behavior

The interval list grows by one full copy per re-sync while the night is inside the sync lookback window. Copies are byte-identical, and the copy count is uniform within a session (67 distinct x 87 for the worst night here, which matches roughly two days of interval syncs at the default cadence). Deduplicated, the set matches the session's own `stages` summary exactly.

## Environment

- Backend: 0.7.0 (`themomentum/open-wearables-backend:sha-c79410a`), Docker Compose
- Provider: Polar, REST pull. Health Connect sessions on the same instance stay at x1.

## Additional Context

The mechanism looks like `EventRecordService._create_or_merge_sleep_inner`: the merge path builds `merged_stages = sorted(adj_stages + new_stages, ...)` and the `same_window` branch then updates the detail in place with that concatenated list. So any re-sync that resolves through the adjacent merge with an identical window appends one copy. The minute fields survive because the overlap recompute walks coverage rather than summing rows. The code path is unchanged on current main.

Side note: intervals can carry an `unknown` stage that the `stages` summary omits.

KaliszS:
Hi. Do you have `pull_sync_lookback` variable set?

navFooh:
Yes, `PULL_SYNC_LOOKBACK=2d`. We verified against our sync logs that each session's copy count is exactly 1 + the number of syncs that ran while the night was inside the lookback, and it stops growing once the night leaves the window.

PR #1547, opened 2026-09-04
fix(backend): replace stages on same-window re-sync instead of concatenating (fixes #1539)

ZacLou:
"## Problem\n\nRe-syncing a Polar night whose window is identical to an existing sleep record currently appends the incoming stages to the stored stages via the adjacent-merge branch, duplicating every interval once per sync (#1539).\n\n## Root cause\n\ncreate_or_merge_sleep concatenates sleep stages for all adjacent merges. When same_window is true (identical start/end/data_source), the merged detail is written back with the concatenated list, so each re-sync adds another copy of the same stages.\n\n## Fix\n\nIn the same_window branch, replace the stored sleep detail with the incoming detail instead of merging. Minute-level aggregates and sleep stages now come from the latest incoming detail.\n\nThe conflict-handling path that catches the unique-constraint violation also uses the same replacement logic.\n\n## Tests\n\n- Added \test_same_window_re_sync_replaces_sleep_stages to verify that re-syncing the exact same window leaves exactly one copy of the incoming stages.\n- The existing \test_merge_concatenates_sleep_stages still validates that non-overlapping adjacent merges continue to concatenate correctly.\n\n## Notes\n\nI don't have a real Polar connection available, so end-to-end hardware validation hasn't been done. The targeted regression test plus the existing backend suite pass locally."

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed sleep record synchronization when the data source is resolved during insertion.
  - Sleep stage details and deep-sleep duration are now correctly replaced during same-window updates.
  - Prevented failures when re-syncing records with missing data source information.

- **Tests**
  - Added coverage for same-window synchronization with unresolved data sources.

### DIFF
--- a/backend/app/services/event_record_service.py
+++ b/backend/app/services/event_record_service.py
@@ -455,6 +455,13 @@ class EventRecordService(
             )
 
             if same_window:
+                # Re-sync of the exact same window: replace sleep stages with the
+                # incoming detail's stages rather than concatenating them, which
+                # would duplicate every interval once per re-sync.
+                merged_detail_fields = {
+                    **merged_detail_fields,
+                    "sleep_stages": list(detail.sleep_stages or []),
+                }
                 self.event_record_detail_repo.delete_by_record_id(db_session, adjacent.id, "sleep")
                 self.event_record_detail_repo.create_and_flush(
                     db_session,
--- a/backend/app/services/event_record_service.py
+++ b/backend/app/services/event_record_service.py
@@ -455,17 +455,24 @@ class EventRecordService(
             )
 
             if same_window:
-                # Re-sync of the exact same window: replace sleep stages with the
-                # incoming detail's stages rather than concatenating them, which
-                # would duplicate every interval once per re-sync.
-                merged_detail_fields = {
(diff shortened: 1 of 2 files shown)


ZacLou:
@coderabbitai's latest review has no actionable comments and merge risk is now minimal. Could a maintainer please approve the workflow/run and review? The only remaining warning is docstring coverage, which is pre-existing in the touched functions and not introduced by this change.

bartmichalak (commented):
@ZacLou in the meantime, could you please correct the PR description? :)

Also, were you able to test these changes on a real Polar connection?

ZacLou:
I don't have a real Polar connection available, so this is validated against the targeted regression test and existing backend suite only. I'd appreciate a hardware owner confirming end-to-end after merge.

KaliszS:
Hi. You haven’t fixed the PR description or the pipeline, and it turns out you don’t even have the device and are passing on the task of checking whether the solution works for us at all, so I’m closing the PR.

The issue is already being tracked here, we or another contributor will sort it out: https://github.com/the-momentum/open-wearables/issues/1539

PR #1562, opened 2026-09-07
fix(sync): steal a shared pull lock whose holder died mid-run

ruloqs:
## Description

A periodic pull that is killed together with its Celery worker — `docker compose up --force-recreate`, SIGKILL after
the container stop grace period, OOM, host reboot — never reaches `release_primary`. Its
`linked_sync:{provider}:{provider_user_id}:pull:primary` key then survives until the 4 h TTL and every following
`sync_vendor_data` run for that account logs `Skipping … pull — another linked profile is syncing`. We hit this with a
single Google Health connection: no new data for hours until the key was deleted by hand.

`sync_vendor_data` already steals the lock when the holder has no active connection. This adds the second stale case:

- `sync_coordination.primary_is_idle(provider, holder_user_id)` reads the holder's sync-status stream (Redis, 24 h
  TTL — the only trace a running task leaves) and reports the lock as orphaned when the holder has no `in_progress`
  run for the provider updated within `PRIMARY_IDLE_SECONDS` (30 min), or no run recorded at all.
- The pull task steals in that case through the existing path (`release_stale_primary` + a second election) and logs
  the reason.
- Garmin backfills keep their own progress keys and GC (`garmin/gc_task.py`) and are not touched.

30 min is well above the longest gap between progress events of a legitimate pull (a 2-day-lookback Google pull
reports every few minutes) and short enough that a dead worker costs at most one polling cycle.

## Checklist

- [x] My code follows the project's code style (`ruff`, `ty` clean)
- [x] I have added tests (`tests/services/test_sync_coordination.py` against the real sync-status service;
      `tests/tasks/test_sync_vendor_data_idle_lock.py` for stealing vs. skipping in the task)
- [x] I have updated relevant documentation in `docs/` (`docs/architecture/multi-account-sync.mdx`)
- [x] `uv run pre-commit run --all-files` passes (backend hooks; this PR does not touch `frontend/`)

## Testing Instructions

**Steps to test:**
1. `cd backend && SECRET_KEY=test ENV=test uv run pytest tests/services/test_sync_coordination.py tests/tasks/test_sync_vendor_data_idle_lock.py tests/tasks/test_sync_vendor_data_task.py -q`
2. Manual: start a pull, `docker kill` the worker mid-run, restart it and wait one polling cycle.

**Expected behavior:** the next cycle logs `Stealing stale google primary lock — holder has no live sync run (worker
died mid-run?)` and syncs normally, instead of skipping until the 4 h TTL.

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronization can now recover from orphaned or inactive provider locks, preventing stalled sync operations.
  * Active lock holders continue to retain their locks, avoiding unnecessary duplicate synchronization.
  * Logs now identify why a lock was considered stale.

* **Documentation**
  * Updated multi-account synchronization documentation to explain stale-lock recovery behavior.

### DIFF
--- a/backend/app/integrations/celery/tasks/sync_vendor_data_task.py
+++ b/backend/app/integrations/celery/tasks/sync_vendor_data_task.py
@@ -14,7 +14,7 @@ from app.schemas.auth import LiveSyncMode
 from app.schemas.responses.upload import ProviderSyncResult, SyncVendorDataResult
 from app.schemas.sync_status import SyncSource, SyncStage, SyncStatus
 from app.services.providers.factory import ProviderFactory
-from app.services.sync_coordination import release_primary, release_stale_primary, try_become_primary
+from app.services.sync_coordination import primary_is_idle, release_primary, release_stale_primary, try_become_primary
 from app.services.sync_status_service import completed, failed, new_run_id, progress, started
 from app.utils.config_utils import format_duration
 from app.utils.context import trace_id_var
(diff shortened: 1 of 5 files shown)


czajkub:
resolved in #1571

PR #1575, opened 2026-09-09
fix(integrations): a manual Strava activity has no streams, not a failure

slavarosin:
Hi!

Strava answers `GET /activities/{id}/streams` with 404 for manual activities — they have no streams by definition. `_ingest_workout_streams` reports that through `log_and_capture_error`, so one hand-logged gym session produces a traceback and a Sentry event on every sync pass that re-covers it.

In our deployment that was twice an hour, indefinitely: the trailing lookback keeps re-covering the activity, and it never gains samples, so nothing ever settles. The workout itself is stored correctly; only the reporting is wrong.

This treats 404 from the streams endpoint as "this activity has no streams" and skips quietly, with an info line carrying the activity id. Every other status keeps its current reporting, including the 500 path.

Real example that prompted it: a manual "Workout" entry with `manual: true`, no device, zero distance.

Verification: `uv run pytest tests/providers/strava/` — 16 passed, including two new cases (404 not reported, other errors still reported); `ruff check` clean.

Unrelated to #1298 and #1447, so this can land independently of either.

https://claude.ai/code/session_01EY3bZLVL41Q5SHG4uW39Xo

## Summary by CodeRabbit

- **Bug Fixes**
  - Manual Strava activities without workout stream data are now handled gracefully without being reported as errors.
  - Other errors returned by Strava’s stream endpoint continue to be recorded for investigation.

- **Tests**
  - Added coverage for manual activities without streams and for handling other stream-related errors.

### DIFF
--- a/backend/app/services/providers/strava/workouts.py
+++ b/backend/app/services/providers/strava/workouts.py
@@ -3,6 +3,8 @@ from decimal import Decimal
 from typing import Any
 from uuid import UUID, uuid4
 
+from fastapi import HTTPException, status
+
 from app.config import settings
 from app.constants.workout_types import get_unified_strava_workout_type
 from app.database import DbSession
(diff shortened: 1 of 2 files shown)


KaliszS:
It seems that this code won’t suppress errors on the sentry, as the exception will still be logged by `api_client.py`, and we don’t want to suppress that, as it would affect the other providers.

This is the path that should be triggered if this error occurs:
```
strava/workouts.py:260   _build_workout_samples → self._make_api_request("/api/v3/activities/{id}/streams")
        ↓
templates/base_workouts.py:144   _make_api_request
        ↓
templates/base_workouts.py:155   return make_authenticated_request(...)
        ↓
api_client.py:97    make_authenticated_request
        ↓
api_client.py:195   response.raise_for_status()      ← 404 → httpx.HTTPStatusError
        ↓
api_client.py:245   log_structured(logger, "error", "API error", ...)   # this is the problem
        ↓
api_client.py:258   raise HTTPException(status_code=404, ...)
        ↓
strava/workouts.py   except HTTPException → PR operates here
```

KaliszS:
Thanks for the PR, but I'm going to close it.

The Sentry premise no longer holds: `log_and_capture_error` in `app/utils/sentry_helpers.py` already skips Sentry for every 4xx `HTTPException` (added in #1203 as a side change), so a 404 from `/streams` is logged as a warning and never reaches Sentry. In practice this change swaps that warning for an info line, while the error-level "API error" line from `api_client.py` stays, because it is emitted before the exception reaches the Strava code.

Issue #1577, opened 2026-09-09
energy timeseries and /summaries/activity calorie totals wildly inflated for Google provider (raw + reconcile default)


rchodagam-tech:
## Summary

For the Google provider, the `energy` timeseries (default `resolution=raw`, default `GOOGLE_USE_RECONCILE=true`) returns values that are not usable as a summable per-sample series — summing a day's raw `energy` points against a single real account produced **229,313 kcal** for one calendar day. This isn't a client-side aggregation mistake: `GET /api/v1/users/{id}/summaries/activity` — the pre-aggregated, "trust this" daily rollup — independently returns `active_calories_kcal: 251435.61` / `total_calories_kcal: 251435.61` for the same day, so the corruption appears to be upstream of both endpoints, in ingestion/storage rather than in either read path.

## Reproduction

Google Health provider, connected via the standard OAuth flow, default granularity settings (`raw` + `GOOGLE_USE_RECONCILE=true`).

**1. Raw timeseries, default resolution:**
```
GET /api/v1/users/{user_id}/timeseries?start_time=2026-09-08&end_time=2026-09-09&types=energy&resolution=raw
```
Sample of the response (real values from one account, redacted to a representative slice):
```json
{"timestamp": "2026-09-08T08:00:04Z", "value": 70.728, "source": {"provider": "google", "source": "google_health_api"}, "is_daily_total": false}
{"timestamp": "2026-09-08T08:01:30Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:01:33Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:02:14Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:02:18Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:03:00Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:04:10Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:04:15Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:04:33Z", "value": 70.728, ...}
{"timestamp": "2026-09-08T08:06:16Z", "value": 72.138, ...}
```
[...]

PR #1625, opened 2026-09-14
fix: Google Health API active/basal energy ingestion

czajkub:
As stated in issue #1577, total energy is currently stored as `energy` (which is active here, going to be changed) and uses broken rollup mechanism, which will be disabled in another PR.

What this PR does is:
- prevent the total energy endpoint from saving to the active energy data point
- use the correct active energy endpoint (using `list` operation) to save it, giving it a fake external id to identify it from wrong previous values
- migration (not added to startup) to remove old values

Also, adding basal energy based on total and active values.
To accomplish this, we use the `dailyRollup` endpoint (new functionality), what happens is we take the total and active energy burned (both from daily rollup), and we save the difference (which is basal) as basal daily total, as shown in this screenshot:

image

Also data migration script working:

```bash
docker compose exec app uv run python scripts/data_migrations/purge_google_total_calories_energy.py --dry-run
[dry-run] would delete 3769 untagged Google Health API energy row(s).


docker compose exec app uv run python scripts/data_migrations/purge_google_total_calories_energy.py
Deleted 3769 untagged Google Health API energy row(s) written from total-calories.


docker compose exec app uv run python scripts/data_migrations/purge_google_total_calories_energy.py
No untagged Google Health API energy rows; nothing to purge.
```

How daily rollup works is basically:
- we pass a range of dates anchored on local midnight (we can also ask for concrete time, e.g. from 6:00 to 6:00, but we don't do this to prevent rollup window issue reappearing)
- the endpoint returns a daily total, and we also store it as daily total in data point series table (though google doesnt return the offset, but it won't hurt since we only need the day)

## Summary by CodeRabbit

* **New Features**
  * Added support for daily rollup data from Google Health.
  * Added derived daily metrics, including basal energy calculated from total calories and active energy burned.
  * Improved handling of civil-day totals and paginated Health data.

* **Bug Fixes**
  * Updated energy tracking to use active-energy intervals and derived basal energy instead of total-calorie rollups.
  * Correctly combines same-day values and prevents negative combined energy values from being recorded.


backend/app/schemas/providers/google/health_api.py
```diff
@@ -79,6 +80,39 @@ class ListSpec:
     extra: tuple[SeriesField, ...] | None = None
 
 
+@dataclass(frozen=True)
+class DailyRollupSpec:
```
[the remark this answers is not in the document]
czajkub: in this file already declared classes are dataclasses and not pydantic models, it should be left, if one is to be changed then all should

backend/app/services/providers/google/health_api/data_247.py
```diff
@@ -92,6 +105,19 @@ def load_and_save_all(
             if counts is not None:
                 results[metric.data_type] = counts
 
+        for derived in DERIVED_DAILY_METRICS:
```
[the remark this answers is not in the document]
czajkub: sync data type has only one caller and it only receives data that google notifies us about and energy has none of that, i dont understand the issue

backend/app/services/providers/google/health_api/data_247.py
```diff
@@ -92,6 +105,19 @@ def load_and_save_all(
...
+            try:
+                with db.begin_nested():
```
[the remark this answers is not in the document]
czajkub: this pattern already exists in codebase, skipping

backend/scripts/data_migrations/purge_google_total_calories_energy.py
```diff
@@ -0,0 +1,83 @@
...
+for, and it mixes active with basal. It was stored in the energy series, which means
+active energy, so the values were inflated by BMR. Worse, rollUp anchors its buckets
+to the requested range start, so every historical sync wrote the same hour at a
+different recorded_at and the unique constraint could not collapse them — a day with
+four syncs summed to four times the real figure (#1577).
+
+Ingestion now reads active-energy-burned and basal-energy-burned instead, which carry
+real records with their own intervals. Those rows are tagged with an external_id
+(see _TAGGED_SERIES in google/health_api/data_247.py), so this purge deletes only the
+untagged legacy rows and leaves anything the current code wrote alone.
```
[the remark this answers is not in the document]
czajkub: who cares

backend/app/services/providers/google/health_api/data_247.py
```diff
@@ -409,6 +531,9 @@ def _sample(
             value=value,
             series_type=series_type,
             is_daily_total=is_daily_total,
+            external_id=(
+                f"{series_type.value}:{recorded_at.isoformat()}" if series_type is SeriesType.energy else None
+            ),
```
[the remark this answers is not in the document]
czajkub: what raw basal intervals? there is no basal energy in google health api, only total and active. there is no raw basal metrics

backend/scripts/data_migrations/purge_google_total_calories_energy.py
```diff
@@ -0,0 +1,83 @@
...
+    db.commit()
+    print(f"Deleted {deleted} untagged Google Health API energy row(s) written from total-calories.")
+    return deleted
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--dry-run", action="store_true", help="report what would be deleted and exit")
+    args = parser.parse_args()
+    with SessionLocal() as db:
```
KaliszS, line 78: It should run in batches, for performance reasons.

backend/scripts/data_migrations/purge_google_total_calories_energy.py
```diff
@@ -0,0 +1,83 @@
...
+from app.database import SessionLocal
+
+PROVIDER = "google"
+SOURCE = "google_health_api"
+SERIES_CODE = "energy"
+
+_PARAMS = {"provider": PROVIDER, "source": SOURCE, "code": SERIES_CODE}
+
+_WHERE = """
+    FROM data_point_series dps
```
KaliszS, line 43: We should also take archive table into consideration.

backend/app/services/providers/google/health_api/metrics/derived.py
```diff
@@ -0,0 +1,20 @@
...
+from app.schemas.enums import SeriesType
+from app.schemas.providers.google import DailyRollupSpec, DerivedDailyMetric
+
+DERIVED_DAILY_METRICS: tuple[DerivedDailyMetric, ...] = (
+    DerivedDailyMetric(
+        "basal-energy-derived",
+        SeriesType.basal_energy,
+        DailyRollupSpec("total-calories", "totalCalories", "kcalSum", max_range_days=14),
+        DailyRollupSpec("active-energy-burned", "activeEnergyBurned", "kcalSum"),
+        lambda total, active: total - active,
```
KaliszS, line 18: Can't we catch total calories and active energy from different data sources? If so, we could end up with negative basal energy or just wrong value.

backend/app/services/providers/google/health_api/data_247.py
```diff
@@ -252,6 +278,102 @@ def _chunk_range(start: datetime, end: datetime, max_days: int) -> Iterator[tupl
...
+        totals: dict[datetime, Decimal] = {}
+        for chunk_start, chunk_end in self._chunk_range(start, end, spec.max_range_days):
+            for point in self._fetch_daily_rollup(db, user_id, endpoint, chunk_start.date(), chunk_end.date()):
+                day = parse_date((point.get("civilStartTime") or {}).get("date"))
+                value_obj = point.get(spec.value_key)
+                if day is None or not isinstance(value_obj, dict):
+                    continue
+                value = read_number(value_obj, spec.field, None, spec.scale)
+                if value is not None:
+                    totals[day] = value
```
KaliszS, line 337: again, we should check whether it can return more than one value (multiple sources), becaue if so, we would end up with updating that value with the lastest one

backend/app/services/providers/google/health_api/metrics/__init__.py
KaliszS, line 1: Shouldn' we also update coverage.py?

backend/app/services/providers/google/health_api/data_247.py
```diff
@@ -252,6 +278,102 @@ def _chunk_range(start: datetime, end: datetime, max_days: int) -> Iterator[tupl
...
+        return totals
+
+    def _fetch_daily_rollup(
```
KaliszS: It's very similar to `_fetch_rollup_window`, looks like we've got some redundancy.

backend/app/schemas/providers/google/health_api.py
```diff
@@ -79,6 +80,39 @@ class ListSpec:
...
+
+    data_type: str
+    value_key: str
+    field: str
+    scale: Decimal = Decimal(1)
+    max_range_days: int = 90
+
+
+@dataclass(frozen=True)
+class DerivedDailyMetric:
```
KaliszS, line 101: Isn't it a little over-engineered?

KaliszS (changes requested):
Don't we also need to change anything in tests?

backend/app/schemas/providers/google/health_api.py
czajkub: It might be, but i felt like it would be cleaner to read this way and more extensible if we were to add more metrics in this manner

backend/app/services/providers/google/health_api/metrics/__init__.py
czajkub: true, added the derived basal metric. for now it wont change the docs since we have the basal type from SDK (and this is only one logical provider), but in the future we would see a change

backend/app/services/providers/google/health_api/metrics/derived.py
czajkub: I tested this and it looks like 3-rd party active energy sources are in fcat not counted into total. Thankfully rollup/dailyRollup has a data source family field, so we can ask only for first party google data and make sure that the calculated basal takes only that into account (verified on my data)

backend/app/services/providers/google/health_api/data_247.py
czajkub: yep, basically same requests with different body, now just passing the body based on request params

backend/scripts/data_migrations/purge_google_total_calories_energy.py
czajkub: I feel like this is gonna be a bit difficult since external id which we depend on here is not present in the archive table (since we aggregate multiple records into one), I would assume this is enough of an edge case that it would take a lot of time to think through and implement, and it would barely affect anyone

backend/scripts/data_migrations/purge_google_total_calories_energy.py
czajkub: Do you mean batching the deletes? I could add that, but what would the batches size be? I don't think the amount of records would be big enough, before the change there'd be at most around thousands of records (i had this many from ~1 year back), I don't know if a difference would be noticable at this scale

backend/scripts/data_migrations/purge_google_total_calories_energy.py
```diff
@@ -0,0 +1,156 @@
...
+    total = 0
+    while True:
+        n = db.execute(stmt, {**params, "batch": batch}).rowcount  # ty: ignore[unresolved-attribute]
+        db.commit()
+        total += n
+        if n < batch:
+            return total
+
+
+def run(db: Session, dry_run: bool, batch: int = DEFAULT_BATCH, include_archive: bool = True) -> PurgeResult:
```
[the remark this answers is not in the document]
KaliszS: We are going to stay with opt-out anyway

Issue #1491, opened 2026-08-27
Split SDK and OAuth connection models (one slug = two independent channels)


bartmichalak:
`UserConnection` has a unique index on `(user_id, provider)`, so for a provider that supports both channels (today: **Google** - `sdk/` Health Connect + `health_api/` OAuth cloud) the SDK and OAuth integrations share **a single row and a single status**. These are independent integrations: different scopes, different data sets, separate lifecycles.

**Impact**

- One `status` and one `last_synced_at` shared across two integrations.
- On an SDK upload, `ensure_sdk_connection` reactivates the shared row and emits `connection.created`, even though the OAuth channel did not change (and vice versa).
- No way to report the state of each channel independently over webhooks.

**Context**

First raised in the review of #1423 (connection `created`/`revoked` webhook events).

czajkub:
Resolved in #1634

PR #1634, opened 2026-09-15
refactor: Google `health_connect` and `google_health` provider split

czajkub:
As proposed in #1491, this PR splits `google` into two providers - `health_connect` for SDK path and `google_health` for the Google Health API, so they both contain one connection row each and it is not merged together.

Adding data migration - which transfers data from `google_health_api` into the google health provider (verified with all ingestion methods that they contain this source with real data), rest goes to SDK

For old `/google/` endpoints - there is a problem to keep everything backwards possible, but the best way I found is this:
- old registered `/google/` webhooks - we add an "alias", so those will still work and deliver data - both on `/google/` and `/google_health/`, but newly registered ones are registered under `google_health/`
- already existing user connections with periodic sync - will work just fine
- also, SDK will still send `google` as provider name in payload body for Health Connect - I kept `google` as a new alias for `health_connect` provider and health connect option for the future - however we will need to make sure that users will have OW version over 0.8.1 that would accept the `health_connect` provider, so with this PR merged (if we do decide to update it, it can stay as google for a good while)
- new user connections oauth - the problem here is that our redirect uri must match with the one in GCP. for sending correct redirect URI, we add a new flag `GOOGLE_LEGACY_OAUTH_PATH`. By default in config it's set to `true` - so already existing connections with `google` dont need to change anything, and we keep the inbound path fully backwards compatible. However in `.env.example` and in docs, we encourage to set it to `false` and use the new `/google_health` path to prevent naming conflicts. In practice - new integrations or ones adjusting to the new provider split need to add one more env variable that's different from the default value. Both endpoints are active, what the flag controls is the `redirect_uri` we send for Google to verify. 

What CAN be breaking: in outgoing webhooks, we send `provider: google`, so someone filtering on that can see an issue. The scope here is: `connection.created/revoked`, data events' `source.provider`, sync events. We can't keep an alias for google here as it would be ambiguous, so this needs to be communicated in release notes.

Another thing - in s3, we still keep old `google` key. This shouldn't be a problem since it should only be used for debugging / storing raw json problems, but it might be worth mentioning in release notes - since after this change it will be separated into two google paths

Also updating old migration to consider not only google but also google_health provider - the purge migration would run after the api/sdk split, so we need to update that as well

image

What this also gives us - two separate rows on the coverage matrix (before it was merged Google, without split we couldn't achive that)

image

image

## Summary by CodeRabbit

* **New Features**
  * Added separate Google Health and Health Connect provider identities.
  * Added support for Health Connect SDK synchronization, including legacy `google` alias normalization.
  * Added provider-specific coverage and connection handling.
  * Updated provider labels and sync displays in the user interface.
* **Bug Fixes**
  * Webhook routes now correctly resolve provider URL aliases.
  * Existing Google webhook paths remain supported for registered integrations.
* **Documentation**
  * Updated provider coverage, OAuth URLs, SDK behavior, and raw payload documentation.


docs/providers/coverage.mdx
```diff
@@ -416,8 +413,8 @@ Key differences and limitations to be aware of when working with each provider:
   - 14 activity type ID mappings plus 25 name-based fallbacks to the unified taxonomy; unrecognized activities map to "other"
 </Accordion>
 
-<Accordion title="Google Health — Health Connect (mobile SDK)">
-  Google is a single provider with two ingestion paths under one `google` identity. This is the **Health Connect** path; stored records carry the originating app's name in `source` (e.g. the app that wrote to Health Connect), which distinguishes them from the cloud API path below.
+<Accordion title="Health Connect (mobile SDK)">
+  The `health_connect` provider — data pushed from an Android device. Stored records carry the originating app's name in `source` (e.g. the app that wrote to Health Connect). The Google Health API cloud path is the separate `google_health` provider, below.
```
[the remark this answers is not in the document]
czajkub: fixed

backend/app/api/routes/v1/oauth.py
```diff
@@ -24,6 +25,15 @@
 settings_service = ProviderSettingsService()
 
 
+def resolve_provider(slug: str) -> ProviderName:
+    """Provider behind a path segment, accepting the pre-split slug. 400 keeps the
```
KaliszS: Redundant docstring

backend/app/constants/sdk_providers.py
```diff
@@ -0,0 +1,19 @@
+from app.schemas.enums import ProviderName
+
+SDK_PROVIDERS: frozenset[str] = frozenset(
```
KaliszS: Thats double source of truth, we should take that information from strategy, which already has information about being SDK-based provider.

KaliszS (changes requested):

backend/app/constants/sdk_providers.py
czajkub: fair enough, updated it to use the capabilites from strategy

KaliszS (approved):

PR #1573, opened 2026-09-09
feat(mcp): add HTTP transport with bearer/OAuth auth

Pixsd:
## What does this change?

Adds an HTTP transport to the MCP server (`MCP_TRANSPORT=http`) alongside the existing stdio mode, so it can run as a standalone server for remote clients, with either a static bearer token or a self-contained OAuth 2.1 flow for auth.

## Why?

stdio only works for clients that spawn the server as a local subprocess (Claude Desktop, Cursor). claude.ai's custom connectors need an HTTP-reachable server with OAuth - a bearer token alone doesn't fit that flow, since there's no authorization server for the client to register against. Bearer stays as the simpler option for clients that accept a manual token (MCPJam and similar).

Considered subclassing fastmcp's `InMemoryOAuthProvider` instead of writing our own - dropped that because it's documented as a testing double and doesn't track the `resource`/audience restriction we need for token scoping.

Related issue: none.

## How did you test this?

- Deployed the `mcp` service behind a reverse proxy on a real HTTPS domain, with `MCP_OAUTH_PASSWORD` set.
- Added it as a custom connector in claude.ai (Settings → Connectors), went through the full OAuth flow end to end (dynamic client registration → password login → token exchange), and queried real wearable data through it.
- Bearer-token mode and the edge cases from a review pass (confused-deputy phishing on the login page, dropped resource restriction on token refresh, config validation, secrets ending up in the Docker image) were verified with targeted checks in an isolated environment, not against a live client.

## AI usage

Built with Claude Code (Sonnet 5): implementation, a code-review pass that surfaced and fixed several issues in the auth flow, and this PR description.

## Summary by CodeRabbit

- **New Features**
  - Added an MCP server for querying wearable health data through AI assistants.
  - Supports local stdio connections and remote HTTP access.
  - Added bearer-token and OAuth 2.1 authentication for remote clients.
  - Added Docker Compose support for running the MCP server locally.
  - Added protections for OAuth client registration, password attempts, and token handling.

- **Documentation**
  - Documented setup, supported assistants, transport modes, authentication, remote connections, and security guidance.

PR #1640, opened 2026-09-15
revert(mcp): HTTP transport with bearer/OAuth auth (#1573)

bartmichalak:
Reverts the-momentum/open-wearables#1573

### Why?
#1573 was merged before we completed internal testing and before the requested review was finished. Since it introduces a new auth surface (bearer token + self-contained OAuth 2.1 flow) and a new deployment mode for the MCP server, we want to verify it end to end on our side and make sure the user experience is right for everyone using this feature.

## Summary by CodeRabbit

- **Removed Features**
  - Removed the MCP server’s HTTP transport, OAuth authentication, and bearer-token authentication.
  - Removed the Docker Compose service and container packaging for running the MCP server remotely.
  - Removed related configuration options for HTTP hosting, ports, public URLs, OAuth, and bearer tokens.

- **Documentation**
  - Updated MCP documentation to describe the remaining supported client and local usage options.


bwickner:
Opened a discussion rather than pile onto this thread: https://github.com/the-momentum/open-wearables/discussions/1668

Short version: as merged, the token names the client rather than the user, so anyone past the shared password reaches every user's data through the instance API key. Seemed like something to settle before this lands again. I have a per-user variant running on my fork if that's useful input.

PR #1646, opened 2026-09-16
refactor(backend): move shared SDK ingestion out of apple/ namespace

### LINKED ISSUE #916: Refactor: backend ingestion under apple/ actually serves all SDK providers

bartmichalak:
## What does this change?

Moves the mobile SDK ingestion code (`import_service`, `sleep_service`, `device_resolution`) from `services/apple/healthkit/` to `services/sdk/`, and renames `constants/workout_types/apple_sdk.py` to `sdk.py`. No behavior changes.

## Why?

This code handles all SDK providers (apple, samsung, health_connect), but the `apple/` path suggests it's Apple-only, which is really confusing. For example, #1123 fixes Health Connect blood glucose, yet the whole diff lands in `services/apple/healthkit/` and `test_apple_sdk_import.py`, so it looks like an Apple change. The same naming is how Apple-specific unit handling ended up applied to Health Connect data in #917.

This is only the rename (option 1 from the issue). Per-provider unit adapters (option 2) are a separate change, since they alter logic rather than just moving files.

`apple_xml/` stays under `apple/`, since it really is Apple-only. Logger names follow `__name__`, so SDK import/sleep logs now show up under `app.services.sdk.*`. Celery task names don't change.

Relates to #916

## How did you test this?

- Full suite relies on CI

## AI usage

Claude Code (Opus 5) for finding references, doing the rename and running the checks.

## Summary by CodeRabbit

* **Enhancements**
  * Unified workout type handling across Apple, Samsung, and Health Connect integrations.
  * Standardized device recognition and sleep data processing across supported mobile health platforms.
  * Improved consistency when importing SDK data and mapping workouts.

* **Tests**
  * Updated automated coverage for shared mobile SDK processing, sleep services, and upload workflows across supported providers.


backend/app/constants/workout_types/sdk.py
```diff
@@ -4,9 +4,10 @@
 
 
 class SDKWorkoutType(StrEnum):
-    """Apple HealthKit SDK workout activity types.
+    """Mobile SDK workout activity types.
 
-    These are the workout types sent from the iOS SDK in snake_case format.
+    Wire vocabulary shared by all SDK providers (apple, google, samsung), in snake_case.
```
knowald: The provider list does not match the actual SDK provider set. `sdk_providers()` derives it from strategies with `client_sdk=True`, which are `apple`, `samsung` and `health_connect`. `google` is only a legacy wire alias for `health_connect` (`SDK_PROVIDER_ALIASES`). Naming `google` here reintroduces the confusion that #1634 removed.

Suggest: `(apple, samsung, health_connect)`.

(Reviewed by Claudex)

backend/app/constants/workout_types/sdk.py
```diff
@@ -4,9 +4,10 @@
...
+    Names follow HealthKit; the Android SDK translates Health Connect types to them.
```
knowald: This contradicts the rest of the file. The Health Connect block (around lines 118-127) says those members mirror `HealthConnectManager.mapExerciseType` and that the wire format is uppercase, and `get_unified_workout_type` lowercases input for exactly that reason (its doctest maps `CYCLING_STATIONARY`). There are also ~27 HC-only members (`CYCLING_STATIONARY`, `SWIMMING_POOL`, `FOOTBALL_AMERICAN`, ...) with no HealthKit counterpart, so "names follow HealthKit" and "in snake_case" on the line above are both wrong for a large part of the enum.

Suggest something like: "Carries the HealthKit snake_case vocabulary plus the Health Connect vocabulary the Android SDK emits in uppercase; lookup normalizes case."

(Reviewed by Claudex)

backend/tests/providers/test_provider_coverage.py
```diff
@@ -38,8 +38,8 @@
 # shared sleep service, so those shared files are scanned for them too.
 SDK_PROVIDERS = {"apple", "samsung", "health_connect"}
 SDK_SHARED_FILES = (
-    Path("app/services/apple/healthkit/import_service.py"),
-    Path("app/services/apple/healthkit/sleep_service.py"),
+    Path("app/services/sdk/import_service.py"),
```
knowald, line 41: The comment two lines above `SDK_PROVIDERS` (line 36) still says "shared healthkit pipeline" while the paths here now point at `app/services/sdk/`. The pipeline also serves samsung and health_connect, so "SDK pipeline" would be more accurate.

(Reviewed by Claudex)

knowald (commented):
Review of the rename. No stale code references found, imports and touched tests pass. Three doc/comment findings inline. (Reviewed by Claudex)

knowald:
@coderabbitai Please verify the three review comments marked "(Reviewed by Claudex)" and say whether you agree with each.

czajkub (approved):
This part looks good, but I believe there are also a few issues to address similar in topic:

- I assume XML processing is still under `services.apple` - we can consider moving this as well
- in `services.providers`, i believe there is a hanging Apple strategy - though we need it for `ProviderName` and strategy, but it still is outside of the rest of SDK processing, same thing with Health Connect and Samsung

So I think that this clears the problem from issue #916, but I still would not consider it to be fully clean

Though it works for now, and has for a few months - so not an urgent topic 😄

bartmichalak:
@czajkub Good points, but I'll keep this PR to the rename only. Moved them to #1649 - would appreciate it if you could pick it up 😀

bartmichalak:
Thanks for the review @knowald, all three fixed 🫡

PR #1680, opened 2026-09-19
fix(integrations): let callers name the statuses they expect from api_client

slavarosin:
Hi!

Follow-up to #1575. You were right that the Sentry angle was moot and that the real noise is the error line in `api_client.py`, logged before the exception reaches the provider code. This goes after that line.

- `make_authenticated_request` gets `quiet_statuses=(...)`. A matching status still raises, so the flow is unchanged, it just isn't logged as an error first.
- Its opt-in, so other providers see no change unless they pass it (your concern on #1575).
- `BaseWorkoutsTemplate._make_api_request` forwards it, and Strava's `/streams` call passes `(404,)`. That closes the manual activity case.
- Both error logs now carry the `endpoint` and status. Before there was nothing to tell which call failed.

Separately, an empty body (204, or a 200 with nothing) now returns `None` instead of going into `.json()`. That JSONDecodeError was turned into an error log plus a 500. Polar answers "nothing new" that way on a few endpoints.

The Sensorbio override grew the same keyword so it stays substitutable.

Verification: full suite 2520 passed, `ruff` and `ty` clean. The Strava test goes through the real template, so a base that accepts the argument but drops it would fail.

Let me know what you think!

https://claude.ai/code/session_01Ft2S4f1NUYm9MoRLktEJdU

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of successful API responses with no content, preventing unnecessary errors.
  * Manual activities without workout stream data are now handled cleanly without generating misleading error reports.
  * Error details now include the affected endpoint, making troubleshooting clearer.

* **Tests**
  * Added coverage for empty responses, expected missing data, error reporting, and normal JSON responses.


slavarosin:
Re the outside-diff note on strava/workouts.py:334: leaving that one on purpose. A quiet 404 branch in the Strava handler is exactly what #1575 did, and it was closed because that only swaps a warning for an info line (4xx already skip Sentry since #1203). This PR removes only the error-level line from api_client that was the actual problem there, the Strava side stays as it is.

bartmichalak (changes requested):
Hi @slavarosin 
We updated our [contributing guidelines](https://github.com/the-momentum/open-wearables/blob/main/CONTRIBUTING.md) a while ago - could you take a look and update the PR description to follow the [template](https://github.com/the-momentum/open-wearables/blob/main/.github/PULL_REQUEST_TEMPLATE.md)? Makes review much faster on our side.

